PR #41, opened 2019-12-16
Geocoding APIs

jsani-radar:
https://app.clubhouse.io/radarlabs/story/987/add-geocoding-apis-to-ios-sdk-v3


RadarSDK/RadarAPIClient.m
```diff
@@ -389,4 +391,109 @@ - (void)searchGeofencesWithLocation:(CLLocation * _Nonnull)location
...
+    NSString *host = [RadarSettings host];
+    NSString *url = [NSString stringWithFormat:@"%@/v1/geocode/ip", host];
+    url = [url stringByAddingPercentEncodingWithAllowedCharacters:[NSCharacterSet URLQueryAllowedCharacterSet]];
+
+    NSDictionary *headers = @{
+        @"Authorization": publishableKey,
+        @"Content-Type": @"application/json",
+        @"User-Agent": [RadarUtils userAgent],
+        @"X-Radar-Config": @"true",
+    };
```
pingxia0823: are these the standard headers for our request? thinking if we shall consolidate that to an util method

RadarSDK/RadarAddress.m
```diff
@@ -0,0 +1,165 @@
...
+        number = (NSString *)addressNumberObj;
+    }
+
+    // TODO (jsani): how to serialize RadarAddressConfidence?
+
+    if (coordinate && confidence) {
+        return [[RadarAddress alloc] initWithCoordinate:coordinate formattedAddress:formattedAddress country:country countryCode:countryCode countryFlag:countryFlag state:state stateCode:stateCode postalCode:postalCode city:city borough:borough county:county neighborhood:neighborhood number:number confidence:confidence];
+    }
+
+    return nil;
```
pingxia0823: There seems to be many boilerplate code. kind of error prone too. I would imagine this would happen for all data types syncing between server and client

Two ideas I could think of now are:
1. we could employ some strong scheme for data serialization during networking, like protobuf
2. we could use some local script (like the remodel tool from Facebook) to auto generate these helper methods

nickpatrick: @pingxia0823 Yeah there's a lot of repetition in the deserialization and networking code. We should def refactor at some point, can tackle separately though

RadarSDK/RadarAPIClient.m
jsani-radar: definitely a good idea, but it seems like there are certain instances where we omit the `X-Radar-Config` key -- https://github.com/radarlabs/radar-sdk-ios/blob/v3/RadarSDK/RadarAPIClient.m#L257-L261

Nick, no rush, but could you provide more insight here? This seems like the `verifyEvent` path, which I don't think gets much usage..

RadarSDK/RadarAddress.m
jsani-radar: yeah, comment 100% makes sense as we mature with more models / services, but I'm not sure it's in scope for the `v3` release. For now, since my PR only adds one new model (and an enum), I decided to stick with existing conventions rather than require a huge refactor

pingxia0823: Yeah, just some food for thought

RadarSDK/RadarAddress.m
```diff
@@ -0,0 +1,187 @@
...
+    NSString *stateCode;
+    NSString *postalCode;
+    NSString *city;
+    NSString *borough;
+    NSString *county;
+    NSString *neighborhood;
+    NSString *number;
+
+    // TODO (jsani): is this a sensible default for 'confidence', or should we have another enum entry?
+    RadarAddressConfidence confidence = RadarAddressConfidenceFallback;
```
tjulien: For event confidences, we default to `"none"` (or `0`) - we probably want to do the same thing here, since `fallback` means something more specific.

https://github.com/radarlabs/radar-sdk-ios/blob/v3/RadarSDK/RadarEvent.m#L77

RadarSDK/RadarAddress.m
```diff
@@ -0,0 +1,186 @@
...
+    if (confidenceObj && [confidenceObj isKindOfClass:[NSString class]]) {
+        NSString *confidenceStr = (NSString *)confidenceObj;
+
+        if ([confidenceStr isEqualToString:@"exact"]) {
+            confidence = RadarAddressConfidenceExact;
+        } else if ([confidenceStr isEqualToString:@"interpolated"]) {
+            confidence = RadarAddressConfidenceInterpolated;
+        } else {
+            confidence = RadarAddressConfidenceFallback;
+        }
```
tjulien: Need to add the `None` case

tjulien: Oops, I just noticed you default to it up top

RadarSDK/RadarAddress.m
```diff
@@ -0,0 +1,186 @@
...
+    NSString *state;
+    NSString *stateCode;
+    NSString *postalCode;
+    NSString *city;
+    NSString *borough;
+    NSString *county;
+    NSString *neighborhood;
+    NSString *number;
+
+    RadarAddressConfidence confidence = RadarAddressConfidenceNone;
```
tjulien: nit: we could move this down by the confidence parsing code

nickpatrick:
@jsani-radar Awesome work on these

RadarSDK/RadarAddress.m
jsani-radar: right, definitely would make things easier to read -- was getting compiler warnings that `confidence` is potentially uninitialized if I wasn't assigning a value here. Could happen for example if the SDK gets something like an array back for `confidence` and it doesn't enter the block to map it to the enum value.

I should mark the enum entry for `None = 0` so we explicitly set it as the default regardless.

jsani-radar: there is precedent for initializing the "Unknown / None" enum variety in our other deserializers - https://github.com/radarlabs/radar-sdk-ios/blob/v3/RadarSDK/RadarEvent.m#L70

nickpatrick:
@jsani-radar Cool if we merge these today? I'll package everything up and ship a beta this weekend, then start updating React Native/Capacitor/Cordova to match. cc @tjulien

jsani-radar:
Sure thing! 🚢

PR #44, opened 2019-12-26
treat arrivals as stops, handle stale inaccurate locations

nickpatrick:


nickpatrick:
https://app.clubhouse.io/radarlabs/story/1142/update-tracking-options-and-stopped-logic-in-sdk-v3

RadarSDK/RadarTrackingOptions.m
```diff
@@ -75,8 +75,8 @@ + (RadarTrackingOptions *)responsive {
     options.showBlueBar = NO;
     options.useStoppedGeofence = YES;
     options.stoppedGeofenceRadius = 200;
-    options.useMovingGeofence = YES;
-    options.movingGeofenceRadius = 200;
+    options.useMovingGeofence = NO;
```
tjulien: no more moving geofence for responsive?

nickpatrick: Not needed anymore. Low power location manager will keep the process awake while moving.

nickpatrick:
@jsani-radar We'll need to keep committing `dist/` until we move it to Circle. Not required in development branches necessarily, but required before we ship. We can figure it out offline.

RadarSDK/RadarTrackingOptions.m
tjulien: Is it worth ripping out `useMovingGeofence` since none of the 3 presets use it now? Or worth keeping around for a custom option.
🤔

nickpatrick: Let's keep it for consistency with Android and in case we ever need or want to add it back.

PR #49, opened 2019-12-30
Reverse geocode current location

jsani-radar:
The previous PR (https://github.com/radarlabs/radar-sdk-ios/pull/41) only had support for sending an explicit location to be reverse geocoded.

This PR adds the ability to reverse geocode the user's current location using `RadarLocationManager`, as we have for search endpoints -- https://github.com/radarlabs/radar-sdk-ios/blob/v3/RadarSDK/Radar.h#L313-L341

Once there's confirmation about the method signatures, I can do the same for the Android SDK


jsani-radar:
https://app.clubhouse.io/radarlabs/story/1108/reverse-geocode-user-s-current-location

PR #50, opened 2019-12-30
Geocode Custom IP

jsani-radar:
The previous PR (#41) only had support for geocoding the IP address of the user's device.

This PR adds the ability to geocode a *custom / provided* IP address.

### DIFF
--- a/RadarSDK/Radar.h
+++ b/RadarSDK/Radar.h
@@ -403,6 +403,12 @@ typedef void(^ _Nonnull RadarIPGeocodeCompletionHandler)(RadarStatus status, Rad
 + (void)ipGeocode:(RadarIPGeocodeCompletionHandler)completionHandler;
 
 /**
+ Geocodes a provided IP address, returning a region.
+ */
++ (void)ipGeocode:(NSString * _Nonnull)ip
+completionHandler:(RadarIPGeocodeCompletionHandler)completionHandler;
+
+/**
  Sets the log level for debug logs.
  
  @param level The log level.
--- a/RadarSDK/Radar.h
+++ b/RadarSDK/Radar.h
@@ -400,15 +400,15 @@ typedef void(^ _Nonnull RadarIPGeocodeCompletionHandler)(RadarStatus status, Rad
 
  @param completionHandler A completion handler.
  */
-+ (void)ipGeocode:(RadarIPGeocodeCompletionHandler)completionHandler;
++ (void)geocodeDeviceIP:(RadarIPGeocodeCompletionHandler)completionHandler;
 
 /**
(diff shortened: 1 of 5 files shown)


RadarSDK/RadarAPIClient.m
```diff
@@ -462,14 +462,24 @@ - (void)reverseGeocode:(CLLocation *)location completionHandler:(RadarGeocodeAPI
...
-- (void)ipGeocode:(RadarIPGeocodeAPICompletionHandler)completionHandler {
+- (void)ipGeocode:(NSString *)ip completionHandler:(RadarIPGeocodeAPICompletionHandler)completionHandler {
     NSString *publishableKey = [RadarSettings publishableKey];
     if (!publishableKey) {
         return completionHandler(RadarStatusErrorPublishableKey, nil, nil);
     }
 
     NSString *host = [RadarSettings host];
-    NSString *url = [NSString stringWithFormat:@"%@/v1/geocode/ip", host];
+
```
pingxia0823: does client need to do some sanity check on the `ip` ?

RadarSDK/Radar.h
```diff
@@ -396,6 +396,15 @@ typedef void(^ _Nonnull RadarIPGeocodeCompletionHandler)(RadarStatus status, Rad
  */
 + (void)ipGeocode:(RadarIPGeocodeCompletionHandler)completionHandler;
```
pingxia0823: nit: maybe it's better to revise the name, so the difference between `ipGeocode:` and `ipGeocode:completionHandler` is more obvious?

jsani-radar: agreed, good call, revised the naming to `geocodeDeviceIP` vs `geocodeIP` -- thoughts?

RadarSDK/RadarAPIClient.m
jsani-radar: @kochis thoughts? I assume whatever validation we do on the FE won't be as comprehensive as https://www.npmjs.com/package/ip-address

I can try splitting on `.` or `:` and making sure there are > 3 elements, but I don't want to trigger any false negatives

jsani-radar:
https://app.clubhouse.io/radarlabs/story/1109/geocode-custom-ip-address

RadarSDK/RadarAPIClient.m
tjulien: We do validation on the server so we should be ok to leave validation out on the client side:

https://github.com/radarlabs/server/blob/8bb83357266a95cdadef9a9040387b9a00e7d578/app.ts#L1363
https://github.com/radarlabs/server/blob/master/server/utils/Validate.ts#L240

PR #51, opened 2020-01-08
v3 alpha.2

nickpatrick:
- Add `setAdIdEnabled()`
- Docs and naming changes for consistency
- Remove IP geocode for non-device ID (if won't be used, don't need surface area)


RadarSDK/Radar.h
```diff
@@ -372,44 +372,40 @@ typedef void(^ _Nonnull RadarIPGeocodeCompletionHandler)(RadarStatus status, Rad
...
- Geocodes a given query string, returning an array of addresses.
+ Geocodes an address, converting address to coordinates.
 
- @param query The address string to geocode.
+ @param query The address to geocode.
  @param completionHandler A completion handler.
  */
-+ (void)geocode:(NSString * _Nonnull)query
-        completionHandler:(RadarGeocodeCompletionHandler)completionHandler;
++ (void)geocodeAddress:(NSString * _Nonnull)query
```
tjulien: on Android is this function just called `geocode`?

tjulien: I guess trying to be idiomatic with ios by putting the first param in the function name?

nickpatrick: Correct, following Objective-C convention here. Compare to https://developer.apple.com/documentation/corelocation/clgeocoder/2890753-geocodeaddressstring. See also the change to `reverseGeocodeWithCompletionHandler:`.

In Swift, they will be `geocode(address:completionHandler:)` and `reverseGeocode(completionHandler:)`, respectively.

PR #54, opened 2020-01-31
Add autocomplete and distance, update params, move headers to helper method

nickpatrick:
- Update params to match docs
- Refactor headers to helper method
- Add `ERROR_FORBIDDEN` and `ERROR_BAD_REQUEST`
- Expose autocomplete and distance APIs with tests and sample code


RadarSDK/RadarAPIClient.m
```diff
@@ -37,6 +37,19 @@ - (instancetype)init {
     return self;
 }
 
++ (NSDictionary *)headersWithPublishableKey:(NSString *)publishableKey {
+    return @{
+        @"Authorization": publishableKey,
+        @"Content-Type": @"application/json",
+        @"X-Radar-Config": @"true",
```
jsani-radar: do we always want to return back the client's config in the SDK?

I know we use this for internal testing / confirmation, but not sure we need to expose that to customers

RadarSDK/RadarAPIClient.m
```diff
@@ -133,46 +133,20 @@ - (void)trackWithLocation:(CLLocation * _Nonnull)location
...
     NSString *url = [NSString stringWithFormat:@"%@/v1/track", host];
     url = [url stringByAddingPercentEncodingWithAllowedCharacters:[NSCharacterSet URLQueryAllowedCharacterSet]];
     
-    NSDictionary *headers = @{
-        @"Authorization": publishableKey,
-        @"Content-Type": @"application/json",
-        @"User-Agent": [RadarUtils userAgent],
-        @"X-Radar-Config": @"true",
-    };
+    NSDictionary *headers = [RadarAPIClient headersWithPublishableKey:publishableKey];
```
jsani-radar: so, it seems like we're going to be sending certain params (ex. `deviceModel`, `deviceOS`, etc.) in the request params as well as the headers? is that intended?

RadarSDK/RadarAPIClient.m
```diff
@@ -37,6 +37,19 @@ - (instancetype)init {
...
++ (NSDictionary *)headersWithPublishableKey:(NSString *)publishableKey {
+    return @{
+        @"Authorization": publishableKey,
+        @"Content-Type": @"application/json",
+        @"X-Radar-Config": @"true",
+        @"X-Radar-Device-Make": [RadarUtils deviceMake],
+        @"X-Radar-Device-Model": [RadarUtils deviceModel],
+        @"X-Radar-Device-OS": [RadarUtils deviceOS],
+        @"X-Radar-Device-Type": [RadarUtils deviceType],
+        @"X-Radar-SDK-Version": [RadarUtils sdkVersion]
```
jsani-radar: are we omitting `User-Agent` now? is that to preserve the original header from the device?

nickpatrick: After discussing with @tjulien, we want to keep the original `User-Agent` (which can contain useful info, especially on Android), and send this info in separate headers instead

RadarSDK/RadarAPIClient.m
nickpatrick: Later we can pick these up from headers server-side, but keeping in the body for backcompat right now

RadarSDK/RadarAPIClient.m
nickpatrick: The config bundle isn't being used on the client right now, but no issues retrieving it with each request

RadarSDK/RadarAPIClient.m
```diff
@@ -49,29 +62,16 @@ - (void)getConfig {
     if (userId) {
         [queryString appendFormat:@"&userId=%@", userId];
     }
-    [queryString appendFormat:@"&deviceId=%@", [RadarUtils deviceId]];
-    [queryString appendFormat:@"&deviceType=%@", [RadarUtils deviceType]];
-    [queryString appendFormat:@"&deviceMake=%@", [RadarUtils deviceMake]];
```
jsani-radar: do we not need to duplicate certain params in the headers / body for backcompat, as we do with track?

PR #55, opened 2020-02-05
add getContext API

corypisano:
To do:

- [x] add context mock & test cases


nickpatrick:
@corypisano Looks great. Some minor nitpicks on the docs, but I can update those after we merge. Let me know if you need help with the merge conflicts

RadarSDK/Radar.h
```diff
@@ -416,6 +424,34 @@ typedef void(^ _Nonnull RadarIPGeocodeCompletionHandler)(RadarStatus status, Rad
...
+ 
+ @param location A location to get context for.
+ @param completionHandler An optional completion handler.
+ 
+ @see https://radar.io/documentation/sdk#ios-manual
+ */
++ (void)getContextWithLocation:(CLLocation * _Nonnull)location
+            completionHandler:(RadarGetContextCompletionHandler _Nullable)completionHandler
+    NS_SWIFT_NAME(getContext(location:completionHandler:));
+
```
pingxia0823: great documentation  👍

RadarSDK/RadarAPIClient.m
```diff
@@ -496,4 +497,39 @@ - (void)ipGeocodeWithCompletionHandler:(RadarIPGeocodeAPICompletionHandler)compl
...
+
+    NSString *host = [RadarSettings host];
+    NSString *url = [NSString stringWithFormat:@"%@/v1/context?%@", host, queryString];
+    
+    NSDictionary *headers = @{
+        @"Authorization": publishableKey,
+        @"Content-Type": @"application/json",
+        @"User-Agent": [RadarUtils userAgent],
+        @"X-Radar-Config": @"true",
+    };
```
pingxia0823: Not in the scope of this PR, but wonder if we could move these into some util function? I spot such headers several places in the codebase

RadarSDK/RadarContext.m
```diff
@@ -0,0 +1,106 @@
...
+    dma = [[RadarRegion alloc] initWithObject:dmaObj];
+    
+    id postalCodeObj = contextDict[@"postalCode"];
+    postalCode = [[RadarRegion alloc] initWithObject:postalCodeObj];
+    
+    if (contextGeofences) {
+        return [[RadarContext alloc] initWithLive:contextLive geofences:contextGeofences place:contextPlace country:country state:state dma:dma postalCode:postalCode];
+    }
+    
+    return nil;
```
pingxia0823: I wonder if we shall do some test on this network deserialization function `initWithObject ` because of a lot of boilerplate code here

RadarSDK/RadarContext.h
```diff
@@ -0,0 +1,68 @@
...
+ @see https://radar.io/documentation/geofences
+ */
+@property (nullable, copy, nonatomic, readonly) NSArray<RadarGeofence *> *geofences;
+
+/**
+ The place a location is at. May be `nil` if the location is not at a place, or if Places is not enabled.
+ 
+ @see https://radar.io/documentation/places
+ */
+@property (nullable, copy, nonatomic, readonly) RadarPlace *place;
```
pingxia0823: any danger that we may be overlapping / nested place? like xxx Mall / yyy store

RadarSDK/RadarAPIClient.m
```diff
@@ -496,4 +497,39 @@ - (void)ipGeocodeWithCompletionHandler:(RadarIPGeocodeAPICompletionHandler)compl
...
+            return completionHandler(status, nil, nil);
+        }
+        
+        id contextObj = res[@"context"];
+        RadarContext *context = [[RadarContext alloc]  initWithObject:contextObj];
+        if (context) {
+            return completionHandler(RadarStatusSuccess, res, context);
+        }
+
+        completionHandler(RadarStatusErrorServer, nil, nil);
```
pingxia0823: hmm, so when the server does return 200 but the result does not include context, we just return a nil result? what's the usage of other fields within 'res'?

RadarSDK/Radar.h
```diff
@@ -416,6 +424,34 @@ typedef void(^ _Nonnull RadarIPGeocodeCompletionHandler)(RadarStatus status, Rad
...
+ TODO(coryp): description and docs
+ Get context for a user's location once in the foreground, without including any user data.
+ 
+ @warning Note that these calls are subject to rate limits.
+ 
+ @param completionHandler An optional completion handler.
+ 
+ @see https://radar.io/documentation/sdk
+ */
++ (void)getContextWithCompletionHandler:(RadarGetContextCompletionHandler _Nullable)completionHandler
```
jsani-radar: I don't think we necessarily need `WithCompletionHandler` at the end here - this could be just `getContext` that fetches a location fix from the device and sends a context request to our BE

RadarSDK/Radar.h
```diff
@@ -416,6 +424,34 @@ typedef void(^ _Nonnull RadarIPGeocodeCompletionHandler)(RadarStatus status, Rad
...
+    NS_SWIFT_NAME(getContext(completionHandler:));
+
+/**
+ TODO(coryp): description and docs
+ Manually get context for a location, without including any user data.
+
+ @warning Note that these calls are subject to rate limits.
```
jsani-radar: I like these warnings 🎉 - wonder how they get rendered in the SDK docs

RadarSDK/Radar.m
```diff
@@ -215,6 +215,33 @@ + (void)ipGeocodeWithCompletionHandler:(RadarIPGeocodeCompletionHandler)completi
...
 }
 
++ (void)getContextWithCompletionHandler:(RadarGetContextCompletionHandler)completionHandler {
+    [[RadarLocationManager sharedInstance] getLocationWithCompletionHandler:^(RadarStatus status, CLLocation * _Nullable location, BOOL stopped) {
+        if (status != RadarStatusSuccess) {
+            if (completionHandler) {
+                completionHandler(status, nil, nil);
+            }
+            
+            return;
```
jsani-radar: not sure if we actually need a `return` here? the convention is a bit sporadic in the rest of the codebase, might be an opportunity to standardize how we handle error / empty state callbacks

RadarSDK/RadarAPIClient.m
jsani-radar: I think Nick has a PR that actually does this.. not sure if it's merged or a rebase is needed

RadarSDK/RadarAPIClient.m
```diff
@@ -496,4 +497,39 @@ - (void)ipGeocodeWithCompletionHandler:(RadarIPGeocodeAPICompletionHandler)compl
...
+        @"X-Radar-Config": @"true",
+    };
+    
+    [self.apiHelper requestWithMethod:@"GET" url:url headers:headers params:nil completionHandler:^(RadarStatus status, NSDictionary * _Nullable res) {
+        if (status != RadarStatusSuccess || !res) {
+            return completionHandler(status, nil, nil);
+        }
+        
+        id contextObj = res[@"context"];
+        RadarContext *context = [[RadarContext alloc]  initWithObject:contextObj];
```
jsani-radar: super nit: I think there's 2 spaces before `initWithObject` here

RadarSDK/RadarAPIClient.m
```diff
@@ -496,4 +497,39 @@ - (void)ipGeocodeWithCompletionHandler:(RadarIPGeocodeAPICompletionHandler)compl
...
+    NSString *url = [NSString stringWithFormat:@"%@/v1/context?%@", host, queryString];
+    
+    NSDictionary *headers = @{
+        @"Authorization": publishableKey,
+        @"Content-Type": @"application/json",
+        @"User-Agent": [RadarUtils userAgent],
+        @"X-Radar-Config": @"true",
+    };
+    
+    [self.apiHelper requestWithMethod:@"GET" url:url headers:headers params:nil completionHandler:^(RadarStatus status, NSDictionary * _Nullable res) {
```
jsani-radar: wonder if our `apiHelper` is smart enough to construct the params differently between `GET` and other requests, doesn't have to be in this PR, but then we'd be able to move code such as https://github.com/radarlabs/radar-sdk-ios/blob/0e09f35c4cd35aec42603f150d6cf716770d277d/RadarSDK/RadarAPIClient.m#L507-L508 into that layer instead of here

RadarSDK/RadarAPIClient.m
jsani-radar: if the result doesn't include the `context` key, then the above block won't get executed, and we'll throw a 500 - I think the point is that we always expect `context` to be a top-level key in the response for this route

RadarSDK/RadarContext+Internal.h
```diff
@@ -0,0 +1,26 @@
...
+
+#import <Foundation/Foundation.h>
+#import "RadarContext.h"
+#import "RadarGeofence.h"
+#import "RadarUserInsights.h"
+
+@interface RadarContext ()
+
+- (instancetype _Nullable)initWithLive:(BOOL)live
+                             geofences:(NSArray * _Nullable)geofences
```
jsani-radar: I don't think we need `_Nullable` for geofences - we're returning an empty array, even if no geofences are found

RadarSDK/RadarContext.m
```diff
@@ -0,0 +1,106 @@
...
+#import "RadarContext+Internal.h"
+#import "RadarGeofence+Internal.h"
+#import "RadarPlace+Internal.h"
+#import "RadarRegion+Internal.h"
+
+@implementation RadarContext
+
+- (instancetype _Nullable)initWithLive:(BOOL)live
+                             geofences:(NSArray *)geofences
+                                 place:(RadarPlace *)place
```
jsani-radar: `_Nullable` place

RadarSDK/RadarContext.m
```diff
@@ -0,0 +1,106 @@
...
+    // LIVE
+    id contextLiveObj = contextDict[@"live"];
+    if (contextLiveObj && [contextLiveObj isKindOfClass:[NSNumber class]]) {
+        NSNumber *contextLiveNumber = (NSNumber *)contextLiveObj;
+        
+        contextLive = [contextLiveNumber boolValue];
+    }
+    
+    // GEOFENCES
+    id contextGeofencesObj = contextDict[@"geofences"];
```
jsani-radar: I know this is more so the case in Android, but for most of our objects, we have `initWithArray` or some variant that traverses through an array field and calls `initWithObject` on each one - maybe we should have that for geofences so we can reuse this elsewhere?

RadarSDK/RadarContext.h
corypisano: hm not sure what you mean, like if the backend has multiple places at the location?

RadarSDK/Radar.h
corypisano: hm I might just be misunderstanding as an obj-c rookie, but I think the idea is that the 2 different options that are exposed are for whether you want to use the user's location (prob more common), or if you want to get context for some other location
similar to the 2 trackOnce handlers here: 
https://github.com/radarlabs/radar-sdk-ios/blob/0e09f35c4cd35aec42603f150d6cf716770d277d/RadarSDK/Radar.h#L228-L243

RadarSDK/RadarContext.h
jsani-radar: the context response is only going to show the top ranked place - we're not supporting `alternatePlaces` just yet

RadarSDK/Radar.h
jsani-radar: sorry, I see how it was ambiguous - this was just a nit about the function name to make it shorter. Theoretically every API call's last argument is the completion handler, so we don't necessarily need to call it out in the function name. Although, it seems like any function called without any arguments follows this convention, so I'm fine with it as is.

jsani-radar: it would then just be `trackOnce` vs. `trackOnceWithLocation`, anyways..

corypisano: ahhh gotcha, yeah i'll defer to others on the naming convention here :)

jsani-radar: it's good as is

PR #56, opened 2020-02-07
added commuting events

asau17:


RadarSDK/RadarUserInsightsState.m
```diff
@@ -30,15 +31,17 @@ - (instancetype _Nullable)initWithObject:(id)object {
     id userInsightsStateHomeObj = userInsightsStateDict[@"home"];
     id userInsightsStateOfficeObj = userInsightsStateDict[@"office"];
     id userInsightsStateTravelingObj = userInsightsStateDict[@"traveling"];
-    if (!userInsightsStateHomeObj || ![userInsightsStateHomeObj isKindOfClass:[NSNumber class]] || !userInsightsStateOfficeObj || ![userInsightsStateOfficeObj isKindOfClass:[NSNumber class]] || !userInsightsStateTravelingObj || ![userInsightsStateTravelingObj isKindOfClass:[NSNumber class]]) {
+    id userInsightsStateCommutingObj = userInsightsStateDict[@"commuting"];
+    if (!userInsightsStateHomeObj || ![userInsightsStateHomeObj isKindOfClass:[NSNumber class]] || !userInsightsStateOfficeObj || ![userInsightsStateOfficeObj isKindOfClass:[NSNumber class]] || !userInsightsStateTravelingObj || ![userInsightsStateTravelingObj isKindOfClass:[NSNumber class]]|| !userInsightsStateCommutingObj || ![userInsightsStateCommutingObj isKindOfClass:[NSNumber class]]) {
```
nickpatrick: @asau17 With this condition, we can't merge until the commuting changes are shipped to prod

nickpatrick:
@asau17 Looks great! Only one comment. We should ship commuting changes to prod before merging this

RadarSDK/RadarUserInsightsState.m
asau17: you're talking about the commuting changes in the SDK right?

asau17: the commuting changes in server?

nickpatrick: Yep. Until those changes are shipped, `userInsightsStateCommutingObj` will be `nil` and we will return `nil` here.

Alternatively, we can start with `userInsightsStateCommuting = NO` and skip this check.

asau17: I just changed it to `userInsightsStateCommuting = NO`.  Currently the commuting stuff is in server already but the state gets sanitized away unless the project settings have `CommuteInsights` set to `true`.

jsani-radar: makes sense, but I assume not all projects that have insights on will necessarily want commute insights as well, so we should account for the case where the key might not be present

asau17: Yeah, thinking a bit about this, even if users don't have commute Insights/insights on, the state will still be assigned. For now, as `commuteInsights` get sanitized out without the project setting on, it makes sense to keep it as `NO`. Once we ship commutes as a permanent feature we can revisit and just add the above clause back

PR #61, opened 2020-02-28
add clang-format

pingxia0823:
https://app.clubhouse.io/radarlabs/story/1646/add-clang-format-to-ios-sdk

Couple of changes here:
1. use the stripe `.clang-format` preset, with minor edits of setting `BreakBeforeBraces: Linux`
2. create a git pre-commit hook, so every time we do `git commit`, this hook will clang-format all changed files from the commit
```
clang_format'ing 1 files

COMMIT SUCCEEDED
[pingxia/ch1646/add-clang-format-to-ios-sdk 8bdf4ef] test
 1 file changed, 1 insertion(+), 1 deletion(-)
```
3. create a make command to setup that git pre-commit hook (e.g. configure the hooksPath, install clang-format CLI if not existent locally)

4. clang-format all objective C files in our repo


.githooks/pre-commit
```diff
@@ -0,0 +1,33 @@
+#!/bin/bash
```
jsani-radar: have you made this file executable with `chmod +x` as well? I think git hooks by default aren't

RadarSDK/Radar.h
```diff
@@ -118,260 +118,262 @@ typedef NS_ENUM(NSInteger, RadarRouteUnits) {
...
- 
+
  @param publishableKey Your publishable API key.
- 
+
  @see https://radar.io/documentation/sdk#ios-initialize
  */
-+ (void)initializeWithPublishableKey:(NSString * _Nonnull)publishableKey
-    NS_SWIFT_NAME(initialize(publishableKey:));
++ (void)initializeWithPublishableKey:(NSString *_Nonnull)publishableKey NS_SWIFT_NAME(initialize(publishableKey:));
```
jsani-radar: what're your thoughts on having `NS_SWIFT_NAME` on the same line as the last param? I kinda prefer it on a new line - is there a setting we can enable to do that?

RadarSDK/Radar.m
```diff
@@ -8,340 +8,420 @@
...
 #import "RadarLocationManager.h"
+#import "RadarLogger.h"
 #import "RadarSettings.h"
 #import "RadarState.h"
 
 @implementation Radar
 
-+ (id)sharedInstance {
++ (id)sharedInstance
+{
```
jsani-radar: this seems to be breaking before the brace? 🤔

clang_format_init.sh
```diff
@@ -0,0 +1,11 @@
+#!/bin/bash
```
jsani-radar: I see this goes and gets the package needed to run `clang-format`, but do we also want to have a `Makefile` incantation to do the formatting, so we can run it ad-hoc without a commit?

.githooks/pre-commit
pingxia0823: yeap

(.venv) ➜  radar-sdk-ios git:(pingxia/ch1646/add-clang-format-to-ios-sdk) ls -l .githooks/*
-rwxr-xr-x  1 pingxia  staff  742 Mar  1 22:33 .githooks/pre-commit

RadarSDK/Radar.m
pingxia0823: yeah, I did a minor change to the stripe preset, to break before the brace for function and classes 
```
use the stripe .clang-format preset, with minor edits of setting BreakBeforeBraces: Linux
```

clang_format_init.sh
pingxia0823: Yeah, so I added two incantations
1. clang-format-init: configure the hooksPath, install clang-format CLI if not existent locally
2. format: to format files ad-hoc without a commit

Makefile
```diff
@@ -33,6 +40,6 @@ build-pretty:
 docs:
 	jazzy
 
-dist: clean-pretty test-pretty build-pretty docs lint
+dist: clean-pretty test-pretty build-pretty docs lint format
```
pingxia0823: Is this `dist` command used by CI for release branch?

nickpatrick: Right now we run it manually before releasing, but we should add it to CI at some point

nickpatrick: We may want to run `format` before `test-pretty` and `build-pretty` so we ensure that formatting doesn't break tests or builds

clang_format_init.sh
nickpatrick: Rather than creating a separate script, thinking we should just add this to `bootstrap.sh` with the rest of the building and formatting dependencies. That way a new dev can run `bootstrap.sh` and have everything they need.

My preference is to remove the pre-commit hook so commits are predictable and just add the `format` target to the `Makefile`

RadarSDK/Radar.m
nickpatrick: My vote would be to use `Attach` instead of `Linux`. I know Apple breaks after function and class definitions, but I feel like it's antiquated, adds extra lines, and doesn't really make the code any more reasonable. I don't think it's a coincidence that they changed this convention in Swift. But let me know if you feel strongly about this.

RadarSDK/Radar.h
pingxia0823: hmm I don't see clang-format has any style options for that specifically https://clang.llvm.org/docs/ClangFormatStyleOptions.html

clang_format.sh
```diff
@@ -0,0 +1,8 @@
+#!/bin/bash
+
+echo "start formatting files ..."
+for file in `find ./ -type f -regex '.*\.[h|m]$'`; do
+    clang-format -i ${file} --verbose || exit 1
```
jsani-radar: nit: do we want to make sure that `clang_format` exists here? in case someone hasn't called `./bootstrap` yet?

RadarSDK/Radar.h
jsani-radar: yeah, no worries, I've seen differing standards, even on SO / iOS Documentation, should be fine as is - https://developer.apple.com/documentation/swift/objective-c_and_c_code_customization/renaming_objective-c_apis_for_swift

clang_format.sh
pingxia0823: I tried once locally, basically the script will exit 1 with message like "clang-format not found". I guess we could really beef up our README, to give instructions there.

PR #64, opened 2020-03-10
support splitting "name" into "addressLabel" and "placeLabel"

kochis:


pingxia0823:
just saw the linked server change. got it

jsani-radar:
@pingxia0823 -- it's so we can better segment the results we return for projects that have places enabled / disabled, we don't want to be returning place records for `project.settings.places = false`

kochis:
I ran `make test-pretty` and it looks like that builds the binaries. Should those not be checked in?

pingxia0823:
the `/dist/` is a bit annoying. It really shall be git ignored, just like the `server` repo

jsani-radar:
Didn't realize `make test-pretty` did that 😕- although we can't completely ignore `dist/` because we need to do so when we actually release, this is a topic of convo that's come up before, would like to streamline the SDK release process as a result

PR #65, opened 2020-03-18
support address in IP Geocode response

kochis:


RadarSDK/RadarAPIClient.m
```diff
@@ -548,11 +547,10 @@ - (void)ipGeocodeWithCompletionHandler:(RadarIPGeocodeAPICompletionHandler)compl
                           return completionHandler(status, nil, nil);
                       }
 
-                      id countryObj = res[@"country"];
-                      RadarRegion *country = [[RadarRegion alloc] initWithObject:countryObj];
+                      RadarAddress *address = [[RadarAddress alloc] initWithObject:res];
 
-                      if (country) {
-                          return completionHandler(RadarStatusSuccess, res, country);
+                      if (address) {
```
nickpatrick: @kochis Thanks for pulling this together. I think you'll need to pick the `address` dict out of the response like so:

```
id addressObj = res[@"address"];
RadarAddress *address = [[RadarAddress alloc] initWithObject:addressObj];
```

RadarSDKTests/Resources/geocode_ip.json
```diff
@@ -1,10 +1,28 @@
 {
```
jsani-radar: is this payload not keyed with the `address` field?

jsani-radar: another thought - do we need this `.json` file anymore? If the response is consistent with the other geocoding endpoints, we might be able to use `geocode.json` as a fixture everywhere

RadarSDK/RadarAPIClient.m
kochis: I don't think we're sending down a top-level address response

jsani-radar: I think we should have it for consistency with the other endpoints - https://github.com/radarlabs/server/pull/2157

RadarSDK/RadarRoutes.m
```diff
@@ -42,7 +42,7 @@ - (nullable instancetype)initWithObject:(nullable id)object {
 
     id geodesicObj = dict[@"geodesic"];
     if (geodesicObj) {
-        geodesic = [[RadarRouteDistance alloc] initWithObject:geodesicObj];
+        geodesic = [[RadarRouteDistance alloc] initWithObject:geodesicObj[@"distance"]];
```
nickpatrick: @kochis We probably want to guard against `geodesicObj[@"distance"]` being `nil` since we don't  guard for that in `RadarRouteDistance` `initWithObject:`.

```
id geodesicDistanceObj = geodesicObj[@"distance"];
if (geodesicDistanceObj) {
  geodesic = [[RadarRouteDistance alloc] initWithObject:geodesicDistanceObj];
}
```

A little excessive but want to make sure the deserialization code never crashes.

RadarSDKTests/Resources/geocode_ip.json
nickpatrick: I think we do since IP returns a single address and forward/reverse return an array

nickpatrick:
@kochis Can you update docs here as well? https://github.com/radarlabs/radar-sdk-ios/blob/master/RadarSDK/Radar.h#L520

`Geocodes the device's current IP address, converting IP address to partial address.`

RadarSDK/Radar.h
```diff
@@ -176,11 +176,11 @@ typedef void (^_Nonnull RadarSearchPointsCompletionHandler)(RadarStatus status,
 typedef void (^_Nonnull RadarGeocodeCompletionHandler)(RadarStatus status, NSArray<RadarAddress *> *_Nullable addresses);
 
 /**
- Called when an IP geocoding request succeeds, fails, or times out. Receives the request status and, if successful, the geocoding result (a country).
+ Called when an IP geocoding request succeeds, fails, or times out. Receives the request status and, if successful, the geocoding result (an address).
```
nickpatrick: @kochis Can you say `(a partial address)` to match the API docs?

PR #72, opened 2020-03-31
RC

nickpatrick:


RadarSDK/RadarAddress.m
```diff
@@ -225,6 +225,19 @@ - (instancetype _Nullable)initWithObject:(id _Nonnull)object {
     return arr;
 }
 
++ (NSString *)stringForConfidence:(RadarAddressConfidence)confidence {
```
jsani-radar: let's have the vanilla serializer use this method as well in addition to the one RN uses? https://github.com/radarlabs/radar-sdk-ios/blob/a7d1b722668f84882dddbf97295fae337bb9d2cf/RadarSDK/RadarAddress.m#L188-L195

would be good to centralize so we don't open ourselves up to potential drift

nickpatrick: The one RN uses goes from `RadarAddressConfidence` to `string`, this logic goes from `string` to `RadarAddressConfidence` (we follow a similar pattern for serialization/deserializing type in `RadarEvent`). In the interest of getting v3 out I think we can punt on this for now, then revisit serialization/deserialization code later. The fact that this is centralized in the native SDKs now too (instead of variations of this in RN/Capacitor/Cordova libs) is a step in the right direction.

jsani-radar: my fault, if it's an established pattern elsewhere, let's go with it

PR #74, opened 2020-04-30
[Beacon] support Beacon in getContext call

pingxia0823:
https://app.clubhouse.io/radarlabs/story/2044/beacon-ios-radar-getcontext


RadarSDK/RadarCollectionAdditions.m
```diff
@@ -44,26 +47,4 @@ - (NSArray *)radar_arrayForKey:(id)key {
     GET_DICT_VALUE_FOR_KEY(key, NSArray);
 }
 
-- (RadarCoordinate *)radar_coordinateForKey:(id)key {
```
pingxia0823, line 47: move into RadarCoordinate

RadarSDK/Radar.m
```diff
@@ -373,14 +394,18 @@ + (NSString *)stringForStatus:(RadarStatus)status {
...
+    case RadarStatusErrorNotFound:
+        str = @"ERROR_NOT_FOUND";
+        break;
     case RadarStatusErrorRateLimit:
         str = @"ERROR_RATE_LIMIT";
         break;
     case RadarStatusErrorServer:
         str = @"ERROR_SERVER";
         break;
-    default:
```
pingxia0823, line 382: remove default so the compiler could catch any unhandled switch case

pingxia0823:
@nickpatrick @jsani-radar this PR is ready to review.

RadarSDK/RadarJSONCoding/RadarBeacon+CLBeacon.m
```diff
@@ -0,0 +1,43 @@
...
+            NSString *secondComponent = [self.uuid substringWithRange:NSMakeRange(8, 4)];
+            NSString *thirdComponent = [self.uuid substringWithRange:NSMakeRange(12, 4)];
+            NSString *fourthComponent = [self.uuid substringWithRange:NSMakeRange(16, 4)];
+            NSString *lastComponent = [self.uuid substringWithRange:NSMakeRange(20, 12)];
+            NSString *formatedUUIDString = [NSString stringWithFormat:@"%@-%@-%@-%@-%@", firstComponent, secondComponent, thirdComponent, fourthComponent, lastComponent];
+            return [[NSUUID alloc] initWithUUIDString:formatedUUIDString];
+        } @catch (NSException *exception) {
+            return nil;
+        }
+    } else {
```
jsani-radar, line 36: do we want to do other validation here? like `length == 36`? There are a few strings that could fall through here that would be invalid

RadarSDK/Radar.h
```diff
@@ -272,6 +282,13 @@ typedef void (^_Nonnull RadarRouteCompletionHandler)(RadarStatus status, RadarRo
  */
 + (void)setAdIdEnabled:(BOOL)enabled;
 
+/**
+Enables beacon tracking
+
+@param enabled A boolean indicating whether beacons should be monitored.
+*/
++ (void)setBeaconEnabled:(BOOL)enabled;
```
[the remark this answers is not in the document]
pingxia0823: per product spec, this will be a client side flag to turn on / off beacon feature, irrespctive of the server side switch

RadarSDK/RadarAPIClient.m
```diff
@@ -273,11 +279,31 @@ - (void)getContextForLocation:(CLLocation *_Nonnull)location completionHandler:(
...
+                            return completionHandler(RadarStatusSuccess, res, context);
+                        }
+                        NSArray<RadarBeacon *> *beaconsToMonitor = [RadarBeacon fromObjectArray:beaconObj];
+                        if (!beaconsToMonitor) {
+                            // deserialization error
+                            return completionHandler(RadarStatusErrorServer, res, context);
+                        } else {
+                            [[RadarBeaconManager sharedInstance] trackOnceForRadarBeacons:beaconsToMonitor
+                                                                          completionBlock:^(RadarStatus status, NSArray<RadarBeacon *> *_Nullable nearbyBeacons) {
+                                                                              if (status != RadarStatusSuccess) {
```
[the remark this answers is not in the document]
pingxia0823: We typically do `completionHandler(status, nil, nil)` when failure happens. However, for beacon detections I guess it's better to _still_ return `res & context` even when beacon detection fails (e.g. timeout error, or errors from core location system lib). At the same time, we could log the beacon error as a warning.

RadarSDK/RadarJSONCoding/RadarBeacon+CLBeacon.m
pingxia0823: we do the other validation in the server. https://github.com/radarlabs/server/blob/master/server/utils/AppleBeaconUtils.ts#L72

Basically after server validation, the uuid string must be either 1) a 32 bit UUID without hyphens or 2) a 36 bit UUID with hyphens. The logic here is just to converting scenario 1 to 2 (since this is a system requirement from Core location API). We also wrapped this inside a try-catch for extra safety

nickpatrick:
@pingxia0823 Closing in favor of #99 (let's keep this around for reference)

PR #77, opened 2020-05-08
Mock tracking

nickpatrick:
https://app.clubhouse.io/radarlabs/story/2605/ios-mock-tracking

```swift
Radar.mockTracking(
  origin: origin,
  destination: destination,
  mode: .car,
  points: 10,
  interval: 2
) { (status, location, events, user) in
  // ...
}
```


nickpatrick:
@blackmad FYI

RadarSDK/RadarAPIClient.h
```diff
@@ -90,9 +93,15 @@ typedef void (^_Nullable RadarRouteAPICompletionHandler)(RadarStatus status, NSD
 
 - (void)getDistanceFromOrigin:(CLLocation *_Nonnull)origin
                   destination:(CLLocation *_Nonnull)destination
-                        modes:(RadarRouteMode)mode
+                        modes:(RadarRouteMode)modes
```
pingxia0823, line 96: nit:  maybe rename it to `mode:(RadarRouteMode)mode` to be consistent with the naming in the new API

RadarSDK/Radar.m
```diff
@@ -111,6 +114,58 @@ + (void)startTrackingWithOptions:(RadarTrackingOptions *)options {
...
+                                             intervalLimit = 60;
+                                         }
+
+                                         for (int i = 0; i < coordinates.count; i++) {
+                                             RadarCoordinate *coordinate = coordinates[i];
+                                             __block CLLocation *location = [[CLLocation alloc] initWithCoordinate:coordinate.coordinate
+                                                                                                          altitude:-1
+                                                                                                horizontalAccuracy:5
+                                                                                                  verticalAccuracy:-1
+                                                                                                         timestamp:[NSDate new]];
```
pingxia0823, line 149: nit: do we need to give caller more flexibility to specify these additional params (we could use these values as a preset of course)?

RadarSDK/RadarCoordinate.m
```diff
@@ -9,6 +9,63 @@
...
+        ![coordinatesLatitudeObj isKindOfClass:[NSNumber class]]) {
+        return nil;
+    }
+
+    NSNumber *coordinatesLongitudeNumber = (NSNumber *)coordinatesLongitudeObj;
+    NSNumber *coordinatesLatitudeNumber = (NSNumber *)coordinatesLatitudeObj;
+
+    float coordinatesLongitudeFloat = [coordinatesLongitudeNumber floatValue];
+    float coordinatesLatitudeFloat = [coordinatesLatitudeNumber floatValue];
+
```
pingxia0823, line 63: nit: there is a helper method `[NSDictionary radar_coordinateForKey]` in `RadarCollectionsAddition`, which does a bit more thorough checking on coordinate fields from server response

RadarSDK/Radar.m
nickpatrick: You mean `altitude`, `horizontalAccuracy`, and `verticalAccuracy`? I could see the server returning `altitude`, but most customers probably won't need it. The iOS simulator always uses `horizontalAccuracy = 5` for locations in the simulator, just following the same convention here.

RadarSDK/RadarAPIClient.h
nickpatrick: Distance API should take `modes` (can return routes for multiple modes, comma-separated), whereas mock API should take `mode` (single mode). The param is called `modes` in @blackmad's mock API PR but we should probably change to `mode` before releasing.

RadarSDK/RadarCoordinate.m
nickpatrick: Good call, will use that here

RadarSDK/Radar.h
```diff
@@ -304,9 +318,27 @@ typedef void (^_Nonnull RadarRouteCompletionHandler)(RadarStatus status, RadarRo
...
+ @param mode The travel mode.
+ @param points The number of mock location updates.
+ @param interval The interval in seconds between each mock location update. A number between 2 and 60.
+
+ @see https://radar.io/documentation/sdk#ios-mock
+ */
++ (void)mockTrackingWithOrigin:(CLLocation *_Nonnull)origin
+                   destination:(CLLocation *_Nonnull)destination
+                          mode:(RadarRouteMode)mode
+                        points:(int)points
```
jsani-radar, line 338: had the same comment on Android - `numPoints`?

PR #82, opened 2020-07-09
Only start updates if background permissions granted or showBlueBar is true

nickpatrick:


RadarSDK/RadarLocationManager.m
```diff
@@ -287,12 +287,13 @@ - (void)updateTracking:(CLLocation *)location {
             if (@available(iOS 11.0, *)) {
                 self.lowPowerLocationManager.showsBackgroundLocationIndicator = options.showBlueBar;
             }
-
+            
+            BOOL startUpdates = options.showBlueBar || [CLLocationManager authorizationStatus] == kCLAuthorizationStatusAuthorizedAlways;
```
jsani-radar, line 291: I assume the top-level condition (`if (tracking)`) here is only true for background mode, correct?

nickpatrick: It's true if you've called `Radar.startTracking()`: https://github.com/radarlabs/radar-sdk-ios/blob/master/RadarSDK/RadarLocationManager.m#L167

PR #84, opened 2020-07-27
move all completion block to main thread

pingxia0823:
https://app.clubhouse.io/radarlabs/story/3713/calling-all-completion-blocks-on-main-thread

The context for this PR is to call all client passed completion blocks in the main thread (a practice learned from Strip iOS SDK). The benefit is to make all our user-facing APIs thread safe.

### DIFF
--- a/RadarSDK/RadarUtils.m
+++ b/RadarSDK/RadarUtils.m
@@ -105,6 +105,22 @@
     return latitudeValid && longitudeValid && horizontalAccuracyValid;
 }
 
+#pragma mark - threading
++ (void)runOnMainThreadAsyncIfNecessary:(dispatch_block_t)block {
+    if (!block) {
+        return;
+    }
+
+    if ([NSThread isMainThread]) {
+        block();
+    } else {
+        dispatch_async(dispatch_get_main_queue(), ^{
+            block();
+        });
+    }
+    return;
+}
+
 #pragma mark - Airship integration
 
 #pragma clang diagnostic push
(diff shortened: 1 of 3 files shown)


RadarSDK/RadarUtils.m
```diff
@@ -105,6 +105,22 @@ + (BOOL)validLocation:(CLLocation *)location {
     return latitudeValid && longitudeValid && horizontalAccuracyValid;
 }
 
+#pragma mark - threading
++ (void)runOnMainThreadAsyncIfNecessary:(dispatch_block_t)block {
```
jsani-radar, line 109: since we're always invoking the callback on the main thread, should be fine to call this fn `runOnMainThreadAsync`

pingxia0823: These blocks could be invoked from either 1) main thread, or 2) the networking thread, or 3) the thread of `Radar` API callers. If they are invoked from the main thread, we don't want to do another main thread dispatch.

pingxia0823:
> this looks good!
> 
> I know the motivation behind this change is for thread safety, but also the added benefit that UI changes in our customer's app are guaranteed to be reflected, right? I vaguely remember that changing anything display-related needs to happen on the main thread

yeah, any UI change needs to be done in main thread. so I guess you mean the caller doesn't need to do dispatch_main within their completion blocks any more? yeah, that's some convenience for them too, since I guess our callback will likely trigger their app's UI to change

nickpatrick:
This is awesome!

PR #86, opened 2020-08-03
migrate two blocks previously missed

pingxia0823:
https://app.clubhouse.io/radarlabs/story/3774/move-two-more-completion-invocations-to-main-thread

missed two completion blocks in my previous PR https://github.com/radarlabs/radar-sdk-ios/pull/84.

### DIFF
--- a/RadarSDK/Radar.m
+++ b/RadarSDK/Radar.m
@@ -91,7 +91,6 @@
                     completionHandler(status, nil, nil, nil);
                 }];
             }
-
             return;
         }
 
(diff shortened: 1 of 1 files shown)


RadarSDK/Radar.m
```diff
@@ -252,8 +251,11 @@ + (void)getContextWithCompletionHandler:(RadarContextCompletionHandler)completio
     [[RadarLocationManager sharedInstance] getLocationWithCompletionHandler:^(RadarStatus status, CLLocation *_Nullable location, BOOL stopped) {
         if (status != RadarStatusSuccess) {
             if (completionHandler) {
-                return completionHandler(status, nil, nil);
+                [RadarUtils runOnMainThreadAsyncIfNecessary:^{
+                    completionHandler(status, nil, nil);
+                }];
             }
+            return;
```
jsani-radar, line 258: out of curiosity, why do the `getLocation` completion handlers have a `return`, but the main ones with the API result do not?

pingxia0823: just an early return, to prevent calling `RadarAPIClient` when the `getLocation` errors out

PR #91, opened 2020-10-16
Prototype syncGeofences tracking option

nickpatrick:
Calls the geofence search API to sync the nearest 10 geofences to the client for improved responsiveness


RadarSDK/RadarLocationManager.m
```diff
@@ -329,23 +332,47 @@ - (void)updateTracking:(CLLocation *)location {
             }
         } else {
             [self stopUpdates];
-            [self removeGeofences];
+            [self removeMovingStoppedGeofence];
```
tjulien, line 335: we may want to also do something like `stopMonitoringForSyncedGeofences`

RadarSDK/RadarLocationManager.m
```diff
@@ -329,23 +332,47 @@ - (void)updateTracking:(CLLocation *)location {
...
             [self.locationManager stopMonitoringVisits];
             [self.locationManager stopMonitoringSignificantLocationChanges];
         }
     });
 }
 
-- (void)replaceGeofence:(CLLocation *)location radius:(int)radius {
-    [self removeGeofences];
+- (void)replaceSyncedGeofences:(NSArray<RadarGeofence *> *)geofences {
+    for (int i = 0; i < geofences.count; i++) {
```
tjulien, line 343: We may want to clear out the old synced geofences here

nickpatrick: Yes, good call

PR #101, opened 2021-01-18
v3.1

nickpatrick:
- Beacon support
- Opt-in and session stats
- New defaults (`sync = ALL`, `syncGeofences = true`)
- Remove points
- PR looks bigger than it is because of docs


tjulien:
https://app.clubhouse.io/radarlabs/story/5618/take-the-beta-tag-off-of-3-1

RadarSDK/Radar.m
```diff
@@ -663,4 +687,15 @@ + (NSDictionary *)dictionaryForLocation:(CLLocation *)location {
...
 
+- (void)applicationWillEnterForeground {
+    BOOL updated = [RadarSettings updateSessionId];
+    if (updated) {
+        [[RadarAPIClient sharedInstance] getConfig];
+    }
+}
+
+- (void)dealloc {
+    [[NSNotificationCenter defaultCenter] removeObserver:self];
```
tjulien, line 698: I guess there's an edge case where radar would not have added itself as an observer if `initializeWithPublishableKey` is never called.  I'm assuming `removeObserver` can handle that case.

RadarSDK/RadarLocationManager.m
```diff
@@ -192,8 +195,6 @@ - (void)startUpdates:(int)interval {
                                                        block:^(NSTimer *_Nonnull timer) {
                                                            [[RadarLogger sharedInstance] logWithLevel:RadarLogLevelDebug message:@"Timer fired"];
 
-                                                           [[RadarBackgroundTaskManager sharedInstance] startBackgroundTask];
```
tjulien, line 195: I know we're removing this to cleanup log messages that are coming out, how confident are we that this is no longer needed for the purpose of get location updates?

RadarSDK/RadarLocationManager.m
```diff
@@ -26,13 +26,16 @@ @interface RadarLocationManager ()
...
+@property (nonnull, strong, nonatomic) NSMutableSet<NSString *> *nearbyBeaconIdentifers;
 
 @end
 
 @implementation RadarLocationManager
 
-static NSString *const kRegionIdentifer = @"radar";
-static NSString *const kRegionSyncIdentifer = @"radar_sync";
+static NSString *const kIdentifierPrefix = @"radar_";
+static NSString *const kBubbleGeofenceIdentifierPrefix = @"radar_bubble_";
```
tjulien, line 36: any one-time upgrade issues that might lead to orphaning regions with old identifiers that are being monitored?

RadarSDK/RadarLocationManager.m
```diff
@@ -372,33 +401,53 @@ - (void)replaceSyncedGeofences:(NSArray<RadarGeofence *> *)geofences {
...
+- (void)replaceSyncedBeacons:(NSArray<RadarBeacon *> *)beacons {
+    [self removeSyncedBeacons];
+
+    BOOL tracking = [RadarSettings tracking];
+    RadarTrackingOptions *options = [RadarSettings trackingOptions];
+    if (!tracking || !options.beacons || !beacons) {
+        return;
+    }
+
+    NSUInteger numBeacons = MIN(beacons.count, 9);
```
tjulien, line 421: Not a big deal, but the `searchBeacons` call is passing `limit = 10`, but we're capping it at 9 here.

RadarSDK/RadarSettings.m
```diff
@@ -120,8 +139,13 @@ + (void)setConfig:(NSDictionary *)config {
 }
 
 + (RadarLogLevel)logLevel {
-    NSInteger logLevelInteger = [[NSUserDefaults standardUserDefaults] integerForKey:kLogLevel];
-    return (RadarLogLevel)logLevelInteger;
+    RadarLogLevel logLevel;
+    if ([[NSUserDefaults standardUserDefaults] objectForKey:kLogLevel] == nil) {
+        logLevel = RadarLogLevelInfo;
```
tjulien, line 144: this makes the default log level info?  Is there any risk of having this on in production, ie taking up space on the device in system logs?

RadarSDK/RadarUtils.m
```diff
@@ -69,6 +69,37 @@ + (BOOL)locationBackgroundMode {
...
+    CLAuthorizationStatus authorizationStatus = CLLocationManager.authorizationStatus;
+    switch (authorizationStatus) {
+    case kCLAuthorizationStatusAuthorizedWhenInUse:
+        return @"GRANTED_FOREGROUND";
+    case kCLAuthorizationStatusAuthorizedAlways:
+        return @"GRANTED_BACKGROUND";
+    case kCLAuthorizationStatusDenied:
+        return @"DENIED";
+    case kCLAuthorizationStatusRestricted:
+        return @"DENIED";
```
tjulien, line 82: Should this be `@"RESTRICTED"`?

nickpatrick: We'll just treat restricted as `DENIED` so we don't have to support a separate enum val on the server. It's sort of a rare state that means "denied with parental controls turned on" https://developer.apple.com/documentation/corelocation/clauthorizationstatus/restricted

RadarSDK/RadarSettings.m
nickpatrick: Nope, should be fine. The only info-level logs are for locations/events/errors (relatively low volume), whereas all other logs are debug-level. We could add a check for environment (`publishableKey` contains test) but I think it's overkill for v1

RadarSDK/RadarLocationManager.m
nickpatrick: I can switch if it reduces server load. We get 20 slots on iOS, need to save one for the bubble geofence, then split the rest between synced geofences and synced beacons

RadarSDK/RadarLocationManager.m
nickpatrick: All legacy regions are prefixed with `radar_`, which is the prefix we use when removing on `stopTracking()`, so we're good

RadarSDK/RadarLocationManager.m
nickpatrick: You used to need to start background tasks to get (and keep extending) 180 seconds of background runtime, but as long as you're also constantly running a location manager, this seemingly stopped being necessary somewhere around iOS 10 (not documented by Apple, just something I observed). What's worse, there's now a scary warning message if you run a background task longer than 30 seconds (which @jsani-radar noted when dogfooding). We left it in to be on the safe side, but I've been testing for a few days without it and I've never seen background time run out, so I think we're safe to remove

RadarSDK/Radar.m
nickpatrick: `Radar` is a singleton and unless you call a method (initialize or something else) it will never be initialized and `dealloc` will never be called. I'll squint at this to make sure, though

RadarSDK/RadarBeaconManager.m
```diff
@@ -0,0 +1,217 @@
...
+        _nearbyBeaconIdentifers = [NSMutableSet new];
+        _failedBeaconIdentifiers = [NSMutableSet new];
+
+        _permissionsHelper = [RadarPermissionsHelper new];
+    }
+    return self;
+}
+
+- (void)callCompletionHandlersWithStatus:(RadarStatus)status beacons:(NSArray<NSString *> *_Nullable)nearbyBeacons {
+    @synchronized(self) {
```
tjulien, line 61: Probably worth adding comment here or somewhere in this class explaining why `synchronized` needed.

nickpatrick: 👍 It's to ensure that we don't try to remove a callback due to a timeout while we're looping through them to call them (same pattern as in `RadarLocationManager`)

RadarSDK/RadarLocationManager.m
tjulien: Probably worth a comment here explaining why it's being limited and where `9` comes from.

RadarSDK/RadarLocationManager.m
```diff
@@ -578,11 +629,35 @@ - (void)sendLocation:(CLLocation *)location stopped:(BOOL)stopped source:(RadarL
...
+    RadarTrackingOptions *options = [RadarSettings trackingOptions];
+    if (options.beacons) {
+        nearbyBeacons = [self.nearbyBeaconIdentifers allObjects];
+
+        [[RadarLogger sharedInstance] logWithLevel:RadarLogLevelDebug
+                                           message:[NSString stringWithFormat:@"Sending nearby beacons | nearbyBeacons = %@", [nearbyBeacons componentsJoinedByString:@","]]];
+
+        if (source != RadarLocationSourceBeaconEnter && source != RadarLocationSourceBeaconExit && source != RadarLocationSourceMockLocation &&
+            source != RadarLocationSourceManualLocation) {
+            [[RadarAPIClient sharedInstance] searchBeaconsNear:location
```
tjulien, line 642: probably worth a comment here explaining why we search for beacons at this point.

nickpatrick: May be worth formalizing best practices and a style guide for comments in our open-source SDKs, then having me come through through and comment everything in this class thoroughly and consistently in a separate PR (alternatively, documenting it even more thoroughly internally). Let's discuss offline this week

RadarSDK/RadarSettings.m
tjulien: Does the device delete these every so often of they do they take up space on the device forever?

nickpatrick: The OS handles it. There's a max buffer size. On Android, I believe the buffer is 4 MB by default. On iOS, not sure it's documented (we call `NSLog`, see https://developer.apple.com/documentation/foundation/1395074-nslogv). iOS apps print a ton of logs in production, so these will be like a drop of water in a 10 gallon bucket. Not worried about it. We'll also document `Radar.setLogLevel()` in case people want to turn this off.

tjulien (approved):

PR #115, opened 2021-05-12
Add callbacks to trip update functions

nickpatrick:


RadarSDK/Radar.m
```diff
@@ -303,28 +303,52 @@ + (RadarTripOptions *)getTripOptions {
 }
 
 + (void)startTripWithOptions:(RadarTripOptions *)options {
-    [RadarSettings setTripOptions:options];
-    [[RadarAPIClient sharedInstance] updateTripWithOptions:options status:RadarTripStatusStarted];
-    [[RadarLocationManager sharedInstance] getLocationWithCompletionHandler:nil];
```
nickpatrick, line 308: No need to "flush" an upsert now that we're calling `PATCH` directly

nickpatrick:
@tjulien Ready for review so we can release an SDK update. These changes already went out in toolkit.

tjulien:
https://app.clubhouse.io/radarlabs/story/6413/add-sdk-callbacks-to-trip-update-functions

RadarSDK/Radar.m
```diff
@@ -303,28 +303,78 @@ + (RadarTripOptions *)getTripOptions {
...
-+ (void)updateTripWithOptions:(RadarTripOptions *)options {
-    [[RadarAPIClient sharedInstance] updateTripWithOptions:options status:RadarTripStatusUnknown];
++ (void)startTripWithOptions:(RadarTripOptions *)options completionHandler:(RadarTripCompletionHandler)completionHandler {
+    [[RadarAPIClient sharedInstance] updateTripWithOptions:options
+                                                    status:RadarTripStatusStarted
+                                         completionHandler:^(RadarStatus status) {
+                                             if (status == RadarStatusSuccess) {
+                                                 [RadarSettings setTripOptions:options];
+
+                                                 [[RadarLocationManager sharedInstance] getLocationWithCompletionHandler:nil];
```
tjulien, line 316: same as the android PR, can we add a comment here explaining why getting a location is needed here, but not in the other `updateTrip` cases

nickpatrick: Thinking about it more, it's actually probably not a bad idea to flush a location update to the server in all cases. Will add and test.

nickpatrick: Done

tjulien (approved):

PR #119, opened 2021-07-13
dont trust CLVisits as stops

tjulien:
@nickpatrick and I were observing that quite often `CLVisit`s don't represent actual stops.  Useful as a location, but not a great signal for a real stop

### DIFF
--- a/RadarSDK/RadarLocationManager.m
+++ b/RadarSDK/RadarLocationManager.m
@@ -518,7 +518,7 @@ static NSString *const kSyncBeaconIdentifierPrefix = @"radar_beacon_";
             if (duration == 0) {
                 duration = -[location.timestamp timeIntervalSinceNow];
             }
-            stopped = (distance <= options.stopDistance && duration >= options.stopDuration) || source == RadarLocationSourceVisitArrival;
+            stopped = distance <= options.stopDistance && duration >= options.stopDuration;
 
             [[RadarLogger sharedInstance]
                 logWithLevel:RadarLogLevelDebug
@@ -534,7 +534,7 @@ static NSString *const kSyncBeaconIdentifierPrefix = @"radar_beacon_";
             }
         }
     } else {
-        stopped = (force || source == RadarLocationSourceVisitArrival);
+        stopped = force;
     }
     BOOL justStopped = stopped && !wasStopped;
     [RadarState setStopped:stopped];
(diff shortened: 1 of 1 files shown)


RadarSDK/RadarLocationManager.m
```diff
@@ -518,7 +518,7 @@ - (void)handleLocation:(CLLocation *)location source:(RadarLocationSource)source
             if (duration == 0) {
                 duration = -[location.timestamp timeIntervalSinceNow];
             }
-            stopped = (distance <= options.stopDistance && duration >= options.stopDuration) || source == RadarLocationSourceVisitArrival;
+            stopped = distance <= options.stopDistance && duration >= options.stopDuration;
```
tjulien, line 521: should we add a tracking option `treatVistisAsStops` or just remove it like this? 🤔

nickpatrick: Considerations for leaving it:
- Although it can cause premature stops, in some cases it can also cause faster stops. Some people may want to leave it on for that reason.
- There could theoretically be some situations where location keeps bouncing around and the device never transitions to stopped, and a `CLVisit` is sort of a backstop for the OS to declare that the device is stopped.

That said, I'm inclined to keep it simple and remove it for now.

RadarSDK/RadarLocationManager.m
```diff
@@ -534,7 +534,7 @@ - (void)handleLocation:(CLLocation *)location source:(RadarLocationSource)source
             }
         }
     } else {
-        stopped = (force || source == RadarLocationSourceVisitArrival);
+        stopped = force;
```
nickpatrick, line 537: @tjulien We need to keep this case. This `else` is used by the `.efficient` preset which on iOS basically just says "rely completely on visits."

tjulien: fixed

RadarSDK/RadarLocationManager.m
tjulien: I could make a tracking option defaulted to false pretty easily.

`useVisitsAsStops = false`?

nickpatrick: I think we can just change the behavior and skip adding a new tracking option. Suggest we merge this in with the rest of the iOS 3.2 beta stuff and do a bunch of testing in a variety of situations (urban and rural areas) in the next few weeks.

RadarSDK/RadarLocationManager.m
jsani-radar: can we add a comment here about that case?

nickpatrick (approved):

PR #120, opened 2021-07-14
Guard against nil beacons

nickpatrick:


RadarSDK/RadarLocationManager.m
```diff
@@ -427,13 +427,20 @@ - (void)replaceSyncedBeacons:(NSArray<RadarBeacon *> *)beacons {
                                                                          major:[beacon.major intValue]
                                                                          minor:[beacon.minor intValue]
                                                                     identifier:identifier];
-        region.notifyEntryStateOnDisplay = YES;
-        [self.locationManager startMonitoringForRegion:region];
-        [self.locationManager requestStateForRegion:region];
+
```
tjulien, line 430: I assume `identifier` needs the extra `kSyncBeaconIdentifierPrefix` here since we need to distinguish between beacon regions used for ranging vs beacon regions used for basic monitoring.

Just double checking

(low priority, we may want to add a second `BeaconManager.regionForBeacon` function that handles this case to avoid accidental bugs in the future)

nickpatrick: Yup, it should be in there (a few lines above): https://github.com/radarlabs/radar-sdk-ios/pull/120/files/91df5e12693d17b3d41e6673721b206ead4119d9#diff-4689395a38be45a14ce3f56437a52efe930adc0d3cf9f644915693ab811efa04R425

tjulien (approved):

PR #121, opened 2021-08-02
add development instructions

tjulien:


tjulien:
https://app.clubhouse.io/radarlabs/story/6840/document-ios-and-android-development-process-in-the-readmes

DEVELOPMENT.md
```diff
@@ -0,0 +1,8 @@
+## Documentation on building the SDK
+To do development on the SDK source, use the following commands:
+
+```
+brew install cocoapods # if you do not have cocoapods installed
+pod install
```
david-goodfellow, line 6: I get the following in the project root folder on this line:
```
❯ pod install
[!] No `Podfile' found in the project directory.
```

david-goodfellow: actually this step was unneeded for me. Only step I had to do was the `xcpretty` was not installed so I had to run `sudo gem install xcpretty`. Was fairly straightforward to debug this so don't know if we need to explicitly list out the dependencies

nickpatrick:
Yeah, you shouldn't need to do a `pod install` to build the SDK. But we do a `pod lint` to make sure the `Podfile` is properly formatted as part of `make dist`.

There is a `make bootstrap` command that will run `bootstrap.sh` to install all dependencies, including `xcpretty` and `cocoapods`.

nickpatrick:
@tjulien Closing this and will rewrite in the 3.2 branch

PR #122, opened 2021-08-03
adding layers + country to autocomplete in ios sdk

david-goodfellow:


dist/RadarSDK.framework/Versions/A/Headers/RadarTrackingOptions.h
```diff
@@ -161,23 +161,23 @@ typedef NS_ENUM(NSInteger, RadarTrackingOptionsSync) {
 @property (nonatomic, assign) BOOL beacons;
 
 /**
- A preset that updates about every 30 seconds and syncs all locations to the server. High battery usage. Shows the flashing blue status bar when tracking.
```
david-goodfellow, line 164: a bit unsure about why this text changed within the `dist` folder. Wanted to bring this to attention to see which is correct before I reverted to what was in master for this file

nickpatrick: I'm guessing this was sitting in `master` but hadn't been released yet. These changes are correct, we should keep them in this release.

corypisano: it looks to me like this PR https://github.com/radarlabs/radar-sdk-ios/pull/118 changed the text, but didnt built. so dist is now being updated to reflect that change

RadarSDK/RadarAPIClient.m
```diff
@@ -503,6 +503,7 @@ - (void)autocompleteQuery:(NSString *)query near:(CLLocation *_Nonnull)near limi
     [queryString appendFormat:@"query=%@", query];
     [queryString appendFormat:@"&near=%.06f,%.06f", near.coordinate.latitude, near.coordinate.longitude];
     [queryString appendFormat:@"&limit=%d", finalLimit];
+    [queryString appendFormat:@"&tags=%@", [layers componentsJoinedByString:@","]];
```
nickpatrick, line 506: May need a `nil` check here

RadarSDK/Radar.m
```diff
@@ -530,9 +530,14 @@ + (void)searchGeofencesNear:(CLLocation *_Nonnull)near
...
 
-+ (void)autocompleteQuery:(NSString *)query near:(CLLocation *)near limit:(int)limit completionHandler:(RadarGeocodeCompletionHandler)completionHandler {
++ (void)autocompleteQuery:(NSString *_Nonnull)query
+                     near:(CLLocation *_Nonnull)near
+                     layers:(NSArray *_Nullable)layers
+                     limit:(int)limit
+                     completionHandler:(RadarGeocodeCompletionHandler)completionHandler {
     [[RadarAPIClient sharedInstance] autocompleteQuery:query
                                                   near:near
+                                                    layers:layers
```
nickpatrick, line 540: `make format` should fix the indentation here

RadarSDK/RadarAPIClient.m
david-goodfellow: nil check now there and using layers not `tag=` 🤦

nickpatrick (approved):

nickpatrick:
@david-goodfellow This looked great, just making a few changes before we ship:
- Adding a missing `&` before `country` when constructing the `GET /search/autocomplete` URL (was getting `ERROR_BAD_REQUEST` when passing `country`)
- Updating the sample code to use `layers = ["locality"]` instead of `layers = ["place"]` since we've talked about removing places from autocomplete
- Consolidating to one method signature for autocomplete in `RadarAPIClient` (can just pass `nil` for `layers` and `country` in `Radar.m` for the original autocomplete method)
- Removing the `country.length == 2` check. I think we'd rather pass through a bad `country` value and return `ERROR_BAD_REQUEST` than silently not pass `country` if invalid
- Bumping the version
- Running `make format` to format the code and `make dist` to build the final framework and docs

Releasing shortly.

PR #125, opened 2021-09-05
Beacon monitoring based on UUID only

nickpatrick:
- Supports monitoring and ranging beacons based on UUID only, regardless of lat/lng. If UUIDs are present in project settings, registers beacon regions based on UUID only and ranges upon entry or exit to get major + minor. If UUIDs are not present in project settings, registers beacon regions based on UUID + major + minor and does not range on entry or exit (old behavior).
- Sends full beacon structs to `POST /track` as `beacons = [{ uuid: string, major: string, minor: string, rssi: number?: proximity: number? }, …]`. We can pick up `rssi` and `proximity` later to set beacon event confidence based on signal strength.
- Moves a lot of beacon logic and state keeping from `RadarLocationManager` into `RadarBeaconManager`.
- `make format` polluted the PR, but most changes are in `RadarAPIClient`, `RadarLocationManager`, and `RadarBeaconManager`.


jrtibbetts (approved):

RadarSDK/Radar.m
```diff
@@ -125,50 +125,69 @@ + (void)trackOnceWithDesiredAccuracy:(RadarTrackingOptionsDesiredAccuracy)desire
...
-                                                                              callTrackAPI(nearbyBeacons);
-                                                                          }];
-                                                                  }];
-                                                              }];
+                             [[RadarAPIClient sharedInstance]
+                                 searchBeaconsNear:location
+                                            radius:1000
+                                             limit:10
+                                 completionHandler:^(RadarStatus status, NSDictionary *_Nullable res, NSArray<RadarBeacon *> *_Nullable beacons,
+                                                     NSArray<NSString *> *_Nullable beaconUUIDs) {
```
andlum, line 152: We should settle on a naming convention for this param at weekly today so we can test this; I'm using `projectBeaconUUIDs` and @kochis has thrown out `monitoredBeacons`

kochis: Yeah, now that we are returning two different types of beacon ids (_nearby_ beacons, a _project monitored_ beacons), it would be good have a naming convention that highlights the difference between them.

My suggestion was `nearbyBeacons` and `monitoredBeacons`, but open to other ideas

RadarSDK/Include/RadarTrackingOptions.h
```diff
@@ -46,21 +46,18 @@ typedef NS_ENUM(NSInteger, RadarTrackingOptionsSyncLocations) {
 
 /**
  An options class used to configure background tracking.
-
```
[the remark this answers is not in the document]
nickpatrick: Hmmm, wondering if `make format` did this. I can revert

PR #130, opened 2021-11-08
add support for CLLocationSourceInformation

tjulien:
Adds support for `CLLocationSourceInformation` and upgrades the build to support iOS 15.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -2,7 +2,7 @@ version: 2
 jobs:
   build_and_test:
     macos:
-      xcode: 12.2.0
+      xcode: 13.1.0
     steps:
       - checkout
       - run:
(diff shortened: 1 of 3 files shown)


nickpatrick (approved):
Nice! Only nit: I might move this up between `source` and `tripOptions` like on Android. Easier to compare how params are passed across iOS and Android when they're passed in the same order.

I noticed there's also a `isProducedByAccessory` flag. Docs are a little sparse, but should we also be looking at this?

tjulien:
@nickpatrick The only content I could find was in https://www.vwidtalk.com/threads/apple-carplay-and-android-auto-gain-electrify-america-app-support.3047/

Which says that it indicates when the location comes from CarPlay.

PR #133, opened 2021-12-10
7993: Get the version number from the bundle's properties

jrtibbetts:
[Shortcut 7993: [iOS] Get SDK version number from the MARKETING_VERSION variable](https://app.shortcut.com/radarlabs/story/7993/ios-get-sdk-version-number-from-the-marketing-version-variable)

* Replaced `RadarUtils.sdkVersion()` (a function) with `Radar.sdkVersion` (a property).
* Added a UI to the Example app that displays the SDK version.

Simulator Screen Shot - iPhone 12 Pro - 2021-12-10 at 16 14 09


RadarSDK/Radar.h
```diff
@@ -239,6 +239,8 @@ typedef void (^_Nonnull RadarRouteMatrixCompletionHandler)(RadarStatus status, R
  */
 @interface Radar : NSObject
 
+@property (class, readonly) NSString *sdkVersion;
```
nickpatrick, line 242: @jrtibbetts Love the idea of exposing this for customers to add to their own logging.

@jsani-radar @phil-brown Should we do the same on Android for consistency?

RadarSDK/RadarUtils.m
```diff
@@ -32,7 +32,14 @@ + (NSNumber *)timeZoneOffset {
 }
 
 + (NSString *)sdkVersion {
-    return @"3.2.2";
+    NSBundle *bundle = [NSBundle bundleForClass:RadarUtils.class];
+    NSString *versionNumber = bundle.infoDictionary[@"CFBundleShortVersionString"];
```
nickpatrick, line 36: @jrtibbetts Very cool. Will this always pull the marketing version from the SDK's plist? Will that still work if the SDK is embedded in a customer's app? (Just want to make extra sure we'd never pull the customer's app version instead.)

Side note: It may actually be useful to pull the customer's app version and send it up to `POST /track` as `appVersion`. This could help with debugging. (When I'm debugging Radar, how do I know which SDK version _and_ app version the user is running?

cc @phil-brown @jsani-radar to consider the same thing for Android.

RadarSDK/RadarUtils.m
```diff
@@ -32,7 +32,14 @@ + (NSNumber *)timeZoneOffset {
...
+    
+    if (versionNumber == nil) {
+        return @"(unknown version)";
```
nickpatrick, line 39: @jrtibbetts @jsani-radar Are we validating the format of `sdkVersion` on the server today? (e.g., `x.y.z`) If so, this may fail. Should we: (a) update the validation, if it exists, (b) send `0.0.0` or some dummy value instead, or (c) throw an exception or log an error here because this should never happen?

Example/Example/Main.storyboard
```diff
@@ -0,0 +1,50 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
nickpatrick, line 1: @jsani-radar @jrtibbetts @phil-brown Cool that we're leveling up the iOS example app. The iOS and Android example apps are starting to diverge a bit though. Do we care? Worth investing in keeping them in sync?

Gemfile
```diff
@@ -0,0 +1,3 @@
+source 'https://rubygems.org'
```
nickpatrick, line 1: @jrtibbetts Any downstream effects of taking a dependency on CocoaPods in our example app?

RadarSDK.xcodeproj/project.pbxproj
```diff
@@ -729,9 +730,10 @@
 		0107A9ED2621FFB9008AB52F /* Debug */ = {
 			isa = XCBuildConfiguration;
 			buildSettings = {
+				CODE_SIGN_IDENTITY = "Apple Development";
 				CODE_SIGN_STYLE = Manual;
 				DEFINES_MODULE = YES;
-				DEVELOPMENT_TEAM = "";
+				DEVELOPMENT_TEAM = 96GHH65B9D;
```
nickpatrick, line 736: @jrtibbetts Is this the Radar team ID? Does it matter?

RadarSDK/RadarUtils.m
jrtibbetts: If they're using the Radar SDK *as a framework*, it will always pull it from the Radar SDK framework. If they statically linked the Radar code into their app (i.e. without using a framework), it would be their app's version. Do we know whether any customers are using it that way?

Example/Example/Main.storyboard
jrtibbetts: If we want to keep this example app headless, maybe I can add the SDK version number to the toolkit instead.

Gemfile
jrtibbetts: This just sets a minimum CocoaPods version. I can take it out.

RadarSDK.xcodeproj/project.pbxproj
jrtibbetts: 👀

RadarSDK/RadarUtils.m
jsani-radar: I don't think we can guarantee whether customers statically link or use a framework to integrate Radar - we'll have to support / be resilient to both. Would it be possible to namespace it such that it always grabs the Radar SDK version?

RadarSDK/RadarUtils.m
jsani-radar: yes, we're doing validation of this field on the server, so this would fail - my preference would be to log an error so we get visibility in the logs feed + avoid crashing the app

jrtibbetts: I can log an error, but what value should it return?

RadarSDK/RadarUtils.m
phil-brown: FWIW this _is_ doable for Android.

Example/Example/Main.storyboard
phil-brown: I think this would be extremely useful in the toolkit app. I also think the example app should be upgraded at some point as well (this could be the test app we've talked about, for example)

PR #135, opened 2021-12-15
7990: Add a GitHub build & test action

jrtibbetts:
[Shortcut 7990: Automate iOS SDK releases using GitHub Actions](https://app.shortcut.com/radarlabs/story/7990/automate-ios-sdk-releases-using-github-actions)

This is a preliminary PR that cleans, builds, tests, and analyzes the `Framework` scheme whenever there's a push to `master` or to PR that targets `master`.

**Note: This is built on top of pr #133, so that one should be approved and merged before this one.**


.github/workflows/build-and-test.yml
```diff
@@ -0,0 +1,21 @@
+name: Xcode - Build, Analyze, & Test
```
phil-brown, line 1: Is there a benefit to having this run in addition to the existing circleci configuration?

.github/workflows/release-sdk.yml
```diff
@@ -0,0 +1,45 @@
...
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+
+      - name: Build, test, and analyze
+        run: xcodebuild clean build analyze test -workspace Example/Example.xcodeproj/project.xcworkspace -scheme RadarSDK -destination "platform=iOS Simulator,name=iPhone 11" | xcpretty
+      
+      - name: Build archive for iPhone simulator
+        run: xcodebuild archive -scheme RadarSDK -archivePath "RadarSDK-iphonesimulator.xcarchive" -sdk iphonesimulator SKIP_INSTALL=NO CODE_SIGN_IDENTITY= CODE_SIGNING_REQUIRED=NO
```
phil-brown, line 26: is CODE_SIGN_IDENTITY intentionally left blank for these?

.github/workflows/release-sdk.yml
```diff
@@ -0,0 +1,45 @@
...
+      - name: Zip XCFramework
+        run: zip -r RadarSDK.xcframework.zip RadarSDK.xcframework -x ".*" -x "__MACOSX" -D
+
+      - name: Generate SDK docs
+        run: jazzy
+
+      - name: Zip SDK docs
+        run: zip -r docs.zip docs -x ".*" -x "__MACOSX" -D
+      
+      - name: Lint CocoaPods spec
```
phil-brown, line 43: Would it make sense for the linter to run first, in case errors are found?

phil-brown (commented):
A few comments/questions

.github/workflows/release-sdk.yml
jrtibbetts: I copied this more or less as-is from the `Makefile` that Nick uses for manual builds.

.github/workflows/build-and-test.yml
jrtibbetts: I added it just to get a feel for how easy or hard it is to do it in GH Actions instead of CircleCI. _My_ preference is to do it in GH, if for no other reason than keeping the configuration in source control. I've also noticed that when I added Swift code to the SDK on a branch, the GH action succeeded, but CircleCI failed. 🤷

.github/workflows/release-sdk.yml
jrtibbetts: This step lints the CocoaPods spec file before we publish it; it's not a general linter like SwiftLint. I suppose it can be run earlier in the process, but again, I'm following the order in which `make dist` does things.

.github/workflows/build-and-test.yml
tjulien: > _My_ preference is to do it in GH

+1, if we can get this working well in GH, would be a nice simplification to run everything in GH Actions.

tjulien:
@jrtibbetts what from https://github.com/radarlabs/radar-sdk-ios/pull/133 is actually necessary to land this PR?  The reason I ask is it would be good to land this release PR this week, but #133 seems like there are still a number of open things to work through, and they may not be critical to ship releasing SDK's via CI.

Put differently - what testing should we do to get confident about merging this PR in this week?

.github/workflows/build-and-test.yml
phil-brown: Might also be a good time to add a status badge for the master branch https://docs.github.com/en/actions/monitoring-and-troubleshooting-workflows/adding-a-workflow-status-badge

tjulien (approved):

.github/workflows/build-and-test.yml
jrtibbetts: Not a bad idea, but let's do it after this is merged.

phil-brown (approved):

tjulien:
https://app.shortcut.com/radarlabs/story/7990/automate-ios-sdk-releases-using-github-actions

PR #137, opened 2021-12-22
iOS SDK - device logs

corypisano:
Story details: https://app.shortcut.com/radarlabs/story/7855


RadarSDK/Radar.h
```diff
@@ -803,6 +813,13 @@ typedef void (^_Nonnull RadarRouteMatrixCompletionHandler)(RadarStatus status, R
  */
 + (NSDictionary *)dictionaryForLocation:(CLLocation *)location NS_SWIFT_NAME(dictionaryForLocation(_:));
 
+
++ (BOOL)isTestKey;
+
++ (void)sendLog:(RadarLogLevel)level message:(NSString *_Nonnull)message;
+
++ (void)flushLogs;
```
phil-brown, line 821: Do these need documentation?

RadarSDK/Radar.m
```diff
@@ -822,6 +823,29 @@ + (NSString *)stringForTripStatus:(RadarTripStatus)status {
     return str;
 }
 
+// move to a util?
```
phil-brown, line 826: Or make it a function on RadarLogLevel itself?

RadarSDK/Radar.m
```diff
@@ -849,4 +873,52 @@ - (void)dealloc {
     [[NSNotificationCenter defaultCenter] removeObserver:self];
 }
 
+
++ (BOOL)isTestKey {
```
phil-brown, line 877: can this be internal/not part of the public API?

RadarSDK/Radar.m
```diff
@@ -849,4 +873,52 @@ - (void)dealloc {
...
+    NSString *publishableKey = [RadarSettings publishableKey];
+    if ([publishableKey hasPrefix:@"prj_test_pk"] ||
+        [publishableKey hasPrefix:@"org_test_pk"]) {
+        return YES;
+    }
+    return NO;
+}
+
++ (void)sendLog:(RadarLogLevel)level message:(NSString *_Nonnull)message {
```
phil-brown, line 886: Same here - are these methods exposed to the public API?

RadarSDK/RadarLogBuffer.h
```diff
@@ -0,0 +1,31 @@
...
+
++ (instancetype)sharedInstance;
+
+- (void)write:(RadarLogLevel)level message:(NSString *)message;
+
+- (void)purgeOldestLogs;
+
+- (void)clearSyncedLogsFromBuffer:(NSUInteger)numLogs;
+
+- (NSArray<RadarLog *> *)getFlushableLogs;
```
phil-brown, line 27: Not sure if we are trying to be consistent with Android - but based on feedback from @jkao on my original PR, this was changed to `getFlushableLogsStash` on my implementation.

RadarSDK/Radar.m
```diff
@@ -849,4 +873,52 @@ - (void)dealloc {
...
+    
+    NSUInteger pendingLogCount = [flushableLogs count];
+    if (pendingLogCount == 0) {
+        return;
+    }
+    
+    RadarSyncLogsAPICompletionHandler onSuccess = ^(RadarStatus status){
+        if (status == RadarStatusSuccess) {
+            // clear the logs up to pendinglog count
+            [[RadarLogBuffer sharedInstance] clearSyncedLogsFromBuffer:pendingLogCount];
```
phil-brown, line 910: Not sure if this will work entirely as expected. Since this is an async operation, the app might log (and purge) more logs before the result comes back - so this could potentially wipe out unexpected logs. Can you instead follow the pattern of removing the logs from the buffer and on error for this callback, add them back to the buffer?

phil-brown (dismissed):
Looks good! Some concerns over the behavior of the use of a size count on the logs-to-remove. Can you also add some Unit Tests to verify the expected behavior, such as after successful and unsuccessful logging?

RadarSDK/RadarLog.m
```diff
@@ -0,0 +1,53 @@
...
+- (NSDictionary *)dictionaryValue {
+    NSMutableDictionary *dict = [NSMutableDictionary new];
+    dict[@"level"] = [Radar stringForLogLevel:self.level];
+    dict[@"message"] = self.message;
+
+    // convert (NSDate)createdAt to ISO8601 string
+    NSDateFormatter *dateFormatter = [[NSDateFormatter alloc] init];
+    dateFormatter.locale = [[NSLocale alloc] initWithLocaleIdentifier:@"en_US_POSIX"];
+    dateFormatter.timeZone = [NSTimeZone timeZoneWithAbbreviation:@"UTC"];
+    [dateFormatter setDateFormat:@"yyyy-MM-dd'T'HH:mm:ss.SSSZ"];
```
jrtibbetts, line 32: 1. `NSDateFormatter`s should be cached because they don't change, and they can be expensive to create.
2. There's an `NSISO8601DateFormatter` class that should do this out of the box, or with some options.

RadarSDK/RadarLogBuffer.h
jrtibbetts: On iOS, it should be a property (`flushableLogs`; no `get`), backed by a computed function.

RadarSDK/RadarLogBuffer.m
```diff
@@ -0,0 +1,74 @@
...
+
+- (void)write:(RadarLogLevel)level message:(NSString *)message {
+    // purge oldest if reached the max buffer size
+    NSUInteger logLength = [mutableLogBuffer count];
+    if (logLength >= MAX_BUFFER_SIZE) {
+        [self purgeOldestLogs];
+    }
+    // add new log to buffer
+    RadarLog *radarLog = [[RadarLog alloc] initWithMessage:message level:level];
+    [mutableLogBuffer addObject:radarLog];
```
jrtibbetts, line 46: Nit: This function takes a `RadarLogLevel` and an `NSString`, but the `RadarLog`'s initializer takes an `NSString` and a `RadarLogLevel`. I would swap the parameters of the `RadarLog`'s initializer.

RadarSDK/RadarLogBuffer.h
```diff
@@ -0,0 +1,31 @@
...
+
+#import "Radar.h"
+#import "RadarLog.h"
+
+NS_ASSUME_NONNULL_BEGIN
+
+@interface RadarLogBuffer : NSObject
+
+- (instancetype)init;
+
```
jrtibbetts, line 18: Add a comment that says the `sharedInstance` should be used instead, or move this declaration to the private interface.

RadarSDK/RadarLog.h
```diff
@@ -0,0 +1,37 @@
...
+ The datetime when the log occurred on the device.
+ */
+@property (nonnull, copy, nonatomic, readonly) NSDate *createdAt;
+
+- (instancetype _Nullable)initWithMessage:(NSString *_Nullable)message
+                                    level:(RadarLogLevel)level;
+
+- (NSDictionary *_Nonnull)dictionaryValue;
+
++ (NSArray<NSDictionary *> *_Nullable)arrayForLogs:(NSArray<RadarLog *> *_Nullable)logs;
```
jrtibbetts, line 35: Should this be on `RadarAPIClient` instead?

jrtibbetts (changes requested):

RadarSDK/RadarLog.h
corypisano: right now the pattern for these toArray type methods is to have it on the models
e.g. https://github.com/radarlabs/radar-sdk-ios/blob/master/RadarSDK/RadarPlace.m#L156

jrtibbetts: Ah, ok.

RadarSDK/RadarLogBuffer.h
corypisano: ok tried to update to property style but let me know if had something different in mind

jrtibbetts: Yep--that's great!

jrtibbetts (changes requested):
I would still cache the date formatter in a class variable.

corypisano:
> I would still cache the date formatter in a class variable.

ah yeah so the reason I didn't touch that yet is the dateformatter is used in a number of places
(https://github.com/radarlabs/radar-sdk-ios/blob/master/RadarSDK/RadarEvent.m#L107-L129, https://github.com/radarlabs/radar-sdk-ios/blob/master/RadarSDK/RadarUserInsightsLocation.m#L113) 
and probably best for someone smarter than me to setup a better pattern as a cleanup PR rather than making broader changes in this PR

jrtibbetts:
LOL, fair enough.

jrtibbetts (approved):
We can punt on caching the date formatters.

RadarSDK/Radar.h
```diff
@@ -794,6 +794,7 @@ typedef void (^_Nonnull RadarRouteMatrixCompletionHandler)(RadarStatus status, R
  */
 + (NSString *)stringForTripStatus:(RadarTripStatus)status NS_SWIFT_NAME(stringForTripStatus(_:));
 
+
```
jsani-radar, line 797: nit: extra `\n`

RadarSDK/Radar.m
jsani-radar: yeah thinking somewhere like `RadarUtils` might be a better home for this fn, especially if we want to use it in other places - had a similar comment about `asBool` that @jrtibbetts added to the PR that exposed `user.mocked`

RadarSDK/Radar.m
```diff
@@ -849,4 +850,55 @@ - (void)dealloc {
     [[NSNotificationCenter defaultCenter] removeObserver:self];
 }
 
+
++ (BOOL)isTestKey {
+    NSString *publishableKey = [RadarSettings publishableKey];
+    if ([publishableKey hasPrefix:@"prj_test_pk"] ||
+        [publishableKey hasPrefix:@"org_test_pk"]) {
+        return YES;
```
jsani-radar, line 858: nit: could return the result of the boolean operation here rather than having the extra nesting with `if`

RadarSDK/Radar.m
```diff
@@ -849,4 +850,55 @@ - (void)dealloc {
...
+        if (status != RadarStatusSuccess) {
+            [[RadarLogBuffer sharedInstance] addLogsToBuffer:flushableLogs];
+        }
+    };
+
+    [[RadarAPIClient sharedInstance] syncLogs:flushableLogs
+                            completionHandler:^(RadarStatus status) {
+                                  if (onComplete) {
+                                      [RadarUtils runOnMainThread:^{
+                                          onComplete(status);
```
jsani-radar, line 898: a little meta, but do we want to add a log here if this operation fails, or maybe at the call site of this fn?

RadarSDK/RadarAPIClient.h
```diff
@@ -109,6 +112,9 @@ typedef void (^_Nonnull RadarMatrixAPICompletionHandler)(RadarStatus status, NSD
                        units:(RadarRouteUnits)units
            completionHandler:(RadarMatrixAPICompletionHandler _Nonnull)completionHandler;
 
+- (void)syncLogs:(NSArray<RadarLog *> *)logs completionHandler:(RadarSyncLogsAPICompletionHandler _Nonnull)completionHandler;
+
+
```
jsani-radar, line 117: nit: extra `\n`

RadarSDK/RadarLog.m
jsani-radar: +1 to the above - let's cache the `dateFormatter` and not recreate it every time this fn is called

RadarSDK/RadarLog.m
```diff
@@ -0,0 +1,75 @@
...
+
+    NSString *createdAtString = [dateFormatter stringFromDate:self.createdAt];
+    dict[@"createdAt"] = createdAtString;
+
+    return dict;
+}
+
++ (NSArray<NSDictionary *> *)arrayForLogs:(NSArray<RadarLog *> *)logs {
+    if (!logs) {
+        return nil;
```
jsani-radar, line 64: nit: should the return value here be `[]` in the empty case?

RadarSDK/RadarLogBuffer.m
```diff
@@ -0,0 +1,80 @@
+//
+//  RadarLogBuffer.m
+//  RadarSDK
+//
+//  Copyright © 2021 Radar Labs, Inc. All rights reserved.
+//
+
+#import "Radar.h"
```
jsani-radar, line 8: do we need to import the main header file here (and in `RadarLogBuffer.h`)? doesn't seem like we're calling any fns on the public interface

RadarSDK/RadarLogger.m
```diff
@@ -7,6 +7,8 @@
 
 #import "RadarLogger.h"
 
+#import "Radar.h"
```
jsani-radar, line 10: ditto here - this import might be redundant to `Radar+Internal.h`, which defines the `sendLog` method

jsani-radar (commented):
looking solid overall! just left a few small comments

RadarSDK/Radar.m
corypisano: i moved this to private (not included in Radar.h) to match android 
https://github.com/radarlabs/radar-sdk-android/blob/master/sdk/src/main/java/io/radar/sdk/Radar.kt#L2198

RadarSDK/RadarLog.m
corypisano: hm ill defer to others, was cargo culting the other `arrayForX`'s which return nil https://github.com/radarlabs/radar-sdk-ios/blob/master/RadarSDK/RadarGeofence.m#L182-L185

in this case it shouldn't be hit either way since higher up, the sync logs call is skipped if no logs to send

RadarSDK/RadarLogBuffer.m
corypisano: ah yes will remove

RadarSDK/RadarLog.m
jrtibbetts: Looks like the exact same format string is used twice in `RadarEvent.m` and once in `RadarUserInsightsLocation.m`.

jrtibbetts: @corypisano, I can make that change and check it in on this branch, if you want.

RadarSDK/RadarLog.m
jrtibbetts: I thought the same thing, but then I saw that it's the same pattern in the other models.

jrtibbetts (approved):

RadarSDK/RadarLog.m
corypisano: sure!

jrtibbetts: Done.

corypisano: awesome thanks 👍

RadarSDK/RadarUtils.m
```diff
@@ -13,6 +13,20 @@
 
 @implementation RadarUtils
 
+static NSDateFormatter *_isoDateFormatter;
+
++ (NSDateFormatter *)isoDateFormatter {
+    if (_isoDateFormatter == NULL) {
```
jsani-radar, line 19: should this be `nil`?

jrtibbetts: Whoops, probably.

jrtibbetts: Done.

RadarSDK/Radar.m
```diff
@@ -859,4 +860,54 @@ - (void)dealloc {
...
+
++ (void)sendLog:(RadarLogLevel)level message:(NSString *_Nonnull)message {
+    [[RadarLogBuffer sharedInstance] write:level message:message];
+}
+
+/**
+ * Sends Radar log events to the server
+ */
++ (void)flushLogs {
+    // user _id has to exist
```
jkao, line 880: hmm not sure what this comment is for

RadarSDK/Radar.m
```diff
@@ -859,4 +860,54 @@ - (void)dealloc {
...
+
+    if (![self isTestKey]) {
+        return;
+    }
+
+    NSArray<RadarLog *> *flushableLogs = [[RadarLogBuffer sharedInstance] flushableLogs];
```
jkao, line 886: I kind of wonder if it's the responsibility of the caller to care about specifying log counts to flush vs just initializing that with the `RadarLogBuffer` to begin with. 

I guess it's a shared instance, but it seems a bit finicky to have to do all these checks / multiple reads (thinking of concurrency as well) from the call-side. does it even matter if the `pendingLogCount` is `0`? couldn't the buffer automagically handle that as well?

jkao: even the API logic feels like it could live in the RadarLogBuffer, or be passed a callback function to process the logs.

RadarSDK/Radar.m
jrtibbetts: I assume that it means that the key has to exist, and not be a test key.

RadarSDK/Radar.m
jrtibbetts: I agree that it shouldn't be the caller's responsibility, but maybe that can be addressed in a future PR.

jrtibbetts (approved):

RadarSDK/Radar.m
jkao: ah, when I see `user` and `_id` I'm assuming it refers to the primary key of a user entity and I didn't see a check for that. I do see the `isTestKey` check here, though

jkao: that being said, I also wonder if the test key check should just be something we check on the server-side and not be the responsibility of the caller? this is so we can help customers debug their test builds optionally

RadarSDK/Radar.m
corypisano: yea i agree some of this logic can shift to the buffer's responsibility, but trying to stay close to android
(check for test key, skip if log count is 0, handle adding logs back to buffer on error)
https://github.com/radarlabs/radar-sdk-android/blob/master/sdk/src/main/java/io/radar/sdk/Radar.kt#L2182

i think for the log count though it should be here, since if no logs we want to skip the api call

RadarSDK/Radar.m
jkao: otherwise, the server can just drop the logs

corypisano: the test key check we wanted client side for this V1 just to be extra careful before rolling out to live implementations. was a call to minimize any risk

RadarSDK/Radar.m
jkao: yeah, I think an implementation of something like:

```
[[RadarLogBuffer sharedInstance] flushLogs:maxLogsCount callback:sendLogsToAPI]
```

where X is the API caller which gets passed a list of logs, might make this a little cleaner. anyways this current impl is fine

PR #138, opened 2021-12-22
8093: Add RadarUser.fraud property and expose the mocked property

jrtibbetts:
[8093: Add "mocked" to RadarUser model](https://app.shortcut.com/radarlabs/story/8092/add-mocked-to-radaruser-model)

* Added a `RadarUser.mocked` property.
* Added a `RadarUser.fraud` property, which is a `RadarFraud` object. This has `proxy` and `mocked` properties, and the `RadarUser` properties of the same name simply pass back the `.fraud`'s properties.

(I'd like to add unit tests for serializing the `RadarUser` to a JSON dictionary.)


RadarSDK/RadarUser.h
```diff
@@ -141,10 +142,20 @@ typedef NS_ENUM(NSInteger, RadarLocationSource);
 @property (assign, nonatomic, readonly) RadarLocationSource source;
 
 /**
- A boolean indicating whether the user's IP address is a known proxy. May be `false` if Fraud is not enabled.
+ `true` if the user's IP address is a known proxy. May be `false` if Fraud is not enabled.
```
jsani-radar, line 145: I'd rather keep this as "A boolean indicating" for consistency with other fields on this object (ex. `stopped`, `foreground`)

RadarSDK/RadarUser.h
```diff
@@ -141,10 +142,20 @@ typedef NS_ENUM(NSInteger, RadarLocationSource);
...
  */
 @property (assign, nonatomic, readonly) BOOL proxy;
 
+/**
+ `true` if` the user's IP address is being mocked, such as in a simulation. May be `false` if Fraud is not enabled.
+ */
+@property (assign, nonatomic, readonly) BOOL mocked;
+
+/**
+ If Fraud is enabled, this will be non-`nil`, and will indicate whether the user's IP address or location is being mocked or proxied.
```
jsani-radar, line 155: Let's follow what we have for the `insights` object here: "Learned fraud state for the user. May be `nil` if Fraud is not enabled" - we'll eventually want to represent things beyond just mocked / proxied

RadarSDK/RadarUser.m
```diff
@@ -369,12 +376,21 @@ - (NSDictionary *)dictionaryValue {
...
-    NSDictionary *fraudDict = @{@"proxy": @(self.proxy)};
-    [dict setValue:fraudDict forKey:@"fraud"];
+    [dict setValue:[self.fraud dictionaryValue] forKey:@"fraud"];
     if (self.trip) {
         [dict setValue:[self.trip dictionaryValue] forKey:@"trip"];
     }
     return dict;
 }
 
+- (bool)asBool:(NSObject *)object {
```
jsani-radar, line 386: should this belong in a more generic class, like `RadarUtils`? seems like an abstract operation that we might want to use beyond just `RadarUser`

jsani-radar (commented):
few small comments, overall looking good

defer to @tjulien on whether we want to bite off the redundancy between top-level fields and the `fraud` object in this iteration

RadarSDK/RadarUser.m
jrtibbetts: It definitely should be on an extension of `NSObject` or something, but I'm trying to take baby steps here. :)

(This is off-topic, but these model classes are prime candidates for converting to `Codable` objects in Swift--it would eliminate almost all of the existing parsing code. Even better would be using [Swagger](https://swagger.io) to define the APIs, and then letting _it_ generate the client- and server-side model code.)

RadarSDK/RadarUser.h
jrtibbetts: Done.

RadarSDK/RadarUser.h
jrtibbetts: Done.

RadarSDK/RadarFraud.h
```diff
@@ -0,0 +1,27 @@
...
+//  Copyright © 2021 Radar Labs, Inc. All rights reserved.
+
+#ifndef RadarFraud_h
+#define RadarFraud_h
+#import <Foundation/Foundation.h>
+
+@interface RadarFraud : NSObject
+
+/**
+ `true` if` the user's IP address is being mocked, such as in a simulation. May be `false` if Fraud is not enabled.
```
tjulien, line 14: `IP address` should be replaced with `location` here. (and also in `RadarUser`)

RadarSDK/RadarFraud.h
```diff
@@ -0,0 +1,27 @@
...
+ */
+@property (nonatomic, readonly) bool mocked;
+
+/**
+ `true` if the user's IP address is a known proxy. May be `false` if Fraud is not enabled.
```
tjulien, line 19: `IP address` here is correct (just clarifying, given my comment above).

tjulien:
Looks good - just had two small comments about docs, but other than that this looks ready to ship

RadarSDK/RadarUser.h
```diff
@@ -145,6 +146,16 @@ typedef NS_ENUM(NSInteger, RadarLocationSource);
  */
 @property (assign, nonatomic, readonly) BOOL proxy;
 
+/**
+ A boolean indicating whether the user's IP address is being mocked or simulated. May be `false` if Fraud is not enabled.
+ */
+@property (assign, nonatomic, readonly) BOOL mocked;
+
+/**
+ Learned fraud state for the user. May be nil if Fraud is not enabled
```
jsani-radar, line 155: nit: `nil` should have single quotes around it as it does for other properties (ex. `segments`, `topChains`)

RadarSDK/RadarFraud.h
jsani-radar: for consistency, these should match the description strings in `RadarUser`

RadarSDK/RadarUser.m
jrtibbetts: 📌

RadarSDK/RadarFraud.h
jrtibbetts: Done.

RadarSDK/RadarUser.h
jrtibbetts: Done.

RadarSDK/RadarUser.h
```diff
@@ -145,6 +146,16 @@ typedef NS_ENUM(NSInteger, RadarLocationSource);
  */
 @property (assign, nonatomic, readonly) BOOL proxy;
 
+/**
+ `true` if` the user's location is being mocked, such as in a simulation. May be `false` if Fraud is not enabled.
+ */
+@property (assign, nonatomic, readonly) BOOL mocked;
+
+/**
+ Learned fraud state for the user. May be `nil` if Fraud is not enabled
```
jsani-radar, line 155: nit: period here

RadarSDK/RadarUser.h
```diff
@@ -145,6 +146,16 @@ typedef NS_ENUM(NSInteger, RadarLocationSource);
  */
 @property (assign, nonatomic, readonly) BOOL proxy;
 
+/**
+ `true` if` the user's location is being mocked, such as in a simulation. May be `false` if Fraud is not enabled.
```
jsani-radar, line 150: nit: dangling single-quote after `if`

RadarSDK/RadarUser.h
jrtibbetts: Fixed.

RadarSDK/RadarUser.h
jrtibbetts: Fixed. (Well, replaced entirely.)

RadarSDK/RadarFraud.h
```diff
@@ -0,0 +1,27 @@
...
+//  Copyright © 2021 Radar Labs, Inc. All rights reserved.
+
+#ifndef RadarFraud_h
+#define RadarFraud_h
+#import <Foundation/Foundation.h>
+
+@interface RadarFraud : NSObject
+
+/**
+ A boolean indicating whether the user's location is being mocked, such as in a simulation. May be `false` is Fraud is not enabled.
```
jsani-radar, line 14: nit: "if Fraud is"

RadarSDK/RadarUser.h
```diff
@@ -145,6 +146,16 @@ typedef NS_ENUM(NSInteger, RadarLocationSource);
  */
 @property (assign, nonatomic, readonly) BOOL proxy;
 
+/**
+ A boolean indicating whether the user's location is being mocked, such as in a simulation. May be `false` is Fraud is not enabled.
```
jsani-radar, line 150: nit: "if Fraud is"

RadarSDK/RadarFraud.h
jrtibbetts: Done.

RadarSDK/RadarUser.h
jrtibbetts: Done.

tjulien (approved):

tjulien:
https://app.shortcut.com/radarlabs/story/8092/add-mocked-to-radaruser-model

phil-brown (approved):

RadarSDK/RadarUser.h
```diff
@@ -145,6 +146,16 @@ typedef NS_ENUM(NSInteger, RadarLocationSource);
...
 
+/**
+ A boolean indicating whether the user's location is being mocked, such as in a simulation. May be `false` if Fraud is not enabled.
+ */
+@property (assign, nonatomic, readonly) BOOL mocked;
+
+/**
+ Learned fraud state for the user. May be `nil` if Fraud is not enabled.
+ */
+@property (nonnull, copy, nonatomic, readonly) RadarFraud *fraud;
```
jsani-radar, line 157: `nonnull` designation here is at odds with the comment about coming back `nil` - seems like this should be nullable

RadarSDK/RadarUser.m
nickpatrick: +1, would be nice to centralize this when we refactor serialization/deserialization logic later.

Does it matter that we're using `bool` instead of `BOOL` here? We use `BOOL` elsewhere, although not so sure it matters in Swift.

nickpatrick: I'm gonna swap it to `BOOL` while releasing to be consistent, if I'm doing something dumb we can undo it later

jrtibbetts: That's fine; I was just slipping into a Swift accent.

RadarSDK/RadarUser.h
jrtibbetts: The following commit removes the second half of that sentence, since the way the code is written, the `fraud` object will never be `null`.

PR #140, opened 2022-01-15
8335: Add guards against stopping CLVisits and SLCs from initializing Radar

jrtibbetts:
[8335: Add guards against stopping CLVisits and SLCs from initializing Radar (SDK P1)](https://app.shortcut.com/radarlabs/story/8335)

`Radar.initializeWithPublishableKey(:)` now calls a new `RadarLocationManager.updateLocationFromInitialize(:)` function. This is the same as `.updateLocation()`, but if its `fromInitialize` argument is YES, then it will _not_ update the location manager if tracking is off.


RadarSDK/RadarLocationManager.m
```diff
@@ -240,10 +240,19 @@ - (void)requestLocation {
 }
 
 - (void)updateTracking {
-    [self updateTracking:nil];
+    [self updateTracking:nil fromInitialize:NO];
+}
+
+- (void)updateTrackingFromInitialize:(BOOL)fromInitialize {
```
phil-brown, line 246: Wondering if this is an unusual name, given you pass the value `fromInitialize`? Wouldn't it be more standard to use `with`? Or just not pass any parameter here (pass-through would always be `YES`)?

RadarSDK/RadarLocationManager.m
```diff
@@ -240,10 +240,19 @@ - (void)requestLocation {
...
+    [self updateTracking:nil fromInitialize:fromInitialize];
 }
 
 - (void)updateTracking:(CLLocation *)location {
+    [self updateTracking:location fromInitialize:NO];
+}
+
+- (void)updateTracking:(CLLocation *)location
+        fromInitialize:(BOOL)fromInitialize {
```
phil-brown, line 255: again here - what's best practice? Could it be `withIsFromInitiailize`?

RadarSDK/RadarLocationManager.m
jrtibbetts: I agree that it's unusual, but it's what was recommended in the Shortcut. I don't think that `with` would make more sense, because it's supposed to indicate whether this function is being called _from_ the initializer. Perhaps `fromInitializer` would make more sense?

RadarSDK/RadarLocationManager.m
jsani-radar: the second parameter and beyond in objective-c tends to be redundant / wordy due to the quirky way it organizes them - https://stackoverflow.com/questions/3911593/how-to-call-a-method-from-another-method-in-objective-c

RadarSDK/RadarLocationManager.m
jsani-radar: my apologies for not setting expectations here - the shortcut ticket isn't meant to be 100% prescriptive in terms of exact names, style, and specific code changes, this one was particularly long since it had been talked about at length, and I wanted to capture all the details

I was thinking we could keep the fn name as `updateTracking` but just have the internal param name be `fromInitializer`, sort of how the `updateTracking` fn that took the location param was before this change

```
- (void) updateTracking:(BOOL)fromInitializer
```

how does that sound? open to feedback

jrtibbetts: That's not possible in Objective-C, because I already declare a function with one parameter (`- (void)updateTracking:(CLLocation *)location`). Objective-C supports function overloading _only_ if the functions have different numbers of arguments, because it ignores parameter names when checking for overloads. The workaround is to change the _name_ of the function (which means you're no longer overloading it), which leads to the redundancy that @phil-brown points out on line 245.

Screen Shot 2022-01-18 at 8 48 58 AM

Both of these functions have a signature of just `updateTracking(:)`, so that's not allowed. (It *is* allowed in Swift, which takes parameter names into account.)

jsani-radar: hmm, that function with one parameter is now going to take 2 after this PR, correct? At least from what I'm seeing on L254-L255
```
- (void)updateTracking:(CLLocation *)location
                 fromInitialize:(BOOL)fromInitialize
```

so the function signature I wrote out in my previous comment would be the only one with 1 param now

jrtibbetts: Not quite. I left the original `(void)updateTracking:(CLLocation *)location` as-is so that all callers _other_ than the initializer could keep calling it, instead of changing all of them.

jsani-radar: ah I see, that's probably a cleaner approach - I'll defer to you then on what the new method / param name should be, I'm fine with `fromInitialize` / `fromInitializer` / or even something better you propose

RadarSDK/RadarLocationManager.m
```diff
@@ -337,8 +346,11 @@ - (void)updateTracking:(CLLocation *)location {
         } else {
             [self stopUpdates];
             [self removeAllRegions];
-            [self.locationManager stopMonitoringVisits];
-            [self.locationManager stopMonitoringSignificantLocationChanges];
+            
+            if (!fromInitialize) {
```
phil-brown, line 350: would it make sense to flip the function argument? For example, the signature could become

```objective-c
- (void)updateTracking:(CLLocation *)location
        withStandardWorkflow:(BOOL)isStandardWorkflow
```

Then the default implementation would pass `YES` and the initialize workflow would pass `NO`, and this if check could be flipped. It's a really small optimization, but could also help with the function naming.

tjulien:
https://app.shortcut.com/radarlabs/story/8335/add-guards-against-stopping-clvisits-and-slcs-from-initializing-radar-sdk-p1

RadarSDK/Radar.m
```diff
@@ -48,7 +48,7 @@ + (void)initializeWithPublishableKey:(NSString *)publishableKey {
     }
 
     [RadarSettings setPublishableKey:publishableKey];
-    [[RadarLocationManager sharedInstance] updateTracking];
+    [[RadarLocationManager sharedInstance] updateTrackingFromInitialize:NO];
```
tjulien, line 51: Should be passing `YES` here, right?

RadarSDK/RadarLocationManager.m
tjulien, line 350: Let's add a comment here explaining why we're being careful about shutting down monitoring on the `CLLocationManager`.

RadarSDK/Radar.m
jrtibbetts: Oof.

tjulien:
@jsani-radar @jrtibbetts The original comment was better, because it explains the "why" behind the code.  The new comment doesn't tell you anything beyond what the code does.

jsani-radar:
whoops! don't mind me then 🙈 feel free to revert the updated comment

tjulien (approved):

PR #144, opened 2022-01-23
Don't log logging calls

tjulien:
This should fix the crashes we've been seeing on internal test builds.  A similar fix was made in https://github.com/radarlabs/radar-sdk-android/pull/156

### DIFF
--- a/RadarSDK/RadarAPIClient.m
+++ b/RadarSDK/RadarAPIClient.m
@@ -88,6 +88,7 @@
                               headers:headers
                                params:nil
                                 sleep:NO
+                           logPayload:YES
                     completionHandler:^(RadarStatus status, NSDictionary *_Nullable res) {
                         if (!res) {
                             return;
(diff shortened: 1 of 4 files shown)


RadarSDK/RadarAPIClient.m
```diff
@@ -911,11 +926,14 @@ - (void)syncLogs:(NSArray<RadarLog *> *)logs completionHandler:(RadarSyncLogsAPI
     NSArray *logsArray = [RadarLog arrayForLogs:logs];
     [params setValue:logsArray forKey:@"logs"];
 
+    // "logPayload = false" enforces "don't log the logging call".
+    // Otherwise, all log entries would continue coalesce and would never fully clear from the buffer
```
jsani-radar, line 930: super nits: "continue to coalesce"

also, we should probably enforce period or no period for both lines

tjulien: fixed

jsani-radar (approved):
🚀

jrtibbetts (approved):

PR #146, opened 2022-01-24
(No ticket) Adding the Gemfiles to source control

jrtibbetts:
Using Ruby's bundler (e.g. `bundle exec make dist`) fixes my Ruby environment issues, so keeping the Gemfiles in source control will ensure that the correct gem versions will be used.


Gemfile.lock
```diff
@@ -0,0 +1,120 @@
+GEM
```
jsani-radar, line 1: does this pretty much work the same as a `package-lock.json` in npm? if it's present the install will download the versions specified here, but if not it might grab subsequent versions that would mess up backcompat?

jsani-radar (commented):
nice! overall lgtm - just had a small q around the exact behavior here

Gemfile.lock
jrtibbetts: Yep, just like `package-lock.json`. @nickpatrick pointed out that using these with `bundle install` is kind of the same thing that `make bootstrap` does--it installs the Ruby gems needed to do command-line builds. But for some reason, `make bootstrap` alone didn't fix my issues. 

(I should also point out that `make bootstrap` _also_ installs the Homebrew tools (`xctool` and `clang-format`) that are needed. Those aren't Ruby gems, so they're not affected by these `Gemfile`s.)

jsani-radar (approved):

PR #152, opened 2022-02-02
8281: Distribute SDK as source

jrtibbetts:
[[sc-8281]: [iOS] Configure CocoaPods, Carthage, and Swift Package Manager to distribute the SDK as source, not as an XCFramework](https://app.shortcut.com/radarlabs/story/8281/ios-configure-cocoapods-carthage-and-swift-package-manager-to-distribute-the-sdk-as-source-not-as-an-xcframework)

This experiment successfully tests whether we can distribute and build the SDK from source, as opposed to precompiling it into an `.xcframework`. Carthage and CocoaPods support this with minimal changes; however, **Swift Package Manager required moving several files around to separate public header files from internal ones**. 

# Changes

* **Modifies** the `RadarSDK.podspec` and `Package.swift` files to point to source folders, not a precompiled `.xcframework`.
* **Updates** the version number to `3.3.1`, since there aren't any breaking changes.
* **Removes** the `dist/` subfolder.
* **Moves** public header files into an `Include/` folder.
* ~~**Moves** non-public header files _and internal-only class files_ into an `Internal/` folder. (SPM requires the `Include/` folder to contain only header files, but it doesn't have the same restrictions on other folders.)~~

# Notes
* It's somewhat suboptimal to have public header files in a separate folder from their corresponding `.m` files, but the good news is that Xcode's `ctrl+cmd+⬆️` keyboard combination to "bounce" between interface and implementation files still works.
* The `#include` paths for internal headers use relative paths, and it's a lot uglier than before, but this is needed _only_ for the SDK's own Objective-C unit tests; it won't be needed for Swift tests (if and when), or for any apps or frameworks that depend on the SDK.
* I've verified that the source-only distribution works in fresh CocoaPods-, Carthage-, and Swift Package Manager-based app projects written in Swift.


Package.swift
```diff
@@ -14,9 +14,14 @@ let package = Package(
...
-        .binaryTarget(
+        .target(
             name: "RadarSDK",
-            path: "dist/RadarSDK.xcframework"
+            path: "RadarSDK",
+            exclude: ["Info.plist", "RadarSDK.h"],
+            publicHeadersPath: "Include",
+            cSettings: [
+                .headerSearchPath("Internal")
+            ]
```
phil-brown, line 24: Can you add comments in here about why these things are included or excluded?

RadarSDK.podspec
```diff
@@ -1,12 +1,12 @@
 Pod::Spec.new do |s|
   s.name                  = 'RadarSDK'
-  s.version               = '3.3.0'
+  s.version               = '3.3.1-beta.1'
```
phil-brown, line 3: Should this be the same number as it is in Example/Example.xcodeproj/project.pbxproj?

RadarSDK/Internal/RadarUtils.m
```diff
@@ -45,7 +45,7 @@ + (NSNumber *)timeZoneOffset {
 }
 
 + (NSString *)sdkVersion {
-    return @"3.3.0";
+    return @"3.3.1-beta.1";
```
phil-brown, line 48: Example/Example.xcodeproj/project.pbxproj has a different number - is that ok?

RadarSDKTests/RadarAPIHelperMock.h
```diff
@@ -6,9 +6,8 @@
 //
 
 #import <Foundation/Foundation.h>
-#import <RadarSDK/RadarSDK.h>
-
-#import "RadarAPIHelper.h"
+#import <RadarSDK.h>
+#import "../RadarSDK/Internal/RadarAPIHelper.h"
```
phil-brown, line 10: Will these changes affect how customers add imports in their code as well?

phil-brown (commented):
All these moves to the Include directory seems unusual - is that a common practice? Was thinking it would make more sense to move the build or package-level stuff up a directory, in order to maintain the existing directory structure. Thoughts?

RadarSDK.podspec
jrtibbetts: Good question, and the answer is no. :) TestFlight & the App Store only support version numbers that consist of `.`-separated integers, and they get the version number from the `.pbxproj`'s `MARKETING_VERSION` value. But CocoaPods (and everything else) _does_ support version numbers like this. The compromise is to use this style in the `.podspec` and in `RadarUtils.m`, and the other style in the `.pbxproj`.

RadarSDK/Internal/RadarUtils.m
jrtibbetts: See above. As long as it matches the `.podspec`, we're good.

RadarSDKTests/RadarAPIHelperMock.h
jrtibbetts: No, because only the internal (non-public) headers need to work this way. There's a lot of confusion out there about what should be done for Swift Package Manager support, and this seems to be one approach.

jrtibbetts:
> All these moves to the Include directory seems unusual - is that a common practice? Was thinking it would make more sense to move the build or package-level stuff up a directory, in order to maintain the existing directory structure. Thoughts?

You're right--they do seem unusual. But Swift Package Manager has certain conventions, and one of them is to put public headers in an `Include` folder, and non-public headers in another folder. What _didn't_ work is the existing project structure, where the public headers were in the `RadarSDK` and `RadarSDK/Models`, because that `publicHeadersPath` property supports only a single path. (I wonder whether it can support multiple paths using some sort of delimiter. 🤔 )

Let's just say that SPM's support for Objective-C is half-baked, at best. (Or maybe it's great, but it's poorly documented.)

tjulien:
I see all of the `Include/` file moves, but I don't see any `Internal/` moves.  Are those missing?

jrtibbetts:
In the most recent commit (right above your comment), I reverted all of the `Internal/` moves because I found a way to avoid them. That reduced the number of files changed in this PR from 185 to 112, and the majority of the remaining file changes are from removing the `dist/` folder.

I updated the PR description to remove the `Internal/` step.

tjulien:
Does Toolkit build and run using the new artifacts from this build?

Package.swift
jrtibbetts: > The choice of Internal was mine, though.

Note that I reverted the `Internal/` move in a later commit, because I found a workaround.

jrtibbetts:
The beauty of it is that there _are_ no artifacts anymore--it's just bare source, and yes, I've confirmed that this source-only approach works with projects that use CocoaPods, Carthage, and Swift Package Manager.

Package.swift
```diff
@@ -14,9 +14,14 @@ let package = Package(
     ],
     dependencies: [],
     targets: [
-        .binaryTarget(
+        .target(
             name: "RadarSDK",
-            path: "dist/RadarSDK.xcframework"
+            path: "RadarSDK",
+            exclude: ["Info.plist", "RadarSDK.h"],
```
tjulien, line 20: Do we need to exclude these 2 things?  I don't see other peer (segment, one signal, amplitude) `Package.swift`'s making use of `exclude`.  Especially curious about `RadarSDK.h`, it seems strange that would need to be excluded.

jrtibbetts: If I don't exclude `Info.plist` (which is used by `RadarSDK.xcodeproj`), then SPM 5.3+ (see the top line of `Package.swift`) complains about it being an unhandled file. Segment is using SPM 5.2, which doesn't complain, so they don't have to exclude the file. The `RadarSDK.h` exclusion may be a relic of when I was trying to get unit tests working in SPM; I'll check to see whether I still need it.

If I edit Segment's `Package.swift` to use SPM 5.3, then it complains, too. :)
Screen Shot 2022-02-08 at 12 17 34 AM

RadarSDK/RadarUtils.m
```diff
@@ -45,7 +45,7 @@ + (NSNumber *)timeZoneOffset {
 }
 
 + (NSString *)sdkVersion {
-    return @"3.3.0";
+    return @"3.3.1-beta.1";
```
jrtibbetts, line 48: Remember, folks, the `.xcodeproj` file just says `3.3.1`, because TestFlight (which uses the project file's version number) doesn't support version strings like `3.3.1-beta.1`, so I have to use the bare version number, and indicate the beta by setting the `BUILD_NUMBER`.

RadarSDKTests/RadarAPIHelperMock.h
```diff
@@ -6,9 +6,8 @@
 //
 
 #import <Foundation/Foundation.h>
-#import <RadarSDK/RadarSDK.h>
-
-#import "RadarAPIHelper.h"
+#import <RadarSDK.h>
+#import "../RadarSDK/RadarAPIHelper.h"
```
jrtibbetts, line 10: Yeah, I don't like this change, either, but 1) it's limited to our internal unit tests, and 2) future unit tests will hopefully be written in Swift. :)

Package.swift
tjulien: 👍

tjulien (approved):

jrtibbetts:
I confirmed that the Swift package works without the `"RadarSDK.h"` in the list of files to exclude.

PR #159, opened 2022-02-16
Log errors

jsani-radar:


RadarSDK/RadarAPIHelper.m
```diff
@@ -76,6 +76,9 @@ - (void)requestWithMethod:(NSString *)method
             void (^dataTaskCompletionHandler)(NSData *data, NSURLResponse *response, NSError *error) = ^(NSData *data, NSURLResponse *response, NSError *error) {
                 if (error) {
                     dispatch_async(dispatch_get_main_queue(), ^{
+                        [[RadarLogger sharedInstance]
+                            logWithLevel:RadarLogLevelError
+                                message:[NSString stringWithFormat:@"Received network error | error = %@", error]];
```
jrtibbetts, line 81: I wonder whether I should audit what thread(s) `logWithLevel()` is getting called on. 🤔

jrtibbetts (approved):

jsani-radar:
hey Jason, totally fine with you merging this in, but I was hoping this could serve as a jumping off point for an audit of where we might be missing error logs, rather than a one-off change - know it's still in `develop` so we can always add more

two examples I found:
1. we're not logging the [publishable key error](https://github.com/radarlabs/radar-sdk-ios/blob/95070d16fd13559761853dbc7118f9695d0231f6/RadarSDK/RadarAPIClient.m#L120) - it'd be notable to see if the device has an incorrect key or the dev had to make multiple attempts to configure the SDK properly
2. our [location timeout log](https://github.com/radarlabs/radar-sdk-ios/blob/aace15e1bc33677ed02a20206aee30ad1325b7bc/RadarSDK/RadarLocationManager.m#L121) currently has a level of `debug`, but should probably be `error` given the status we're returning

PR #166, opened 2022-04-01
[9522] Add Github Action that notifies cross-platform sdk repos of new release

andlum:
## Summary
Using [repository-dispatch](https://github.com/peter-evans/repository-dispatch) github action to notify cross-platform SDK repos of new release so we can start automating version bumps


.github/workflows/notify.yml
```diff
@@ -0,0 +1,22 @@
+# When a new release is created on GitHub, this workflow sends an event
+# to cross-platform SDK repos that depend on native iOS SDK
+name: Notify
```
tjulien, line 3: Maybe we `name` this `Downstream version bump`?  And name the file `downstream-version-bump.yml`?

.github/workflows/notify.yml
```diff
@@ -0,0 +1,22 @@
...
+  workflow_dispatch:
+
+jobs:
+  notify:
+    runs-on: macos-latest
+    steps:
+      - name:
+        uses: peter-evans/repository-dispatch@v2
+        with:
+          token: ${{ secrets.REPO_ACCESS_TOKEN }}
```
tjulien, line 19: We can use `${{ secrets.RADAR_CIRCLECI_PAT }}`

.github/workflows/downstream-version-bump.yml
```diff
@@ -0,0 +1,22 @@
+# When a new release is created on GitHub, this workflow sends an event
+# to cross-platform SDK repos that depend on native iOS SDK
+name: Downstream version bump
```
tjulien, line 3: Let's make this consistent with Android, but other than that, this looks good.

tjulien (approved):

PR #182, opened 2022-06-02
Support UNLocationNotificationTrigger for synced geofences

nickpatrick:
If `radar:notificationText` is added to `geofence.metadata`, the SDK will add local notification with a `UNLocationNotificationTrigger` (https://developer.apple.com/documentation/usernotifications/unlocationnotificationtrigger) for that geofence when it is synced to the SDK for `trackingOptions.syncGeofences`. This will display a notification when a user enters the client-side geofence _even if the end user has only granted foreground_ permissions. It will _not_ trigger a track request that generates an event or creates a user.


RadarSDK/RadarLocationManager.m
```diff
@@ -455,16 +456,53 @@ - (void)replaceSyncedGeofences:(NSArray<RadarGeofence *> *)geofences {
...
+                    UNLocationNotificationTrigger *trigger = [UNLocationNotificationTrigger triggerWithRegion:region repeats:YES];
+
+                    UNNotificationRequest *request = [UNNotificationRequest requestWithIdentifier:identifier content:content trigger:trigger];
+
+                    UNUserNotificationCenter *center = [UNUserNotificationCenter currentNotificationCenter];
+                    [center addNotificationRequest:request
+                             withCompletionHandler:^(NSError *_Nullable error) {
+                                 if (error) {
+                                     [[RadarLogger sharedInstance]
+                                         logWithLevel:RadarLogLevelDebug
```
jrtibbetts, line 481: Should we use `RadarLogLevelError`?

PR #190, opened 2022-06-24
Expose custom events API

jsani-radar:
Support custom events

Analogue of https://github.com/radarlabs/radar-sdk-android/pull/227


RadarSDK/RadarAPIClient.m
```diff
@@ -932,6 +932,50 @@ - (void)getMatrixFromOrigins:(NSArray<CLLocation *> *)origins
...
+        return completionHandler(RadarStatusErrorPublishableKey, nil, nil);
+    }
+
+    NSMutableDictionary *params = [NSMutableDictionary new];
+    params[@"id"] = [RadarSettings _id];
+    params[@"installId"] = [RadarSettings installId];
+    params[@"userId"] = [RadarSettings userId];
+    params[@"deviceId"] = [RadarUtils deviceId];
+    params[@"name"] = name;
+    params[@"metadata"] = metadata;
```
jsani-radar, line 950: there are def params that I'm missing here, things like `foreground` / `stopped` / `replayed`

RadarSDK/RadarEvent.m
```diff
@@ -38,6 +38,22 @@ @implementation RadarEvent
...
+
+    NSDictionary *eventObj = (NSDictionary *)object;
+
+    RadarEvent *event = [[RadarEvent alloc] initWithObject:eventObj];
+
+    if (!event) {
+        return nil;
+    }
+
+    return event;
```
jrtibbetts, line 54: `return event` is sufficient. Better yet, just `return [[RadarEvent alloc] initWithObject:eventObj];`.

RadarSDK/RadarAPIClient.m
jsani-radar, line 950: there are def params that I'm missing here, things like foreground / stopped / replayed /cc @david-goodfellow

RadarSDK/RadarAPIClient.m
```diff
@@ -932,6 +932,50 @@ - (void)getMatrixFromOrigins:(NSArray<CLLocation *> *)origins
...
+            return completionHandler(status, nil, nil);
+        }
+
+        id eventObj = res[@"event"];
+        RadarEvent *event = [RadarEvent eventFromObject:eventObj];
+        if (event) {
+            return completionHandler(RadarStatusSuccess, res, event);
+        }
+
+        completionHandler(RadarStatusErrorServer, nil, nil);
```
jrtibbetts, line 975: Stick this in an `else`, so that it's clear.

RadarSDK/Radar.m
```diff
@@ -711,6 +711,53 @@ + (void)getMatrixFromOrigins:(NSArray<CLLocation *> *_Nonnull)origins
                                         }];
 }
 
++ (void)sendEvent:(NSString *)name
+     withMetadata:(NSDictionary *_Nullable)metadata
+completionHandler:(RadarSendEventCompletionHandler)completionHandler {
+
+    // TODO(jsani): check if permissions are enabled for events that can occur before location permission prompts
+    // if not, send the request with location = NULL
```
jrtibbetts, line 719: Stick this in a Shortcut ticket so that we don't forget about it.

jsani-radar, line 719: there is a valid use case here of sending events without a location, but that pattern doesn't exist anywhere else in this file - would it make sense to have a local instance of `RadarPermissionsHelper` here? `RadarLocationManager` doesn't expose `locationAuthorizationStatus` like the permissions helper does https://github.com/radarlabs/radar-sdk-ios/blob/develop/RadarSDK/RadarPermissionsHelper.h#L15

/cc @jrtibbetts @nickpatrick

RadarSDK/Include/Radar.h
```diff
@@ -765,6 +774,33 @@ typedef void (^_Nonnull RadarRouteMatrixCompletionHandler)(RadarStatus status, R
...
+ */
++ (void)sendEvent:(NSString *)name
+     withMetadata:(NSDictionary *_Nullable)metadata
+completionHandler:(RadarSendEventCompletionHandler)completionHandler NS_SWIFT_NAME(sendEvent(name:metadata:completionHandler:));
+
+/**
+ Sends a custom event with a manually provided location.
+
+ @param name The name of the event.
+ @param location The location of the event.
```
jrtibbetts, line 794: Add `@param metadata The metadata associated with the event.`

RadarSDK/Include/Radar.h
```diff
@@ -219,6 +219,15 @@ typedef void (^_Nonnull RadarRouteCompletionHandler)(RadarStatus status, RadarRo
  */
 typedef void (^_Nonnull RadarRouteMatrixCompletionHandler)(RadarStatus status, RadarRouteMatrix *_Nullable matrix);
 
+/**
+ Called when a request to send a custom event succeeds, fails, or times out.
+
+ Receives the request status and, if successful, the events.
```
jrtibbetts, line 225: `event` (singular)

RadarSDK/Include/Radar.h
```diff
@@ -765,6 +774,33 @@ typedef void (^_Nonnull RadarRouteMatrixCompletionHandler)(RadarStatus status, R
...
+completionHandler:(RadarSendEventCompletionHandler)completionHandler NS_SWIFT_NAME(sendEvent(name:metadata:completionHandler:));
+
+/**
+ Sends a custom event with a manually provided location.
+
+ @param name The name of the event.
+ @param location The location of the event.
+ @param completionHandler A completion handler.
+
+ @see https://radar.com/documentation/api#send-event
```
jsani-radar, line 797: on me to write up the API spec in our docs

jrtibbetts (approved):

RadarSDK/RadarAPIClient.m
jsani-radar: tbh I was just cribbing off the pattern for `trackWithLocation` - https://github.com/radarlabs/radar-sdk-ios/blob/develop/RadarSDK/RadarAPIClient.m#L283-L288

seems like we do it across all the fns in this file - https://github.com/radarlabs/radar-sdk-ios/blob/develop/RadarSDK/RadarAPIClient.m#L628-L631

I guess that might be why we have `return` in the first statement but not second

RadarSDK/RadarEvent.m
jsani-radar: good looks, much cleaner

RadarSDK/Include/Radar.h
jsani-radar: good catch

RadarSDK/Radar.m
jsani-radar: I actually want to include this functionality in the PR, asked you and Nick about the best way to achieve it - https://github.com/radarlabs/radar-sdk-ios/pull/190#discussion_r906421525

we'll need some way to expose the `locationAuthorizationStatus` to this class

RadarSDK/Include/Radar.h
```diff
@@ -765,6 +774,34 @@ typedef void (^_Nonnull RadarRouteMatrixCompletionHandler)(RadarStatus status, R
...
+ Sends a custom event.
+
+ @param name The name of the event.
+ @param metadata The metadata associated with the event.
+ @param completionHandler A completion handler.
+
+ @see https://radar.com/documentation/api#send-event
+ */
++ (void)sendEvent:(NSString *)name
+     withMetadata:(NSDictionary *_Nullable)metadata
```
david-goodfellow, line 787: There are future iterations we want to improve on but this requires schema changes / backfills for our events table in athena. Specifically the events table does not have a column for metadata or a user's external ID
```
CREATE EXTERNAL TABLE `events_by_project_v2`(
  `_id` string COMMENT '', 
  `createdat` timestamp COMMENT '', 
  `actualcreatedat` timestamp COMMENT '', 
  `organization` string COMMENT '', 
  `live` boolean COMMENT '', 
  `locationaccuracy` double COMMENT '', 
  `confidence` int COMMENT '', 
  `place` string COMMENT '', 
  `placecategories` string COMMENT '', 
  `placechainname` string COMMENT '', 
  `placechainslug` string COMMENT '', 
  `placegroup` string COMMENT '', 
  `placename` string COMMENT '', 
  `alternateplaces` array<string> COMMENT '', 
  `placecandidates` array<string> COMMENT '', 
  `duration` double COMMENT '', 
  `longitude` double COMMENT '', 
  `latitude` double COMMENT '', 
  `user` string COMMENT '', 
  `geofence` string COMMENT '', 
  `geofencedescription` string COMMENT '', 
  `geofencetag` string COMMENT '', 
  `geofenceexternalid` string COMMENT '', 
  `region` string COMMENT '', 
  `countryregionname` string COMMENT '', 
  `countryregioncode` string COMMENT '', 
  `countryregion` string COMMENT '', 
  `dmaregion` string COMMENT '', 
  `dmaregionname` string COMMENT '', 
  `dmaregioncode` string COMMENT '', 
  `stateregion` string COMMENT '', 
  `stateregionname` string COMMENT '', 
  `stateregioncode` string COMMENT '', 
  `postalcoderegion` string COMMENT '', 
  `postalcoderegioncode` string COMMENT '', 
  `replayed` boolean COMMENT '', 
  `type` string COMMENT '', 
  `payload` string COMMENT '', 
  `timezoneoffset` int COMMENT '', 
  `trip` string COMMENT '', 
  `beacon` string COMMENT '', 
  `foreground` boolean COMMENT '', 
  `locationaccuracyauthorization` string COMMENT '', 
  `locationauthorization` string COMMENT '', 
  `devicetype` string COMMENT '', 
  `devicemake` string COMMENT '', 
  `devicemodel` string COMMENT '', 
  `deviceos` string COMMENT '', 
  `sdkversion` string COMMENT '', 
  `stopped` boolean COMMENT '', 
  `beacontag` string COMMENT '', 
  `beaconexternalid` string COMMENT '', 
  `beacontype` string COMMENT '', 
  `beaconmajor` string COMMENT '', 
  `beaconminor` string COMMENT '', 
  `beaconuuid` string COMMENT '', 
  `tripexternalid` string COMMENT '', 
  `tripdestinationgeofencetag` string COMMENT '', 
  `tripdestinationgeofenceexternalid` string COMMENT '')
```

RadarSDK/RadarAPIClient.m
david-goodfellow: they can also give an optional `confidence`, `duration`, and `createdAt` time. The event name goes by `type` to be aligned with the other endpoints in the server. The endpoint can be seen here https://github.com/radarlabs/server/pull/4144/files

RadarSDK/Include/Radar.h
jsani-radar: helpful to have the schema, thanks for providing

yeah and no worries re: backfill, this is a net new capability for custom events, as existing events don't have any `metadata` attached - just needs to be there going forward

RadarSDK/RadarAPIClient.m
jsani-radar: per our decision this week I'm passing the `user` from `/track` to denorm all the necessary fields onto the custom event

jrtibbetts:
@jsani-radar, do you need to do anything else in this PR before it's merged?

RadarSDK/Radar.m
```diff
@@ -711,6 +711,66 @@ + (void)getMatrixFromOrigins:(NSArray<CLLocation *> *_Nonnull)origins
...
+                [RadarUtils runOnMainThread:^{
+                    completionHandler(status, nil, nil, nil);
+                }];
+            }
+
+            return;
+        }
+
+        [[RadarAPIClient sharedInstance] sendEvent:name withMetadata:metadata user:user completionHandler:^(RadarStatus status, NSDictionary * _Nullable res, RadarEvent * _Nullable event) {
+            if (completionHandler) {
```
jrtibbetts, line 762: What if `status != RadarStatusSuccess` here?

RadarSDK/Radar.m
```diff
@@ -711,6 +711,66 @@ + (void)getMatrixFromOrigins:(NSArray<CLLocation *> *_Nonnull)origins
...
```
jrtibbetts, line 733: What if `status != RadarStatusSuccess` here?

RadarSDK/Radar.m
```diff
@@ -711,6 +711,66 @@ + (void)getMatrixFromOrigins:(NSArray<CLLocation *> *_Nonnull)origins
...
+                NSMutableArray *finalEvents = [NSMutableArray arrayWithArray:events];
+                [finalEvents insertObject:event atIndex:0];
+
+                [RadarUtils runOnMainThread:^{
+                    completionHandler(status, location, finalEvents, user);
```
jrtibbetts, line 738: This can either be

```
NSArray *finalEvents = @[events[0]]
 
[RadarUtils runOnMainThread:^{
    completionHandler(status, location, finalEvents, user);
```

or just 

```
[RadarUtils runOnMainThread:^{
    completionHandler(status, location, @[events[0], user);
```

You may also want to check whether `events` is empty before accessing the zeroth element, or else you'll get an `NSRangeException`.

RadarSDK/Radar.m
```diff
@@ -711,6 +711,66 @@ + (void)getMatrixFromOrigins:(NSArray<CLLocation *> *_Nonnull)origins
...
+                }];
```
jrtibbetts, line 768: See my comment on line 738.

RadarSDK/Radar.m
jsani-radar: great call, fixed

RadarSDK/Include/Radar.h
jsani-radar: https://github.com/radarlabs/documentation/pull/156

RadarSDK/Radar.m
```diff
@@ -711,6 +711,93 @@ + (void)getMatrixFromOrigins:(NSArray<CLLocation *> *_Nonnull)origins
                                         }];
 }
 
++ (void)sendEvent:(NSString *)name
+     withMetadata:(NSDictionary *_Nullable)metadata
+completionHandler:(RadarSendEventCompletionHandler)completionHandler {
+    [self trackOnceWithCompletionHandler:^(RadarStatus status, CLLocation * _Nullable location, NSArray<RadarEvent *> * _Nullable events, RadarUser * _Nullable user) {
+        if (status != RadarStatusSuccess) {
+            if (completionHandler) {
```
jrtibbetts, line 719: The `RadarSendEventCompletionHandler` is marked as `NONNULL`, so we don't need these `if (completionHandler)` checks.

RadarSDK/RadarAPIClient.m
```diff
@@ -932,6 +932,51 @@ - (void)getMatrixFromOrigins:(NSArray<CLLocation *> *)origins
...
+    NSString *host = [RadarSettings host];
+    NSString *url = [NSString stringWithFormat:@"%@/v1/events", host];
+    url = [url stringByAddingPercentEncodingWithAllowedCharacters:[NSCharacterSet URLQueryAllowedCharacterSet]];
+
+    NSDictionary *headers = [RadarAPIClient headersWithPublishableKey:publishableKey];
+
+    [self.apiHelper requestWithMethod:@"POST"
+                                  url:url
+                              headers:headers
+                               params:nil
```
andlum, line 962: Should this be `params: params` instead of `params:nil` since this is a post request?

andlum: Yep! This was the source of the bad request error. I changed this to `params` and got a success:
```
Send event: status = SUCCESS; location = Optional(<+37.78583400,-122.40641700> +/- 5.00m (speed -1.00 mps / course -1.00) @ 7/19/22, 11:24:01 AM Eastern Daylight Time); events = Optional([<RadarEvent: 0x600001e05100>]); user = Optional(<RadarUser: 0x60000077c420>)
```

RadarSDK/Include/Radar.h
```diff
@@ -765,6 +774,34 @@ typedef void (^_Nonnull RadarRouteMatrixCompletionHandler)(RadarStatus status, R
...
+/**
+ Sends a custom event.
+
+ @param name The name of the event.
+ @param metadata The metadata associated with the event.
+ @param completionHandler A completion handler.
+
+ @see https://radar.com/documentation/api#create-a-custom-event
+ */
++ (void)sendEvent:(NSString *)name
```
nickpatrick, line 786: Are we overloading `event.type`, or is this a new `event.name` field?

RadarSDK/Radar.m
```diff
@@ -711,6 +711,93 @@ + (void)getMatrixFromOrigins:(NSArray<CLLocation *> *_Nonnull)origins
...
+                    [RadarUtils runOnMainThread:^{
+                        completionHandler(status, nil, nil, nil);
+                    }];
+                }
+
+                return;
+            }
+
+            if (completionHandler) {
+                // construct the array of events to return in the callback - custom event at index 0, followed by track events
```
nickpatrick, line 740: Feels a little messy. Could/should this happen in the API client?

RadarSDK/Include/Radar.h
```diff
@@ -765,6 +774,34 @@ typedef void (^_Nonnull RadarRouteMatrixCompletionHandler)(RadarStatus status, R
...
            completionHandler:(RadarRouteMatrixCompletionHandler)completionHandler NS_SWIFT_NAME(getMatrix(origins:destinations:mode:units:completionHandler:));
 
+/**
+ Sends a custom event.
+
+ @param name The name of the event.
+ @param metadata The metadata associated with the event.
+ @param completionHandler A completion handler.
+
+ @see https://radar.com/documentation/api#create-a-custom-event
```
nickpatrick, line 784: Create/send/log? Should we use send everywhere? Will it be clear to developers that you "generate" events like geofence entries but "send" custom events?

RadarSDK/Include/Radar.h
jsani-radar: will rename for consistency

RadarSDK/Include/Radar.h
jsani-radar: good call - will amend here and in docs to make the difference explicit

RadarSDK/Radar.m
jsani-radar: good point, prob shouldn't live in the most exposed SDK file - I guess there's a q of whether this should happen on the client at all, should the server be aggregating events instead?

RadarSDK/Radar.m
jsani-radar: makes sense, prob a cleanup item within a larger refactor as this mismatch exists elsewhere in the SDK, but out of scope for this PR

RadarSDK/Include/Radar.h
```diff
@@ -765,6 +774,34 @@ typedef void (^_Nonnull RadarRouteMatrixCompletionHandler)(RadarStatus status, R
...
+/**
+ Sends a custom event.
+
+ @param customType The user-defined type of the event.
+ @param metadata The metadata associated with the event.
+ @param completionHandler A completion handler.
+
+ @see https://radar.com/documentation/api#send-a-custom-event
+ */
++ (void)sendEvent:(NSString *)customType
```
jsani-radar, line 786: do you think `customType` is a better named parameter than `name`? `event.customType` can be an internal impl detail and doesn't necessarily need to be indicated to the caller

I'm also fine if you believe strongly that `customType` is better - at the end of the day, it is overwriting `event.type`

RadarSDK/RadarAPIClient.h
```diff
@@ -52,6 +52,9 @@ typedef void (^_Nonnull RadarDistanceAPICompletionHandler)(RadarStatus status, N
 typedef void (^_Nonnull RadarMatrixAPICompletionHandler)(RadarStatus status, NSDictionary *_Nullable res, RadarRouteMatrix *_Nullable matrix);
 
 typedef void (^_Nonnull RadarConfigAPICompletionHandler)(RadarStatus status, RadarMeta *_Nullable meta);
+
+typedef void (^_Nonnull RadarSendEventsAPICompletionHandler)(RadarStatus status, NSDictionary *_Nullable res, NSArray<RadarEvent *> *_Nullable event);
```
jsani-radar, line 56: nit: my bad, this should probably be singular `RadarSendEventAPICompletionHandler`, as we're only supporting sending one event, even though the callback comes back with an array (since events from `/track` are included)

RadarSDKTests/RadarAPIHelperMock.h
```diff
@@ -16,6 +16,9 @@ NS_ASSUME_NONNULL_BEGIN
 @property (assign, nonatomic) RadarStatus mockStatus;
 @property (nonnull, strong, nonatomic) NSDictionary *mockResponse;
 
+- (void)setMockResponse:(NSDictionary *)response
```
jsani-radar, line 19: nice, this is a solid addition

jsani-radar (commented):
looks good on my end, will want final 👀 from @nickpatrick and @tjulien before 🚀

RadarSDK/Include/Radar.h
jrtibbetts: That's what @nickpatrick suggested, and I'm ok with it.

RadarSDK/RadarAPIClient.h
jrtibbetts: Ok, I'll change it.

jrtibbetts: Done.

tjulien (approved):

PR #191, opened 2022-07-07
sc-10902: Fix radar event raw strings

jrtibbetts:
[sc-10902: [iOS SDK] Fix the RadarEventTypeUserExitedRegionDMA raw string](https://app.shortcut.com/radarlabs/story/10902/ios-sdk-fix-the-radareventtypeuserexitedregiondma-raw-string)

* `RadarEvent.RadarEventTypeUserExitedRegionDMA`'s raw string type is now `user.exited_region_dma` instead of `user.exited_region_country`.
* Rearranged the order of `RadarEvent.initializeWithObject(:)` and `RadarEvent.stringForType(:)` to match the order in the server's event constants file.

### DIFF
--- a/RadarSDK/RadarEvent.m
+++ b/RadarSDK/RadarEvent.m
@@ -121,48 +121,50 @@
     if (typeObj && [typeObj isKindOfClass:[NSString class]]) {
         NSString *typeStr = (NSString *)typeObj;
 
+        // These strings should match the values (and order) of the server's event
+        // constants.
         if ([typeStr isEqualToString:@"user.entered_geofence"]) {
             type = RadarEventTypeUserEnteredGeofence;
         } else if ([typeStr isEqualToString:@"user.exited_geofence"]) {
             type = RadarEventTypeUserExitedGeofence;
+        } else if ([typeStr isEqualToString:@"user.dwelled_in_geofence"]) {
+            type = RadarEventTypeUserDwelledInGeofence;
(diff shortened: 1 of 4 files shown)


RadarSDK/RadarEvent.m
```diff
@@ -121,48 +121,50 @@ - (instancetype _Nullable)initWithObject:(id)object {
...
             type = RadarEventTypeUserExitedRegionDMA;
+        } else if ([typeStr isEqualToString:@"user.entered_region_state"]) {
+            type = RadarEventTypeUserEnteredRegionState;
+        } else if ([typeStr isEqualToString:@"user.exited_region_state"]) {
+            type = RadarEventTypeUserExitedRegionState;
+        } else if ([typeStr isEqualToString:@"user.entered_region_postal_code"]) {
+            type = RadarEventTypeUserEnteredRegionPostalCode;
+        } else if ([typeStr isEqualToString:@"user.exited_region_postal_code"]) {
+            type = RadarEventTypeUserExitedRegionPostalCode;
+        } else if ([typeStr isEqualToString:@"user.nearby_place_chain"]) {
```
jsani-radar, line 152: think we should keep this one grouped with the other place events given the context type

jsani-radar (approved):

RadarSDK/RadarEvent.m
jrtibbetts: I only moved it because that's where it is in the server's file.

PR #194, opened 2022-08-02
sc-11053: Improve the Example app

jrtibbetts:
*Note: converted to a draft PR because it includes the #196 changes.*

[sc-11053: [iOS SDK] Example app cleanup](https://app.shortcut.com/radarlabs/story/11053/ios-sdk-example-app-cleanup)

* Renamed the Example app to Radar Demo.
* Instead of popping up notifications every time an event is received or a location is updated, every message is displayed with a timestamp in a simple table.
* The Radar SDK calls have been put into the `AppDelegate.runDemo()` function, and they're run in the same order that they're listed in the 
[Radar SDK documentation](https://radar.com/documentation/sdk/ios).

https://user-images.githubusercontent.com/708510/182423216-40512794-4927-4fe3-82ca-9c7d79be48a9.mp4


RadarSDK/Include/Radar.h
```diff
@@ -615,7 +615,32 @@ completionHandler:(RadarSendEventCompletionHandler)completionHandler NS_SWIFT_NA
...
+ @param categories An array of categories to filter. See: https://radar.com/documentation/places/categories
+ @param groups An array of groups to filter. See https://radar.com/documentation/places/groups
+ @param limit The max number of places to return. A number between 1 and 100.
+ @param completionHandler A completion handler.
+
+ @see https://radar.com/documentation/api#search-places
+ */
++ (void)searchPlacesWithRadius:(int)radius
+                        chains:(NSArray<NSString *> *_Nullable)chains
+                 chainMetadata:(NSDictionary<NSString *, NSString *> *_Nullable)chainMetadata
```
jsani-radar, line 638: do these changes belong here? rebase needed?

jsani-radar (commented):
nice!! I really like these changes - makes things more visual / explicit, think this can be super helpful for someone who's just getting started with the SDK

a few nits:
- I think it'd be helpful to have some demarcation between successive calls - something like `=========` or `---------`, your call
- the last event says `type = unknown` - is deserializing the `customType` happening correctly?
- do we want to indicate somewhere that we're running through all the SDK functions with the pasted in pub key? for fresh eyes this might not be obvious

looking solid overall!

RadarSDK/Include/Radar.h
jrtibbetts: Yeah, I marked this as a draft PR because it includes all of the `chainMetadata` changes from PR #196. Once 196 is merged, then I'll remove the draft status.

jrtibbetts:
> I think it'd be helpful to have some demarcation between successive calls - something like ========= or ---------, your call

Do you mean between where they're called in `runDemo()`, or between their declarations?

> The last event says type = unknown - is deserializing the customType happening correctly?

I thought I fixed that in commit #8fb676. Maybe it needs to be added somewhere else, too.

> do we want to indicate somewhere that we're running through all the SDK functions with the pasted in pub key? for fresh eyes this might not be obvious

Yeah, I buried that in the `initializeSDK()` call. Would it be enough to make it a top-level local variable, or a property on `AppDelegate`?

jsani-radar:
I'm mainly assessing this from the perspective of someone who's just cloned the repo and running the Example app for the first time, viewing the output logs and trying to understand what happened

> Do you mean between where they're called in runDemo(), or between their declarations?

Between output logs for each function - it'd be helpful to have a demarcation of where one call starts / ends

> Yeah, I buried that in the initializeSDK() call. Would it be enough to make it a top-level local variable, or a property on AppDelegate?

Not sure what you mean in terms of placing it in code - I was thinking an initial log before the APIs are called, or indicating somewhere in the permission prompt

PR #198, opened 2022-08-12
sc-11410: Radar.sendEvent() should fail if the server returns no events

jrtibbetts:
[sc-11410: iOS should fail with a server error if POST /events doesn't return an event](https://app.shortcut.com/radarlabs/story/11410/ios-should-fail-with-a-server-error-if-post-events-doesn-t-return-an-event)

* If `RadarApiClient.sendEvent()` returns a `200` response, but contains no events, then a server error is passed to the completion handler and logged as an error.


RadarSDK/RadarAPIClient.m
```diff
@@ -970,7 +970,15 @@ - (void)sendEvent:(NSString *)customType
 
         id eventObj = res[@"event"];
         RadarEvent *customEvent = [[RadarEvent alloc] initWithObject:eventObj];
-        if (customEvent) {
+
+        if (!customEvent) {
+            // If the server didn't send back an event, then there was a
+            // problem.
+            [[RadarLogger sharedInstance] logWithLevel:RadarLogLevelError
+                                               message:@"POST /event did not return a new event"];
```
jsani-radar, line 978: nit: `/events`

jsani-radar (approved):
small nit, otherwise lgtm!

RadarSDK/RadarAPIClient.m
jrtibbetts: Good catch.

jrtibbetts: Fixed.

PR #199, opened 2022-08-16
sc-11337: Add metadata checks for custom events

jrtibbetts:
[sc-11337: [iOS SDK] Add metadata checks for Radar.sendEvent() unit tests](https://app.shortcut.com/radarlabs/story/11337/ios-sdk-add-metadata-checks-for-radar-sendevent-unit-tests)

* Added unit tests and Example app checks for `RadarEvent.metadata` parsing, now that the server is sending back metadata as JSON data, not a string.

### DIFF
--- a/RadarSDK/RadarEvent.m
+++ b/RadarSDK/RadarEvent.m
@@ -291,11 +291,11 @@
                                              verticalAccuracy:-1
                                                     timestamp:(createdAt ? createdAt : [NSDate date])];
         }
+    }
 
-        id metadataObj = dict[@"metadata"];
-        if (metadataObj && [metadataObj isKindOfClass:[NSDictionary class]]) {
-            metadata = (NSDictionary *)metadataObj;
-        }
+    id metadataObj = dict[@"metadata"];
+    if (metadataObj && [metadataObj isKindOfClass:[NSDictionary class]]) {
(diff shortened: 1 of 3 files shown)


RadarSDK/RadarEvent.m
```diff
@@ -291,11 +291,11 @@ - (instancetype _Nullable)initWithObject:(id)object {
                                              verticalAccuracy:-1
                                                     timestamp:(createdAt ? createdAt : [NSDate date])];
         }
+    }
 
-        id metadataObj = dict[@"metadata"];
```
jsani-radar, line 295: was this happening in the `locationObj` block before, hence the extra `}`? my fault if so 🤦🏽‍♂️

jsani-radar (approved):
nice, thanks for adding these checks!

RadarSDK/RadarEvent.m
jrtibbetts: Yeah, it was. It took me a while to figure out why the breakpoint I set on it wasn't being hit. :)

jsani-radar: theoretically shouldn't `metadata` still have been serialized? since the check was happening anyways, just nested within another conditional?

the fix is necessary and important, but I don't think things were "broken" beforehand?

PR #210, opened 2022-09-20
sc-11497: New trip functions

jrtibbetts:
[sc-11497: [iOS SDK] New trip functions](https://app.shortcut.com/radarlabs/story/11497)

* Added `+[Radar startTrip:trackingOptions:completionBlock:]`, which starts tracking automatically after the trip has successfully started. When canceling or completing a trip, it also resets the pre-trip tracking options.


RadarSDK.xcodeproj/xcshareddata/xcschemes/XCFramework.xcscheme
```diff
@@ -0,0 +1,77 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<Scheme
+   LastUpgradeVersion = "1330"
+   version = "1.3">
```
jrtibbetts, line 4: Accidentally removed this file in a recent change.

RadarSDK/Include/RadarTrackingOptions.h
```diff
@@ -179,7 +179,7 @@ typedef NS_ENUM(NSInteger, RadarTrackingOptionsSyncLocations) {
 + (RadarTrackingOptionsReplay)replayForString:(NSString *)str;
 + (NSString *)stringForSyncLocations:(RadarTrackingOptionsSyncLocations)syncLocations;
 + (RadarTrackingOptionsSyncLocations)syncLocationsForString:(NSString *)str;
-+ (RadarTrackingOptions *_Nonnull)trackingOptionsFromDictionary:(NSDictionary *_Nonnull)dictionary;
++ (RadarTrackingOptions *_Nullable)trackingOptionsFromDictionary:(NSDictionary *_Nonnull)dictionary;
```
jrtibbetts, line 182: `nil` `trackingOptions` now indicate that a trip is _not_ in progress, so I had to relax this constraint.

RadarSDK/RadarAPIClient.m
```diff
@@ -198,22 +199,6 @@ - (void)trackWithLocation:(CLLocation *_Nonnull)location
...
-        }
-        if (tripOptions.destinationGeofenceTag) {
-            tripOptionsDict[@"destinationGeofenceTag"] = tripOptions.destinationGeofenceTag;
-        }
-        if (tripOptions.destinationGeofenceExternalId) {
-            tripOptionsDict[@"destinationGeofenceExternalId"] = tripOptions.destinationGeofenceExternalId;
-        }
-        tripOptionsDict[@"mode"] = [Radar stringForMode:tripOptions.mode];
-        params[@"tripOptions"] = tripOptionsDict;
-    }
```
jrtibbetts, line 216: This has been factored out into a new `-[RadarAPIClient paramsFromTripOptions:]` function.

RadarSDK/RadarAPIClient.m
```diff
@@ -386,29 +421,16 @@ - (void)updateTripWithOptions:(RadarTripOptions *)options status:(RadarTripStatu
...
-    if (options.destinationGeofenceExternalId) {
-        params[@"destinationGeofenceExternalId"] = options.destinationGeofenceExternalId;
-    }
-    params[@"mode"] = [Radar stringForMode:options.mode];
-    if (options.scheduledArrivalAt) {
-        params[@"scheduledArrivalAt"] = [[RadarUtils isoDateFormatter] stringFromDate:options.scheduledArrivalAt];
-    }
-    if (options.approachingThreshold > 0) {
-        params[@"approachingThreshold"] = [NSString stringWithFormat:@"%d", options.approachingThreshold];
-    }
```
jrtibbetts, line 409: This is now handled by the `paramsFromTripOptions` call a few lines above.

RadarSDK/RadarSettings.h
```diff
@@ -36,6 +36,10 @@ NS_ASSUME_NONNULL_BEGIN
 + (void)setTracking:(BOOL)tracking;
 + (RadarTrackingOptions *_Nullable)trackingOptions;
 + (void)setTrackingOptions:(RadarTrackingOptions *_Nonnull)options;
++ (void)removeTrackingOptions;
++ (RadarTrackingOptions *_Nullable)previousTrackingOptions;
++ (void)setPreviousTrackingOptions:(RadarTrackingOptions *_Nonnull)options;
++ (void)removePreviousTrackingOptions;
```
jrtibbetts, line 42: I would prefer to just make `+[RadarSettings setPreviousTrackingOptions:]` take a nullable argument to indicate removal, but this way follows the pattern of the other `trackingOptions` settings.

RadarSDK/RadarSettings.m
```diff
@@ -126,36 +127,78 @@ + (void)setTracking:(BOOL)tracking {
...
++ (void)removePreviousTrackingOptions {
+    [[NSUserDefaults standardUserDefaults] removeObjectForKey:kPreviousTrackingOptions];
 }
 
 + (RadarTrackingOptions *_Nullable)remoteTrackingOptions {
     NSDictionary *optionsDict = [[NSUserDefaults standardUserDefaults] dictionaryForKey:kRemoteTrackingOptions];
-    return optionsDict ? [RadarTrackingOptions trackingOptionsFromDictionary:optionsDict] : nil;
+
+    if (optionsDict != nil) {
+        return optionsDict ? [RadarTrackingOptions trackingOptionsFromDictionary:optionsDict] : nil;
```
jrtibbetts, line 170: Technically unnecessary since I added the `nil` check in `+[RadarTrackingOptions trackingOptionsFromDictionary:]`, but I think that this is a little clearer.

RadarSDK/Include/Radar.h
```diff
@@ -522,7 +522,8 @@ completionHandler:(RadarSendEventCompletionHandler)completionHandler NS_SWIFT_NA
 
  @see https://radar.com/documentation/trip-tracking
  */
-+ (void)startTripWithOptions:(RadarTripOptions *_Nonnull)options NS_SWIFT_NAME(startTrip(options:));
++ (void)startTripWithOptions:(RadarTripOptions *_Nonnull)options
+    NS_SWIFT_NAME(startTrip(options:)) __deprecated_msg("Use +[Radar startTripWithOptions:trackingOptions:completionHandler:] instead.");
```
nickpatrick, line 526: @corypisano @jrtibbetts Are we deprecating this on Android as well? Do we need to?

RadarSDK/RadarAPIClient.m
nickpatrick: @jrtibbetts Probably a good move, but this now makes it harder to eyeball the params being passed on iOS vs. Android. Do we want to make the analogous change on Android? @corypisano

RadarSDK/Include/Radar.h
```diff
@@ -533,7 +534,13 @@ completionHandler:(RadarSendEventCompletionHandler)completionHandler NS_SWIFT_NA
  @see https://radar.com/documentation/trip-tracking
  */
 + (void)startTripWithOptions:(RadarTripOptions *_Nonnull)options
-           completionHandler:(RadarTripCompletionHandler _Nullable)completionHandler NS_SWIFT_NAME(startTrip(options:completionHandler:));
+           completionHandler:(RadarTripCompletionHandler _Nullable)completionHandler
+    NS_SWIFT_NAME(startTrip(options:completionHandler:)) __deprecated_msg("Use +[Radar startTripWithOptions:trackingOptions:completionHandler:] instead.");
+
```
nickpatrick, line 539: @jrtibbetts Needs docs. Let's make them consistent with Android and incorporate this feedback in both places: https://github.com/radarlabs/radar-sdk-android/pull/236/files#r988873895 @corypisano

RadarSDK/RadarSettings.h
nickpatrick: Generally prefer consistency here, internally and with Android. @corypisano

RadarSDK/Include/Radar.h
jrtibbetts: I'll un-deprecate it.

RadarSDK/Include/Radar.h
jrtibbetts: Added docs.

RadarSDK/RadarAPIClient.m
jrtibbetts: I put it back in and dropped the `paramsFromTripOptions`.

RadarSDK/RadarAPIClient.m
```diff
@@ -198,21 +199,18 @@ - (void)trackWithLocation:(CLLocation *_Nonnull)location
...
-        if (tripOptions.destinationGeofenceTag) {
-            tripOptionsDict[@"destinationGeofenceTag"] = tripOptions.destinationGeofenceTag;
-        }
-        if (tripOptions.destinationGeofenceExternalId) {
-            tripOptionsDict[@"destinationGeofenceExternalId"] = tripOptions.destinationGeofenceExternalId;
-        }
-        tripOptionsDict[@"mode"] = [Radar stringForMode:tripOptions.mode];
-        params[@"tripOptions"] = tripOptionsDict;
+        NSMutableDictionary *tripParams = [NSMutableDictionary new];
+        tripParams[@"v2"] = @(YES);
```
corypisano, line 207: note: will update to `version: '2'`

RadarSDK/RadarAPIClient.m
```diff
@@ -325,7 +323,9 @@ - (void)trackWithLocation:(CLLocation *_Nonnull)location
                         if (events && user) {
                             [RadarSettings setId:user._id];
                             
-                            if (!user.trip) {
+                            // if user was on a trip that ended server side, restore previous tracking options
+                            if (!user.trip && [RadarSettings tripOptions]) {
```
tjulien, line 327: @corypisano @nickpatrick 

I realize we're building on the existing `if (!user.trip)` logic - but should this be gated behind `if (events)` (line 323) or just  `if (user)`?  I see cases when `events` could be `nil` but we still would want to run this logic.  Maybe in practice events is always at least `[]`and never `nil`?

RadarSDK/RadarLocationManager.m
```diff
@@ -396,6 +396,19 @@ - (void)updateTrackingFromMeta:(RadarMeta *_Nullable)meta {
...
 }
 
+- (void)restartPreviousTrackingOptions {
+    RadarTrackingOptions *previousTrackingOptions = [RadarSettings previousTrackingOptions];
+    [[RadarLogger sharedInstance] logWithLevel:RadarLogLevelDebug message:@"Restarting previous tracking options"];
+
+    if (previousTrackingOptions) {
+        [Radar startTrackingWithOptions:previousTrackingOptions];
+    } else {
+        [Radar stopTracking];
```
tjulien, line 406: @corypisano @nickpatrick 

So one thing I notice about the `restartPreviousTrackingOptions` function is that it's called from a few places (cancel, complete, and track) and it's not resilient to multiple of those being called.  The second call will call `stopTracking`.  Maybe this is fine in the happy path but I could imagine it causing issues.  Maybe this is something we're fine with for this release and we iterate on it in a future version.

RadarSDK/Radar.m
```diff
@@ -420,22 +420,37 @@ + (void)startTripWithOptions:(RadarTripOptions *)options {
...
+            // flush location update to generate events
+            [[RadarLocationManager sharedInstance] getLocationWithCompletionHandler:nil];
+        }
+
+        if (completionHandler) {
+            [RadarUtils runOnMainThread:^{
+                completionHandler(status, trip, events);
+            }];
+        }
+    }];
```
tjulien, line 453: @corypisano @nickpatrick 

I want to call out two thing with this code block:
1) We're starting tracking and doing `getLocationWithCompletionHandler` in parallel.  Should be ok but - any concerns about hitting rate limits or race conditions with that?
2) We're calling `completionHandler` in parallel with both the calls to `startTracking` and `getLocationWithCompletionHandler`.  Should be ok since call to `createTrip` has completed, which is the expectation for calling the `completionHandler`.  Just wanted to raise awareness of that.

corypisano: 1 - hm not concerned with race condition as we're waiting for start trip to return now, but rate limit will squint

for 2 - getLocationWithCompletionHandler is passed in nil so shouldn't be calling completionHandler

tjulien (approved):

PR #214, opened 2022-12-19
Include approachingThreshold in tripOptions

lmeier:
Part of the work for React Native parity https://linear.app/radarlabs/issue/MOB-56/include-trip-approaching-threshold-in-radartripoptions-helper

New to native iOS dev, but I think this is the meat of it? Existing tests pass as well. Will poke around a bit more tomorrow before opening this up

### DIFF
--- a/RadarSDK/Include/RadarTripOptions.h
+++ b/RadarSDK/Include/RadarTripOptions.h
@@ -57,7 +57,7 @@ NS_ASSUME_NONNULL_BEGIN
  */
 @property (nonatomic, assign) RadarRouteMode mode;
 
-@property (nonatomic, assign) UInt8 approachingThreshold;
+@property (nonatomic, assign) UInt16 approachingThreshold;
 
 + (RadarTripOptions *_Nullable)tripOptionsFromDictionary:(NSDictionary *)dict;
 - (NSDictionary *)dictionaryValue;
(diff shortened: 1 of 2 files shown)


RadarSDK/Include/RadarTripOptions.h
```diff
@@ -57,7 +57,7 @@ NS_ASSUME_NONNULL_BEGIN
  */
 @property (nonatomic, assign) RadarRouteMode mode;
 
-@property (nonatomic, assign) UInt8 approachingThreshold;
+@property (nonatomic, assign) UInt16 approachingThreshold;
```
lmeier, line 60: We don't limit this number server side, so I imagine we'd want to a) allow more bits, b) think about a server side limit

RadarSDK/RadarTripOptions.m
```diff
@@ -77,6 +79,7 @@ - (NSDictionary *)dictionaryValue {
     dict[kDestinationGeofenceExternalId] = self.destinationGeofenceExternalId;
     dict[kMode] = [Radar stringForMode:self.mode];
     dict[kScheduledArrivalAt] = self.scheduledArrivalAt;
+    dict[kApproachingThreshold] = @(self.approachingThreshold);
```
tjulien, line 82: Need to add this to `isEqual` below

lmeier: durgh, had that in my local commit history. Just pushed up

tjulien (approved):

PR #215, opened 2023-01-04
Location verification

nickpatrick:
- Exposes `trackVerified()`
- Exposes new `RadarFraud` flags
- Uses `DeviceCheck` to verify app and device integrity (https://developer.apple.com/documentation/devicecheck)
- Exposes `setVerifiedHost()` and uses App Transport Security to support SSL pinning (https://developer.apple.com/news/?id=g9ejcf8y)
- Adds `RadarConfig` object for consistency with Android
- Ran `make format` and cleaned up some docs


RadarSDK/RadarAPIClient.m
```diff
@@ -265,30 +274,34 @@ - (void)trackWithLocation:(CLLocation *_Nonnull)location
...
+    NSMutableDictionary *requestParams = [NSMutableDictionary new];
 
     BOOL replaying = options.replay == RadarTrackingOptionsReplayAll && replayCount > 0;
     
-    if (replaying) {
+    if (verified) {
+        url = [NSString stringWithFormat:@"%@/v1/track", host];
+        
+        requestParams = [params mutableCopy];
+    } else if (replaying) {
```
lmeier, line 285: Hmmm, given that the `if (verified)` clause and the `else` are the same, it would be DRYer if we could just check `replaying` first.

`replaying` is only true iff `replay: all` and there are replays waiting to be uploaded (line 279 just above), so I imagine it would be fine to do it that way.

Up to you.

lmeier: alternatively, to keep the order but make it dry, `if (verified || !replaying) ... else`

RadarSDK/RadarAPIClient.h
```diff
@@ -65,27 +64,32 @@ typedef void (^_Nonnull RadarSyncLogsAPICompletionHandler)(RadarStatus status);
 
 + (NSDictionary *)headersWithPublishableKey:(NSString *)publishableKey;
 
-+ (RadarMeta *_Nullable)parseMeta:(NSDictionary *_Nullable)res;
```
lmeier, line 68: I like the rename from meta 👍

RadarSDK/RadarAPIHelper.m
```diff
@@ -67,11 +67,13 @@ - (void)requestWithMethod:(NSString *)method
...
             }
 
-            NSURLSessionConfiguration *configuration = [NSURLSessionConfiguration defaultSessionConfiguration];
+            NSURLSessionConfiguration *configuration;
             if (extendedTimeout) {
+                configuration = [NSURLSessionConfiguration defaultSessionConfiguration];
                 configuration.timeoutIntervalForRequest = 25;
                 configuration.timeoutIntervalForResource = 25;
             } else {
+                configuration = [NSURLSessionConfiguration ephemeralSessionConfiguration];
```
lmeier, line 76: Is this done for the purposes of `verified`? If so, a brief comment explaining why would be helpful. (Not obvious to me why we alternate between default and ephemeral)

lmeier (approved):
small comments, should get @tjulien's eyes as well but lgtm

RadarSDK/RadarAPIClient.h
nickpatrick: Yeah, tried to make this consistent across iOS and Android

RadarSDK/RadarAPIHelper.m
nickpatrick: Will do

nickpatrick: Added

RadarSDK/RadarAPIClient.m
nickpatrick: @lmeier This simplifies the code a bit and also makes it more line-by-line consistent with Android: https://github.com/radarlabs/radar-sdk-ios/pull/215/commits/43fd39c8ccf38bbb38970157018bfd2014761898#diff-57a83f22ae98eebd6b400f6f7baecfb73b170d06629cbf8718e0a9000069d5a8R272

lmeier: Nice! I think we want to clear out `requestParams` in that if statement, as the server is expecting params in the format:
```
{ "replays": [ {...}, {...}, ...]}
```

PR #216, opened 2023-01-20
Offline replay v1

lmeier:
For: https://linear.app/radarlabs/issue/MOB-70/offline-replay-for-ios

This PR implements an in-memory replay buffer that stores up to the last 120 / one-hour of track calls.

### DIFF
--- a/RadarSDK/RadarAPIClient.m
+++ b/RadarSDK/RadarAPIClient.m
@@ -20,6 +20,8 @@
 #import "RadarLogger.h"
 #import "RadarMeta+Internal.h"
 #import "RadarPlace+Internal.h"
+#import "RadarReplay.h"
+#import "RadarReplayBuffer.h"
 #import "RadarRouteMatrix+Internal.h"
 #import "RadarRoutes+Internal.h"
 #import "RadarSettings.h"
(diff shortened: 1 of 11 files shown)


lmeier:
Throwing up a draft of my first pass at implementing offline replay to get a check on general implementation approach. If you turn on continuous mode and point it at my dev server via toolkit, it should work.

One interesting thing I've come across (possibly introduced here, possibly an unresolved bug): if you go offline for an hour, then go back online, the `POST /v1/logs` returns a 413 / Payload too large, and repeatedly tries to upload this too large log payload without discard logic seemingly.

@tjulien would love to get some feedback at your leisure.

RadarSDK/RadarAPIClient.m
```diff
@@ -244,106 +333,66 @@ - (void)trackWithLocation:(CLLocation *_Nonnull)location
...
-                            [[RadarDelegateHolder sharedInstance] didFailWithStatus:status];
-
-                            return completionHandler(status, nil, nil, nil, nil, nil);
-                        }
-
-                        [Radar flushLogs];
-                        [RadarState setLastFailedStoppedLocation:nil];
-
-                        RadarMeta *_Nullable meta = [RadarAPIClient parseMeta:res];
+    if (options.replay == RadarTrackingOptionsReplayAll) {
```
tjulien, line 336: Think we probably want: ` && replay buffer size > 0`, so regular location updates would still use `/track`

RadarSDK/RadarAPIClient.m
```diff
@@ -244,106 +333,66 @@ - (void)trackWithLocation:(CLLocation *_Nonnull)location
...
+        NSString *bulkUrl = [NSString stringWithFormat:@"%@/v1/track/bulk", host];
+        bulkUrl = [bulkUrl stringByAddingPercentEncodingWithAllowedCharacters:[NSCharacterSet URLQueryAllowedCharacterSet]];
 
-                            return completionHandler(RadarStatusSuccess, res, events, user, nearbyGeofences, meta);
-                        }
-
-                        [[RadarDelegateHolder sharedInstance] didFailWithStatus:status];
+        NSDictionary *trackParamsDict = @{@"tracks": trackParams};
+        
+        [self.apiHelper requestWithMethod:@"POST"
```
tjulien, line 350: I think it will be a little simpler if we can just have a single `[self.apiHelper requestWithMethod ...]` call and above it pick `/track` vs `/track/bulk` and set up the other params.

Then when we're dealing with failed requests below we can check `options.replay` and handle  `RadarTrackingOptionsReplayStops` vs `RadarTrackingOptionsReplayAll`.

RadarSDK/RadarAPIClient.m
lmeier: The thinking of omitting that condition was that if we do this, we'll need to add conditional logic about adding failed calls to the buffer in the `/track` callback.

But it makes sense to get as many calls as possible the tried & tested `/track` codepath, so I'll go ahead and make this change.

tjulien (approved):

PR #226, opened 2023-02-15
CORE-227 Change signature of custom events function

joeychik-radar:
### Summary of changes
- Renamed `sendEvent:withMetadata:completionHandler` -> `logConversionWithType:metadata:completionHandler`
- Added a `logConversion` signature with `revenue` as a parameter
- Removed `sendEvent:withMetadata:location:completionHandler`, see [here](https://radarlabs.slack.com/archives/C02MF6PAWBH/p1677790478849349)
- Changed the signature for `RadarSendEventCompletionHandler` -> `RadarLogConversionCompletionHandler`
    - `location` and `user` have been removed as parameters
    - `events` -> `event` single conversion event is passed instead of an array of events from calling `trackOnce` + the conversion event
- Deduplicate `/track` calls
    - Added methods to update + get most recent time of `trackOnce` completions
    - added logic in `logConversion` to avoid calling `/track` again if it had been called within the past 60 seconds
- Permissions check before calling `/track`
    - Check for permissions and skip calling `/track` if permissions aren't granted


RadarSDK/Include/Radar.h
```diff
@@ -477,32 +477,47 @@ typedef void(^_Nonnull RadarSendEventCompletionHandler)(RadarStatus status, CLLo
...
+ @param metadata The metadata associated with the conversion.
  @param completionHandler A completion handler.
 
  @see https://radar.com/documentation/api#send-a-custom-event
  */
-+ (void)sendEvent:(NSString *)customType
-     withMetadata:(NSDictionary *_Nullable)metadata
-completionHandler:(RadarSendEventCompletionHandler)completionHandler NS_SWIFT_NAME(sendEvent(customType:metadata:completionHandler:));
++ (void)logConversionWithName:(NSString *)name
+     metadata:(NSDictionary *_Nullable)metadata
```
jsani-radar, line 489: nit: let's try to line up the colons between params like we do for other fns

RadarSDK/Include/Radar.h
```diff
@@ -477,32 +477,47 @@ typedef void(^_Nonnull RadarSendEventCompletionHandler)(RadarStatus status, CLLo
...
+ @param revenue The revenue generated by the conversion.
+ @param metadata The metadata associated with the conversion.
+ @param completionHandler A completion handler.
+ 
+ @see https://radar.com/documentation/api#send-a-custom-event
+ */
++ (void)logConversionWithName:(NSString *)name
+                      revenue:(NSNumber *)revenue
+                     metadata:(NSDictionary *_Nullable)metadata
+            completionHandler:(RadarLogConversionCompletionHandler)completionHandler NS_SWIFT_NAME(logConversion(name:metadata:completionHandler:));
```
jsani-radar, line 505: this has the same `NS_SWIFT_NAME` as the fn above, need to add `revenue`

RadarSDK/Include/Radar.h
```diff
@@ -477,32 +477,47 @@ typedef void(^_Nonnull RadarSendEventCompletionHandler)(RadarStatus status, CLLo
...
+ @param location The location of the conversion.
+ @param metadata The metadata associated with the conversion.
  @param completionHandler A completion handler.
 
  @see https://radar.com/documentation/api#send-a-custom-event
  */
-+ (void)sendEvent:(NSString *)customType
-     withLocation:(CLLocation *_Nullable)location
++ (void)logConversionWithName:(NSString *)name
+     location:(CLLocation *_Nullable)location
```
jsani-radar, line 518: ditto

jsani-radar (commented):
nitpicky comments, lgtm overall

we'll also need to update the call in the example app

also not sure why the build is failing

Example/Example/AppDelegate.swift
```diff
@@ -134,30 +134,30 @@ class AppDelegate: UIResponder, UIApplicationDelegate, UNUserNotificationCenterD
             print("Matrix: status = \(Radar.stringForStatus(status)); matrix[0][0].duration.text = \(String(describing: matrix?.routeBetween(originIndex: 0, destinationIndex: 0)?.duration.text)); matrix[0][1].duration.text = \(String(describing: matrix?.routeBetween(originIndex: 0, destinationIndex: 1)?.duration.text)); matrix[1][0].duration.text = \(String(describing: matrix?.routeBetween(originIndex: 1, destinationIndex: 0)?.duration.text)); matrix[1][1].duration.text = \(String(describing: matrix?.routeBetween(originIndex: 1, destinationIndex: 1)?.duration.text))")
         }
 
-        Radar.sendEvent(customType: "custom_event", metadata: ["data": "test"]) { (status, location, events, user) in
+        Radar.logConversion(name: "custom_event", metadata: ["data": "test"]) { (status, location, events, user) in
```
corypisano, line 137: just want to confirm cc @tjulien - can we remove the old signature completely? May have to deprecate the `sendEvent` method if there's callers in the wild so we don't break, but if not can remove

tjulien: Yeah, we’re going to remove sendEvent. We’ll put it in the Migration guide

RadarSDK/Include/Radar.h
```diff
@@ -477,32 +477,47 @@ typedef void(^_Nonnull RadarSendEventCompletionHandler)(RadarStatus status, CLLo
...
+ @param metadata The metadata associated with the conversion.
  @param completionHandler A completion handler.
 
  @see https://radar.com/documentation/api#send-a-custom-event
  */
-+ (void)sendEvent:(NSString *)customType
-     withMetadata:(NSDictionary *_Nullable)metadata
-completionHandler:(RadarSendEventCompletionHandler)completionHandler NS_SWIFT_NAME(sendEvent(customType:metadata:completionHandler:));
++ (void)logConversionWithName:(NSString *)name
+                     metadata:(NSDictionary *_Nullable)metadata
```
david-goodfellow, line 489: just double checking we want to move `withMetadata` to `metadata`

RadarSDK/RadarEvent+Internal.h
```diff
@@ -22,7 +22,7 @@
                      actualCreatedAt:(NSDate *_Nonnull)actualCreatedAt
                                 live:(BOOL)live
                                 type:(RadarEventType)type
-                          customType:(NSString *_Nullable)customType
+                      conversionType:(NSString *_Nullable)conversionType
```
tjulien, line 25: @david-goodfellow Do you foresee making this same change across the board in frontend, server, infra (kinesis, glue, athena, rollups, backfill, etc)?  Or leaving all of that as-is?

In my head I had sort of thought we would leave custom events as-in except for the syntactic sugar in the SDK, where we'd rename everything from "custom event" to "conversion"

jsani-radar: also, it looks like we wanted to change the `customType` param to `name` https://linear.app/radarlabs/issue/CORE-227/change-signature-of-custom-events-function-revenue-ios

should we keep this consistent between the external callsite and internal header?

RadarSDK/Radar.m
```diff
@@ -342,75 +350,67 @@ + (void)rejectEventId:(NSString *)eventId {
...
+        }
     }];
 }
 
-+ (void)sendEvent:(NSString *)customType
-     withLocation:(CLLocation *_Nullable)location
-         metadata:(NSDictionary *_Nullable)metadata
-completionHandler:(RadarSendEventCompletionHandler)completionHandler {
-    [self trackOnceWithLocation:location completionHandler:^(RadarStatus status, CLLocation * _Nullable location, NSArray<RadarEvent *> * _Nullable events, RadarUser * _Nullable user) {
++ (void)logConversionWithType:(NSString *)type
```
tjulien, line 375: Signatures look good - one small thing, we wanted to name the first param "name":

https://linear.app/radarlabs/issue/CORE-227#comment-1263621b

cc @david-goodfellow @jsani-radar

RadarSDK/RadarEvent+Internal.h
tjulien: cc @joeychik-radar @jsani-radar

david-goodfellow: the infra is going to stay the same but the logical lines in the sand I was making for namings was around repositories. E.g. in the sdk repos we should be consistent in using conversion everywhere

david-goodfellow: per the `name` conversation, I thought this was just the name given in the ticket when Tim quickly wrote this up while I was OOO. @joeychik-radar and I had some discussions around this as shown in the image. We can go either way. 
image

david-goodfellow: per the naming for `name`/`customType`, I thought this was just the name given in the ticket when Tim quickly wrote this up while I was OOO. @joeychik-radar and I had some discussions around this as shown in the image. We can go either way.

Joey asking in our personal threads:
```
from the logConversion signature fun logConversion(name, metadata, callback), should we change name to type? since we pass the RadarEvent to the client, which doesn't have a name field
```

jsani-radar: `name` is more approachable than `type`, which seems like an internal impl detail

joeychik-radar: should we rename the `customType` property on `RadarEvent` to `name` as well? or should we go for something more specific, as having a property called `name` seems a bit weird outside of the context of conversions

jsani-radar: my fault, for the internal `RadarEvent` model, having both `name` and `type` is inconsistent with the naming on the server and could be confusing

joeychik-radar: `customType` here is syntactic sugar; events on the server don't have this field. The server passes down `type` as a string, and that gets stored as an enum in `RadarEvent`. The value of `customType` is whatever gets passed down from the server as `type` when `type` doesn't fit any of the enum values

jsani-radar: sgtm - since this is an internal impl detail, I'll defer to you all on the most intuitive / easiest to understand approach for future devs

Example/Example/AppDelegate.swift
```diff
@@ -134,31 +134,12 @@ class AppDelegate: UIResponder, UIApplicationDelegate, UNUserNotificationCenterD
             print("Matrix: status = \(Radar.stringForStatus(status)); matrix[0][0].duration.text = \(String(describing: matrix?.routeBetween(originIndex: 0, destinationIndex: 0)?.duration.text)); matrix[0][1].duration.text = \(String(describing: matrix?.routeBetween(originIndex: 0, destinationIndex: 1)?.duration.text)); matrix[1][0].duration.text = \(String(describing: matrix?.routeBetween(originIndex: 1, destinationIndex: 0)?.duration.text)); matrix[1][1].duration.text = \(String(describing: matrix?.routeBetween(originIndex: 1, destinationIndex: 1)?.duration.text))")
         }
 
-        Radar.sendEvent(customType: "custom_event", metadata: ["data": "test"]) { (status, location, events, user) in
-            if let customEvent = events?.first,
-               customEvent.type == .custom {
-                print("Custom type: \(customEvent.customType!)")
+        Radar.logConversion(type: "conversion_event", metadata: ["data": "test"]) { (status, event) in
```
tjulien, line 137: This should be `name` instead of `type` I think?

RadarSDK/Include/Radar.h
```diff
@@ -477,32 +477,32 @@ typedef void(^_Nonnull RadarSendEventCompletionHandler)(RadarStatus status, CLLo
...
+            completionHandler:(RadarLogConversionCompletionHandler)completionHandler NS_SWIFT_NAME(logConversion(type:metadata:completionHandler:));
```
tjulien, line 490: replace `type` with `name`

MIGRATION.md
```diff
@@ -1,5 +1,28 @@
...
+
+Radar.logConversion(name: "conversion_event", metadata: metadata) { (status, event) in
+    let conversionName = event?.conversionName // should be "conversion_event"
+    let conversionType = event?.type // should be RadarEventType.conversion
+}
+
+Radar.logConversion(name: "conversion_with_revenue", revenue: 0.2, metadata: metadata) { (status, event) in
+    let revenue = event?.metadata?["revenue"] // should be 0.2
+}
+```
```
joeychik-radar, line 24: let me know if this looks weird, this is my first time touching swift

MIGRATION.md
```diff
@@ -1,5 +1,28 @@
 # Migration guides
 
+## 3.5.12 to 3.5.13
```
tjulien, line 3: This is probably going to be the `3.6.1` release

MIGRATION.md
```diff
@@ -1,5 +1,28 @@
...
+- Custom events have been renamed to conversions.
+      - `Radar.sendEvent(customType:metadata:completionHandler:)` is now `Radar.logConversion(name:metadata:completionHandler)`.
+      - `Radar.logConversion(name:revenue:metadata:callback:)` has been added.
+      - `Radar.sendEvent(customType:metadata:location:callback:)` has been removed.
+      - `RadarSendEventCompletionHandler(status, location, events, user)` is now `RadarLogConversionCompletionHandler(status, event)`.
+            - `location` and `user` are no longer available, and only the conversion event is returned as `event` instead of a coalesced list of events.
+      - On `RadarEvent`, `customType` is now `conversionName`, and `RadarEventType.custom` is now `RadarEventType.conversion`.
+
+```swift
+// 3.5.13
```
tjulien, line 13: If you scroll back through some of the other entries in MIGRATION.MD you can see the pattern we want to use for the code snippets.  Basically an example with the new version of the SDK and then underneath that an example on the old version.

RadarSDK/Radar.m
```diff
@@ -383,80 +391,55 @@ + (void)rejectEventId:(NSString *)eventId {
...
-                                     }
-                                 }];
+    CLAuthorizationStatus authorizationStatus = [[RadarLocationManager sharedInstance].permissionsHelper locationAuthorizationStatus];
+    if (authorizationStatus == kCLAuthorizationStatusDenied || isLastTrackRecent) {
+        [self sendLogConversionRequestWithName:name metadata:metadata completionHandler:completionHandler];
+        
+        return;
+    }
+    
+    [self trackOnceWithCompletionHandler:^(RadarStatus status, CLLocation * _Nullable location, NSArray<RadarEvent *> * _Nullable events, RadarUser * _Nullable user) {
```
tjulien, line 429: @joeychik-radar @david-goodfellow @jsani-radar 

Why are we doing a `trackOnceWithCompletionHandler` here if we don't pass in any of the response data from it to `sendLogConversionRequestWithName`?  Related, why are we checking for a recent track request?

jsani-radar: the standard customer flow is `trackOnce` --> surface in-app use case --> log conversion event - if logging the conversion event were to call `trackOnce` again, that'd be redundant

so, there's a check if a recent enough location fix has been grabbed - if not, then we should flush the user state with a new one to attribute the right location context to the conversion

tjulien: "to attribute the right location context to the conversion" - Are we attaching a lat long to the Event here or does this all happen on the server side?

david-goodfellow: all the location context (geofence, beacon, trip, etc) is attached server side

RadarSDK/Radar.m
```diff
@@ -150,6 +150,10 @@ + (void)trackOnceWithDesiredAccuracy:(RadarTrackingOptionsDesiredAccuracy)desire
                                            beacons:beacons
                                  completionHandler:^(RadarStatus status, NSDictionary *_Nullable res, NSArray<RadarEvent *> *_Nullable events, RadarUser *_Nullable user,
                                                      NSArray<RadarGeofence *> *_Nullable nearbyGeofences, RadarConfig *_Nullable config) {
+                                     if (status == RadarStatusSuccess) {
```
tjulien, line 153: You can probably add this directly in `RadarAPIClient.trackWithLocation`

RadarSDK/Radar.m
```diff
@@ -218,6 +218,10 @@ + (void)trackOnceWithLocation:(CLLocation *)location completionHandler:(RadarTra
                                                beacons:nil
                                      completionHandler:^(RadarStatus status, NSDictionary *_Nullable res, NSArray<RadarEvent *> *_Nullable events, RadarUser *_Nullable user,
                                                          NSArray<RadarGeofence *> *_Nullable nearbyGeofences, RadarConfig *_Nullable config) {
+                                         if (status == RadarStatusSuccess) {
+                                             [RadarSettings updateLastTrackedTime];
```
tjulien, line 222: Can remove this one now

tjulien (approved):

PR #228, opened 2023-02-21
Log request latency and replay buffer size

lmeier:
Note that we'll need some dashboard log work to make this actually visible on the dashboard.


RadarSDK/RadarAPIHelper.m
```diff
@@ -129,9 +134,17 @@ - (void)requestWithMethod:(NSString *)method
...
                     res = (NSDictionary *)resObj;
 
-                    [[RadarLogger sharedInstance]
-                        logWithLevel:RadarLogLevelDebug
-                             message:[NSString stringWithFormat:@"📍 Radar API response | method = %@; url = %@; statusCode = %ld; res = %@", method, url, (long)statusCode, res]];
+                    if (params && [params objectForKey:@"replays"]) {
+                        NSArray *replays = [params objectForKey:@"replays"];
+                        [[RadarLogger sharedInstance]
+                            logWithLevel:RadarLogLevelDebug
+                                 message:[NSString stringWithFormat:@"📍 Radar API response | method = %@; url = %@; statusCode = %ld; res = %@; latency = %f; replays = %lu", method, url, (long)statusCode, res, latency, (unsigned long)replays.count]];
```
jsani-radar, line 141: add units for `latency`?

RadarSDK/RadarAPIHelper.m
```diff
@@ -78,7 +78,12 @@ - (void)requestWithMethod:(NSString *)method
                 configuration.timeoutIntervalForResource = 10;
             }
 
+            NSDate *requestStart = [NSDate date];
+
             void (^dataTaskCompletionHandler)(NSData *data, NSURLResponse *response, NSError *error) = ^(NSData *data, NSURLResponse *response, NSError *error) {
+                NSTimeInterval latency = [requestStart timeIntervalSinceNow] * -1;
```
jsani-radar, line 84: didn't the multiplication for a bit, maybe add a comment that timeIntervalSinceNow returns a negative number?

RadarSDK/RadarAPIHelper.m
```diff
@@ -78,7 +78,12 @@ - (void)requestWithMethod:(NSString *)method
...
+
+
```
jsani-radar, line 86: nit: extra `\n`

jsani-radar (commented):
lgtm, small comments

adding @corypisano and @palmakkar for 👀

RadarSDK/RadarAPIHelper.m
lmeier: Hmmm, the other values here don't have units, and it sort of breaks the convention going on here. I noted in the comment you suggested that the value is in seconds. Changing this string to something like `latency (s)=` seems non-ideal, but open to switching if you feel strongly.

nickpatrick (approved):

PR #229, opened 2023-02-22
Fix client side rate limiting

lmeier:
There was a bug in the previous implementation. We were incrementing the semaphore `dispatch_semaphore_signal(self.semaphore);` every time we called `requestWithMethod` though we were decrementing the semaphore only when we ran into the iVar `self.wait`. This meant that the semaphore grew rapidly as more often than not, we'd not run into `self.wait == true`, increasing the amount of concurrent `trackLocation`s we'd allow.

This PR changes the logic to check the semaphore (now initialized `dispatch_semaphore_create(1)`) only if `SLEEP: true`, which is only true for `trackWithLocation` where we have the perSecondRateLimit.

### DIFF
--- a/RadarSDK/RadarAPIHelper.m
+++ b/RadarSDK/RadarAPIHelper.m
@@ -24,7 +24,7 @@
     self = [super init];
     if (self) {
         _queue = dispatch_queue_create("io.radar.api", DISPATCH_QUEUE_SERIAL);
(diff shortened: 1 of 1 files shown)


nickpatrick:
Great catch

nickpatrick (approved):

PR #233, opened 2023-02-28
CORE-291 log app opens by default

joeychik-radar:


RadarSDK/Radar.m
```diff
@@ -1051,6 +1051,10 @@ - (void)applicationWillEnterForeground {
             [[RadarLocationManager sharedInstance] updateTrackingFromMeta:meta];
         }];
     }
+    
+    [[RadarAPIClient sharedInstance] sendEvent:@"app_open" withMetadata:nil user:nil trackEvents:nil completionHandler:^(RadarStatus status, NSDictionary * _Nullable res, NSArray<RadarEvent *> * _Nullable events) {
```
david-goodfellow, line 1055: is it required to pass `withMetadata`, `user` and `trackEvents` with the value `nil` or can we remove these?

joeychik-radar: i think it's required, objective-c doesnt have default parameters. we could define new function signatures to emulate default params, but that would be a less clean solution

david-goodfellow:
let's merge this one. While we wait for open questions on the android, let's push forward with this https://linear.app/radarlabs/issue/CORE-306/ios-leverage-the-users-last-location-check-tracking-permissions

PR #238, opened 2023-03-08
App Attest

nickpatrick:
Changes needed:
- Only call `attestKey` once
- Call `generateAssertion` and pass that assertion to `POST /track` instead
- See https://developer.apple.com/documentation/devicecheck/preparing_to_use_the_app_attest_service?language=objc


tderouin:
@nick - can you pass `keyId` to the server too? I need to use it on the server side. 

I can do this for now: 

```
completionHandler([NSString stringWithFormat:@"%@::::%@", assertionString, keyId], nil);
```

To explain, the [documentation](https://developer.apple.com/documentation/devicecheck/validating_apps_that_connect_to_your_server) says:

_This produces an attestation object that your app passes to your server, along with the corresponding key identifier._

Also, they recommend storing this key Id so it can be reused later, instead of generating one each time:

_Your server deems the app instance to be valid if it can successfully verify the attestation object. In this case, be sure to persistently store the key identifier — not the attestation object — in your app for future use in signing server requests._

tderouin:
The server PR for this is being reviewed, a few things:

* We need to send the `keyId` to `trackVerified` along with `attestationString`
* I noticed when testing on my iPhone, I would successfully get an attestation object back from Apple only about 50% of the time. It seemed pretty consistent that it would succeed on every second call. Something to think about as we test.

lmeier:
@tderouin I updated this PR to include the `keyId`. Not exactly sure how to test `trackVerified` against staging environments, so once you're back from OOO, let's figure out how to validate that this PR is where you want it to be.

tderouin:
[How to test App Attest on iOS ](https://www.notion.so/radarlabs/How-to-test-App-Attest-on-iOS-ccb59fe1264d4241a9af9d7609cf7209)

I think we are good to go. I've got 30 successful validations with an iPhone 11 (iOS 16.5.1) and 10 validations with iPhone 8 (iOS 15.6) when running against this [PR](https://github.com/radarlabs/server/pull/4974).

tjulien (approved):

PR #243, opened 2023-03-13
CORE-373 add opened_app in initialize()

joeychik-radar:


lmeier (approved):
lgtm

RadarSDK/Radar.m
```diff
@@ -57,6 +57,10 @@ + (void)initializeWithPublishableKey:(NSString *)publishableKey {
                                      completionHandler:^(RadarStatus status, RadarConfig *config) {
                                          [[RadarLocationManager sharedInstance] updateTrackingFromMeta:config.meta];
                                      }];
+    [self logConversionWithName:@"opened_app" metadata:nil completionHandler:^(RadarStatus status, RadarEvent * _Nullable event) {
+        NSString *message = [NSString stringWithFormat:@"Conversion name = %@: status = %@; event = %@", event.conversionName, [Radar stringForStatus:status], event];
+        [[RadarLogger sharedInstance] logWithLevel:RadarLogLevelInfo message:message];
+    }];
 }
```
tjulien, line 64: @david-goodfellow @joeychik-radar @lmeier @nickpatrick @jsani-radar 

1) Are we going to be logging app opens twice now? (once in `applicationWillEnterForeground` and once here)?
2) Will this new one in `initializeWithPublishableKey` always correspond to an app open? Seems like it could also happen in the background or other times depending on when people call `initializeWithPublishableKey`.
3. Is the issue with the `applicationWillEnterForeground` call site that `initializeWithPublishableKey` may not have completed yet, so an api call may not succeed (no publishable key)?

lmeier: AFAICT, I don't really see the `applicationWillEnterForeground` events in my SDK logs, and so my suspicion is that #3 is correct. 

The best way of doing `app_opened` events might be to cache these events upon `applicationWillEnterForeground` or perhaps `applicationDidBecomeActive` (list of lifecycle events at the bottom of: https://developer.apple.com/documentation/uikit/uiapplicationdelegate/1622956-applicationdidbecomeactive?language=objc ) and then periodically clear the cache (every initialize and some repeating amount thereafter?)

IMO we should dogfood a bit with both and pay close attention to the number of events coming across (perhaps we should slightly differentiate b/t each call of `app_opened` with a suffix?), which should hopefully tell us what the most reliable method is here (ideally one that doesn't require more than a simple API call)

lmeier: One way of preventing the worry articulated in 2 is to check `[RadarUtils foreground]` before that `initializeWithPublishableKey` `app_opened` call.

jsani-radar: as discussed in [Slack](https://radarlabs.slack.com/archives/C02MF6PAWBH/p1678830720484899), let's keep parity with how we measure sessions today to have a 1:1 with the existing data so there's no surprises for CS / customers

PR #251, opened 2023-03-17
CORE-373 Improve logic for opened_app conversions

joeychik-radar:


RadarSDK/Radar.m
```diff
@@ -406,6 +407,18 @@ + (void)sendLogConversionRequestWithName:(NSString * _Nonnull) name
     }];
 }
 
++ (void)logOpenedAppConversion {
+    // if opened_app has been logged in the last 1000 milliseconds, don't log it again
+    NSTimeInterval lastAppOpenTimeInterval =[[NSDate date] timeIntervalSinceDate:[RadarSettings lastAppOpenTime]];
+    if (lastAppOpenTimeInterval * 1000 > 1000) {
```
david-goodfellow, line 413: if wea re multiplying the left by 1k why don't we just do `if (lastAppOpenTimeInterval > 1)`

david-goodfellow:
need to version bump

lmeier:
I'm happy we're rolling with the 1 second window to present overlapping / duplicate events.

I think that in addition to being called ourselves by the initialize, hence putting ourselves in the `didFinishLaunchingWithOptions` path, we'll also want to instruct customers to call this in the `applicationDidBecomeActive` path, to capture app opens where it's coming back from being backgrounded (not a cold start). Docs here: https://developer.apple.com/documentation/uikit/uiapplicationdelegate/1622956-applicationdidbecomeactive?language=objc

If that's right, the only non-documentation change needed is moving the function header from `Radar+Internal.h` to `Radar.h`.

@nickpatrick does that seem right, in terms of handling the app lifecycle the right way?

RadarSDK/Radar+Internal.h
```diff
@@ -14,4 +14,6 @@
 
 + (void)flushLogs;
 
++ (void)logOpenedAppConversion;
```
lmeier, line 17: See top level note. I suspect we should move this to `Radar.h`

joeychik-radar:
> we'll also want to instruct customers to call this in the `applicationDidBecomeActive` path, to capture app opens where it's coming back from being backgrounded (not a cold start)

Should we just use the same [pattern](https://developer.apple.com/documentation/uikit/uiapplicationdidbecomeactivenotification) that we already use for `applicationWillEnterForeground`? And if we do this it seems like we might not want to send `opened_app` from `applicationWillEnterForeground` at all

RadarSDK/Radar.m
```diff
@@ -406,6 +407,18 @@ + (void)sendLogConversionRequestWithName:(NSString * _Nonnull) name
...
+    NSTimeInterval lastAppOpenTimeInterval = [[NSDate date] timeIntervalSinceDate:[RadarSettings lastAppOpenTime]];
+    if (lastAppOpenTimeInterval > 1) {
+        [RadarSettings updateLastAppOpenTime];
+        [self sendLogConversionRequestWithName:@"opened_app" metadata:nil completionHandler:^(RadarStatus status, RadarEvent * _Nullable event) {
+            NSString *message = [NSString stringWithFormat:@"Conversion name = %@: status = %@; event = %@", event.conversionName, [Radar stringForStatus:status], event];
+            [[RadarLogger sharedInstance] logWithLevel:RadarLogLevelInfo message:message];
+        }];
+    }
+}
+
```
lmeier, line 421: As noted during lunch, we'll need a separate `logOpenedAppConversionWithMetadata` that customers can manually insert into their `UNUserNotificationCenterDelegate`. This function should:
- have priority over the other `logOpenedAppConversion` calls
- maintain the deduplication logic


Currently I have:
```+ (void)logOpenedAppConversionWithMetadata:(NSDictionary *_Nullable)metadata {
    // log that we're in this
    [[RadarLogger sharedInstance] logWithLevel:RadarLogLevelInfo message:@"logOpenedAppConversionWithMetadata"];
    // if opened_app has been logged in the last 1000 milliseconds, don't log it again
    NSTimeInterval lastAppOpenTimeInterval =[[NSDate date] timeIntervalSinceDate:[RadarSettings lastAppOpenTime]];
    if (lastAppOpenTimeInterval * 1000 > 1000) {
        [RadarSettings updateLastAppOpenTime];
        [self sendLogConversionRequestWithName:@"opened_app" metadata:metadata completionHandler:^(RadarStatus status, RadarEvent * _Nullable event) {
            NSString *message = [NSString stringWithFormat:@"Conversion name = %@: status = %@; event = %@", event.conversionName, [Radar stringForStatus:status], event];
            [[RadarLogger sharedInstance] logWithLevel:RadarLogLevelInfo message:message];
        }];
    }
}

+ (void)logOpenedNotificationConversionWithMetadata:(NSDictionary *_Nullable)metadata {
    [self sendLogConversionRequestWithName:@"opened_notification" metadata:metadata completionHandler:^(RadarStatus status, RadarEvent * _Nullable event) {
        NSString *message = [NSString stringWithFormat:@"Conversion name = %@: status = %@; event = %@", event.conversionName, [Radar stringForStatus:status], event];
        [[RadarLogger sharedInstance] logWithLevel:RadarLogLevelInfo message:message];
    }];
}
```
which seems to be prioritizing the other `app_opened` over the metadata one

lmeier: I take this back — I no longer think we should be unifying `notification_opened` and `app_opened`. Sending a note in channel now

david-goodfellow (approved):

PR #256, opened 2023-04-13
Pass location, events, user to trackVerified completion handler

lmeier:


nickpatrick (approved):
Good catch. I'll take this out later today and may also release this for 3.6.x as a hotfix

PR #259, opened 2023-04-26
Add polygon geometry to geofence dictionary

lmeier:


RadarSDK/RadarCoordinate.m
```diff
@@ -78,4 +78,12 @@ - (NSDictionary *)dictionaryValue {
     return @{@"type": @"Point", @"coordinates": @[@(self.coordinate.longitude), @(self.coordinate.latitude)]};
 }
 
++ (NSArray<NSArray *>*)arrayForCoordinates:(NSArray<RadarCoordinate *> *)coordinates {
```
nickpatrick, line 81: We should reconstruct `geometry` exactly, so this may need to be a `GeoJSON` struct instead of just an array of coordinates

Nit but I would also probably move this to `RadarGeofence.m` since I think this helper applies more to geofences than to individual coordinates

RadarSDK/RadarGeofence.m
```diff
@@ -124,7 +124,7 @@ - (instancetype _Nullable)initWithObject:(id)object {
 
         if ([type isEqualToString:@"circle"]) {
             geometry = [[RadarCircleGeometry alloc] initWithCenter:center radius:radius];
-        } else if ([type isEqualToString:@"polygon"] || [type isEqualToString:@"isochrone"]) {
+        } else if ([type isEqualToString:@"polygon"] || [type isEqualToString:@"Polygon"] || [type isEqualToString:@"isochrone"]) {
```
nickpatrick, line 127: Are we seeing `"Polygon"` coming back from the API?

nickpatrick: Oh, I'm guessing we were thinking `GeoJSON` `type = "Polygon"`, which would be `geofence.geometry.type`, not `geofence.type`

lmeier: that's correct

RadarSDK/RadarCoordinate.m
lmeier: Makes sense, will do that, and move it over

PR #262, opened 2023-06-13
Allow user.debug for live projects

lmeier:


RadarSDK/Radar.m
```diff
@@ -1162,7 +1162,7 @@ - (void)dealloc {
 
 + (BOOL)isTestKey {
     NSString *publishableKey = [RadarSettings publishableKey];
-    if ([publishableKey hasPrefix:@"prj_test_pk"] || [publishableKey hasPrefix:@"org_test_pk"]) {
+    if ([publishableKey hasPrefix:@"prj_test_pk"] || [publishableKey hasPrefix:@"org_test_pk"] || [RadarSettings userDebug]) {
```
tjulien, line 1165: This is good for consistency with Android, maybe in a future version we pull this out of the `isTestKey` function

RadarSDK/RadarSettings.m
```diff
@@ -239,6 +240,15 @@ + (NSString *)verifiedHost {
     return verifiedHost ? verifiedHost : kDefaultVerifiedHost;
 }
 
++ (BOOL)userDebug {
```
tjulien, line 243: We also need the iOS analog of [this change on Android](https://github.com/radarlabs/radar-sdk-android/commit/7f7ffbe800ab08662ea39a135fc88408a362453a#diff-e17c923825657ab70c94cca4e224e5a6c65d35a8fc2bfecb7e42dc9527a89d8bR256-R257) in the `logLevel` function

RadarSDK/RadarUser.m
```diff
@@ -95,6 +97,7 @@ - (instancetype _Nullable)initWithObject:(NSObject *)object {
     RadarLocationSource source = RadarLocationSourceUnknown;
     RadarTrip *trip;
     RadarFraud *fraud;
+    BOOL debug = YES;
```
tjulien, line 100: [Android seems to default to false here](https://github.com/radarlabs/radar-sdk-android/commit/7f7ffbe800ab08662ea39a135fc88408a362453a#diff-97ba6ea35c8dc52c65848399a1480a111e3fab8a0226ea4d7e26b0fbbd4560f0R128) (though confusingly the setting defaults to `true` on Android, which I don't love).

lmeier: By "the setting defaults to true on Android", do you mean this functions default: https://github.com/radarlabs/radar-sdk-android/blob/b94182de6c830c985f69c85121bbc9ca48a67b47/sdk/src/main/java/io/radar/sdk/RadarSettings.kt#L300 ?

Think some logic like that is why I had flipped it here, but I just changed it back.

RadarSDK/Radar.m
lmeier: SG.

tjulien (approved):

PR #264, opened 2023-06-28
Handle scheduledArrivalAt timestamp in MS on iOS

lmeier:
[We do this on Android](https://github.com/radarlabs/radar-sdk-android/blob/master/sdk/src/main/java/io/radar/sdk/RadarTripOptions.kt#L70), so replicating here in order to make this work for React Native.


lmeier:
With this change, we should remove the custom `scheduledArrivalAt` handling in `RNradar.m` in the react native. Have the other PR set to do that: https://github.com/radarlabs/react-native-radar/pull/231

RadarSDK/RadarTripOptions.m
```diff
@@ -55,6 +55,8 @@ + (RadarTripOptions *)tripOptionsFromDictionary:(NSDictionary *)dict {
         scheduledArrivalAt = [RadarUtils.isoDateFormatter dateFromString:(NSString *)scheduledArrivalAtObj];
```
jsani-radar, line 55: nit: should these 3 conditions be nested?

might be easier to read / reason about if you're checking the presence of the object first, then checking the type within

it'd also short-circuit slightly quicker when `scheduledArrivalAt` isn't present

lmeier: https://github.com/radarlabs/radar-sdk-ios/pull/264/commits/1a5ae8a297bf0ab0c9c11dc5ec1fca6b770536ed

tjulien (approved):

PR #270, opened 2023-08-16
add trackVerifiedToken()

nickpatrick:


Example/Example/AppDelegate.swift
```diff
@@ -22,9 +22,14 @@ class AppDelegate: UIResponder, UIApplicationDelegate, UNUserNotificationCenterD
         self.requestLocationPermissions()
 
         // Replace with a valid test publishable key
-        Radar.initialize(publishableKey: "prj_test_pk_0000000000000000000000000000000000000000")
+        Radar.initialize(publishableKey: "org_test_pk_5857c63d9c1565175db8b00750808a66a002acb8")
```
nickpatrick, line 25: Will revert this

RadarSDK/RadarUtils.m
```diff
@@ -156,6 +158,57 @@ + (NSString *)dictionaryToJson:(NSDictionary *)dict {
     } 
 }
 
++ (BOOL)isJailbroken {
```
nickpatrick, line 161: May revert this

RadarSDK/RadarAPIClient.m
```diff
@@ -327,10 +331,21 @@ - (void)trackWithLocation:(CLLocation *_Nonnull)location
...
 
+                        if (encrypted) {
+                            if (!tokenObj) {
+                                return completionHandler(status, nil, nil, nil, nil, nil, nil);
+                            }
+
+                            NSString *token = (NSString *)tokenObj;
+
+                            return completionHandler(status, nil, nil, nil, nil, nil, token);
+                        }
```
lmeier, line 347: This codepath is completely forgoing updating the event & location update handlers, trips, & place updating.

As such, `trackVerified` and `trackVerifiedEncrypted` behave significantly differently -- wonder how we should make it clear to users the difference in behavior and effects (could see myself getting confused by this).

RadarSDK/Include/Radar.h
```diff
@@ -185,6 +185,15 @@ typedef void (^_Nullable RadarBeaconCompletionHandler)(RadarStatus status, NSArr
  */
 typedef void (^_Nullable RadarTrackCompletionHandler)(RadarStatus status, CLLocation *_Nullable location, NSArray<RadarEvent *> *_Nullable events, RadarUser *_Nullable user);
 
+/**
+ Called when an encrypted track request succeeds, fails, or times out.
+
+ Receives the request status and, if successful, a JSON Web Token (JWT) containing an array of the events generated and the user. Decrypt the JWT server-side using your private key.
+
+ @see https://radar.com/documentation/sdk/ios
```
jsani-radar, line 193: should we be pointing to the iOS docs or fraud docs? comments above the callback and method are inconsistent at the moment

jsani-radar (commented):
looking good overall - few small things

RadarSDK/RadarUtils.m
```diff
@@ -156,6 +158,57 @@ + (NSString *)dictionaryToJson:(NSDictionary *)dict {
...
+    
+    NSError *error;
+    [data writeToFile:testPath options:NSDataWritingAtomic error:&error];
+    
+    // If no error occurred, the device is likely jailbroken
+    if (!error) {
+        NSLog(@"failed check 2");
+        [[NSFileManager defaultManager] removeItemAtPath:testPath error:nil];
+        return YES;
+    }
```
lmeier, line 189: The first test makes sense to me.

For this test here, I'd love to read up on why this test works — is there a SO or other resource you can point me to?

Example/Example/AppDelegate.swift
```diff
@@ -22,9 +22,14 @@ class AppDelegate: UIResponder, UIApplicationDelegate, UNUserNotificationCenterD
...
         Radar.setDelegate(self)
+        
+        Radar.trackVerified { status, location, events, user in
+
```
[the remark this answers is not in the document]
tderouin: I ran into a weird issue with this:

```
        
        Radar.trackVerifiedEncrypted{ ( status, token ) in
            NSLog("Got token:")
            NSLog(token!)
        }
```

The `token` printed was truncated to 1,024 characters, but the token returned from the server was 7,000+ characters. 

While running the debugger, and clicking `Copy`, would sometimes copy the truncated value, but not always:

image

I had to run `print (void)CFShow(token)` in the debugging console to show the entire token to confirm it was being captured.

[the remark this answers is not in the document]
nickpatrick: I ran into a similar issue testing on Android where `logcat` lines seem to be truncated to `4000` chars, but my token was something like `4300`. I just printed it in two parts. In any case, I reverted this. We can revisit inclusion in sample app when we revisit sample app code more broadly

RadarSDK/RadarAPIClient.m
nickpatrick: Will mention on /documentation. Most people using this aren't going to be using `startTracking()`, so shouldn't cause unexpected behavior in practice

RadarSDK/Include/Radar.h
nickpatrick: Good catch, will update

RadarSDK/Radar.m
```diff
@@ -219,7 +219,7 @@ + (void)trackOnceWithLocation:(CLLocation *)location completionHandler:(RadarTra
                                               replayed:NO
                                                beacons:nil
                                      completionHandler:^(RadarStatus status, NSDictionary *_Nullable res, NSArray<RadarEvent *> *_Nullable events, RadarUser *_Nullable user,
-                                                         NSArray<RadarGeofence *> *_Nullable nearbyGeofences, RadarConfig *_Nullable config) {
+                                                         NSArray<RadarGeofence *> *_Nullable nearbyGeofences, RadarConfig *_Nullable config, NSString *_Nullable token) {
```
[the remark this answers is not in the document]
nickpatrick: This callback is only ever used internally, not worried about

RadarSDK/RadarUtils.m
nickpatrick: This doesn't actually work, reverting

RadarSDK/RadarUtils.m
nickpatrick: Reverted

Example/Example/AppDelegate.swift
nickpatrick: Reverted

RadarSDK/RadarAPIClient.h
```diff
@@ -86,6 +87,7 @@ typedef void (^_Nonnull RadarSyncLogsAPICompletionHandler)(RadarStatus status);
         attestationString:(NSString *_Nullable)attestationString
                     keyId:(NSString *_Nullable)keyId
          attestationError:(NSString *_Nullable)attestationError
+                encrypted:(BOOL)encrypted
```
lmeier, line 90: was going to suggest we think about changing this to `verified` given discussion in slack & recent changes, but of course that's already used.. 🤔

lmeier: Looking at the JWT docs, I think `encoded` would perhaps be most accurate

nickpatrick: `token` param or `POST /track/token` endpoint are both options, too

PR #271, opened 2023-08-17
Persistent storage for replays (liammeier/fence-1196)

lmeier:
We utilize `NSUserDefaults` to persist replays every time we write to the replay buffer, and load the replay buffer into memory upon `initialize`.

Local testing is proceeding well. I'll implement in android, and cut a Waypoint with Kenny this afternoon for larger testing.


RadarSDK/Radar.m
```diff
@@ -53,6 +54,8 @@ + (void)initializeWithPublishableKey:(NSString *)publishableKey {
         [RadarSettings updateSessionId];
     }
 
+    [[RadarReplayBuffer sharedInstance] loadReplaysFromPersistentStore];
```
jsani-radar, line 57: don't have a strong opinion, but maybe we should load replays into memory after the call to `/config` is made and tracking options are updated

RadarSDK/RadarReplay.m
```diff
@@ -17,6 +17,18 @@ - (instancetype)initWithParams:(NSDictionary *)replayParams {
     return self;
 }
 
+- (instancetype)initWithCoder:(NSCoder *)coder {
```
jsani-radar, line 20: n00b question - should we declare that `RadarReplay` abides by the `NSCoding` interface anywhere, or is that implicit within `NSObject`?

RadarSDK/RadarReplayBuffer.m
```diff
@@ -42,6 +43,10 @@ - (void)writeNewReplayToBuffer:(NSMutableDictionary *)replayParams {
     // add new replay to buffer
     RadarReplay *radarReplay = [[RadarReplay alloc] initWithParams:replayParams];
     [mutableReplayBuffer addObject:radarReplay];
+
+    // persist buffer
```
jsani-radar, line 47: in an ideal world would this be encrypted?

RadarSDK/RadarReplayBuffer.m
```diff
@@ -42,6 +43,10 @@ - (void)writeNewReplayToBuffer:(NSMutableDictionary *)replayParams {
...
+    NSData *replaysData = [NSKeyedArchiver archivedDataWithRootObject:mutableReplayBuffer];
+    [[NSUserDefaults standardUserDefaults] setObject:replaysData forKey:@"radar-replays"];
```
jsani-radar, line 49: just want to be cognizant that we're rewriting the entire buffer to persistent storage each time here

RadarSDK/RadarReplayBuffer.m
```diff
@@ -57,6 +62,19 @@ - (void)writeNewReplayToBuffer:(NSMutableDictionary *)replayParams {
...
+}
+
+- (void)loadReplaysFromPersistentStore {
+    [[RadarLogger sharedInstance] logWithLevel:RadarLogLevelDebug message:@"Loading replays from persistent store"];
+    NSData *replaysData = [[NSUserDefaults standardUserDefaults] objectForKey:@"radar-replays"];
+    if (replaysData) {
+        NSArray *replays = [NSKeyedUnarchiver unarchiveObjectWithData:replaysData];
+        [[RadarLogger sharedInstance] logWithLevel:RadarLogLevelDebug message:[NSString stringWithFormat:@"Loaded replays with length %lu", (unsigned long)[replays count]]];
+        mutableReplayBuffer = [NSMutableArray arrayWithArray:replays];
+    }
```
jsani-radar, line 77: is it worth having a log line for "No replays found from persistent store" if `replaysData` is empty?

jsani-radar (commented):
looking solid overall - simpler PR than I thought

RadarSDK/Radar.m
lmeier: Yea -- perhaps we gate this on `replays: all` after the config.

jsani-radar: I like that

RadarSDK/RadarReplayBuffer.m
lmeier: can be inferred from lack of the "loaded replays with..."

RadarSDK/RadarReplayBuffer.m
tjulien: We could have a manifest of entries but that's probably a premature optimization at this point.

tjulien (approved):

RadarSDK/RadarReplayBuffer.m
```diff
@@ -42,6 +43,23 @@ - (void)writeNewReplayToBuffer:(NSMutableDictionary *)replayParams {
     // add new replay to buffer
     RadarReplay *radarReplay = [[RadarReplay alloc] initWithParams:replayParams];
     [mutableReplayBuffer addObject:radarReplay];
+
+    NSMutableArray<RadarReplay *> *prunedBuffer;
+
+    // If buffer length is above 50, remove every fifth replay from the persisted buffer 
+    if ([mutableReplayBuffer count] > 50) {
```
jsani-radar, line 50: hmm, wondering under what circumstances this would happen? when would `/track/replay` realistically return a non-200?

could be wrong but if I understand correctly, the endpoint returns a 201 and does everything async after the fanout - regardless, we should handle SQS being down gracefully

lmeier: In early replay testing, I witnessed some 413 Payload Too Large's

PR #274, opened 2023-08-31
Liammeier/fence 1220 flush replays in more sdk lifecycle methods ios

lmeier:


RadarSDK/RadarAPIClient.m
```diff
@@ -283,13 +318,22 @@ - (void)trackWithLocation:(CLLocation *_Nonnull)location
...
+            if (status != RadarStatusSuccess) {
+                [[RadarLogger sharedInstance] logWithLevel:RadarLogLevelDebug message:[NSString stringWithFormat:@"Failed to flush replays"]];
+                [[RadarDelegateHolder sharedInstance] didFailWithStatus:status];
+            } else {
+                [[RadarLogger sharedInstance] logWithLevel:RadarLogLevelDebug message:[NSString stringWithFormat:@"Succeded in flushing replays"]];
+                [RadarState setLastFailedStoppedLocation:nil];
+                [RadarSettings updateLastTrackedTime];
+            }
+
+            [[RadarDelegateHolder sharedInstance] didFailWithStatus:status];
```
corypisano, line 332: Should just need the one `didFailWithStatus` from above in `if (status != RadarStatusSuccess)`

lmeier: we need just one, but this is the one we should keep

current behavior is that even if `RadarStatusSuccess`, we fall through to `[[RadarDelegateHolder sharedInstance] didFailWithStatus:status];` when calling `/track/replay`

corypisano: hm is that current behavior intentional though? Seemed like that was the "Radar error received" logs even though no error

lmeier: Ok, I actually figured out how to get toolkit working here while making this behave in a more standard manner. https://github.com/radarlabs/radar-sdk-ios/pull/274/commits/a44b46eb4710b4a67f66f6d40c04b195a290fb90

corypisano (approved):

PR #279, opened 2023-10-19
startTrackingVerified()

nickpatrick:
- adds `startTrackingVerified(token:interval:beacons:)` which calls `trackVerified()` or `trackVerifiedToken()` on an `interval` or on network change (e.g., new IP address, connect to different wi-fi network) and delivers updates to a delegate
- refactors `trackVerified()` and `trackVerifiedToken()` logic into `RadarVerificationManager`
- adds `RadarVerifiedDelegate` with `didUpdateToken()` callback (modifying `RadarDelegate` would be a breaking change and potentially disruptive to non-Fraud users)

### DIFF
--- a/Example/Example/AppDelegate.swift
+++ b/Example/Example/AppDelegate.swift
@@ -10,7 +10,7 @@ import UserNotifications
 import RadarSDK
 
 @UIApplicationMain
-class AppDelegate: UIResponder, UIApplicationDelegate, UNUserNotificationCenterDelegate, CLLocationManagerDelegate, RadarDelegate {
+class AppDelegate: UIResponder, UIApplicationDelegate, UNUserNotificationCenterDelegate, CLLocationManagerDelegate, RadarDelegate, RadarVerifiedDelegate {
 
     let locationManager = CLLocationManager()
 
@@ -22,9 +22,13 @@ class AppDelegate: UIResponder, UIApplicationDelegate, UNUserNotificationCenterD
         self.requestLocationPermissions()
 
         // Replace with a valid test publishable key
-        Radar.initialize(publishableKey: "prj_test_pk_0000000000000000000000000000000000000000")
+        Radar.initialize(publishableKey: "org_test_pk_5857c63d9c1565175db8b00750808a66a002acb8")
         Radar.setDelegate(self)
+        Radar.setVerifiedDelegate(self)
+        
+        Radar.startTrackingVerified(token: true, interval: 60, beacons: true)
 
+        /*
         if UIApplication.shared.applicationState != .background {
             Radar.getLocation { (status, location, stopped) in
                 print("Location: status = \(Radar.stringForStatus(status)); location = \(String(describing: location))")
@@ -175,6 +179,7 @@ class AppDelegate: UIResponder, UIApplicationDelegate, UNUserNotificationCenterD
 
             print("Log Conversion: status = \(Radar.stringForStatus(status)); event = \(String(describing: event))")
         }
+         */
 
         return true
     }
@@ -248,4 +253,8 @@ class AppDelegate: UIResponder, UIApplicationDelegate, UNUserNotificationCenterD
         self.notify(message)
     }
 
+    func didUpdateToken(_ token: String) {
+        
+    }
+    
 }
--- a/Example/Example/AppDelegate.swift
+++ b/Example/Example/AppDelegate.swift
@@ -22,13 +22,10 @@ class AppDelegate: UIResponder, UIApplicationDelegate, UNUserNotificationCenterD
         self.requestLocationPermissions()
 
         // Replace with a valid test publishable key
-        Radar.initialize(publishableKey: "org_test_pk_5857c63d9c1565175db8b00750808a66a002acb8")
+        Radar.initialize(publishableKey: "prj_test_pk_0000000000000000000000000000000000000000")
         Radar.setDelegate(self)
         Radar.setVerifiedDelegate(self)
-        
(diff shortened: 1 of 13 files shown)


RadarSDK/Include/Radar.h
```diff
@@ -463,6 +464,13 @@ typedef void (^_Nonnull RadarLogConversionCompletionHandler)(RadarStatus status,
  */
 + (void)trackVerifiedTokenWithCompletionHandler:(RadarTrackTokenCompletionHandler _Nullable)completionHandler NS_SWIFT_NAME(trackVerifiedToken(completionHandler:));
 
+/**
+ Starts tracking the user's location with device integrity information for location verification use cases.
+ 
+ @warning Note that you must configure SSL pinning before calling this method.
+ */
++ (void)startTrackingVerified:(BOOL)token NS_SWIFT_NAME(startTrackingVerified(token:));
```
lmeier, line 472: We already have the pattern above elsewhere, but a minor style change I'd advocate for is naming bools in a more bool-y way, e.g. `useToken` or `encoded` instead of `token` (which elsewhere is a `NSString`)

RadarSDK/RadarVerificationManager.m
```diff
@@ -31,6 +45,144 @@ + (instancetype)sharedInstance {
...
+                                                                 completionHandler(status, token);
+                                                             }];
+                                                         }
+                                                     }];
+                                             }];
+                             }];
+        }];
+}
+
+- (void)startTrackingVerified:(BOOL)token {
```
lmeier, line 146: We're not presently `[RadarSettings setTracking:YES];` so `[Radar isTracking`] = NO;`.

IDK yet the best way of having incorporating / handling ordinary tracking with verified tracking. Maybe we want a separate bool?

Also we probably need a `stopTracking`?

RadarSDK/Radar.m
```diff
@@ -244,101 +244,15 @@ + (void)trackOnceWithLocation:(CLLocation *)location completionHandler:(RadarTra
...
-                                                               [RadarUtils runOnMainThread:^{
-                                                                   completionHandler(status, token);
-                                                               }];
-                                                           }
-                                                       }];
-                                               }];
-                                 }];
-            }];
++ (void)startTrackingVerified:(BOOL)token {
+    [[RadarVerificationManager sharedInstance] startTrackingVerified:token];
```
lmeier, line 255: as this function potentially takes a whole Interval (presently 300s) to call `trackVerified`, do we want to manually call `trackVerified` and then start the timer?

lmeier (commented):
some thoughts on first pass. also thinking about what the queue / concurrency solve should be

RadarSDK/RadarVerificationManager.m
nickpatrick: I think this needs to operate totally separately from `startTracking()`. Most common usage pattern is going to be to start this on app open / login and then never stop it. For now, if you want more fine-grained control, you can call `trackVerified()` manually. We can always add `stopTrackingVerified()` later if needed

RadarSDK/Include/Radar.h
nickpatrick: I'd rather follow existing patterns. I think this is fine with good docs

RadarSDK/Radar.m
nickpatrick: Good call. In practice it seems to get called right away because the "network connected" callback seems to fire right away, but I added an explicit call

I also added some synchronization with `@synchronized(self)`

nickpatrick: I ended up removing the synchronization bc I don't think it's needed with the logic in `RadarAPIHelper`. Can always add it back later if needed

Example/Example/AppDelegate.swift
```diff
@@ -248,4 +249,8 @@ class AppDelegate: UIResponder, UIApplicationDelegate, UNUserNotificationCenterD
         self.notify(message)
     }
 
+    func didUpdateToken(_ token: String) {
+
```
lmeier, line 253: Nit, but maybe we'd want `self.notify(token)` here, right?

RadarSDK/RadarAPIClient.m
```diff
@@ -384,6 +384,10 @@ - (void)trackWithLocation:(CLLocation *_Nonnull)location
                                 }
 
                                 NSString *token = (NSString *)tokenObj;
+                                
+                                if (location) {
+                                    [[RadarDelegateHolder sharedInstance] didUpdateToken:token];
+                                }
```
lmeier, line 390: Why `if (location)`? It's a nonnull, so it's always true?

lmeier (commented):
looking good after my first pass this morning

RadarSDK/RadarAPIClient.m
nickpatrick: Good Q. I think I just copied and pasted from https://github.com/radarlabs/radar-sdk-ios/blob/f63705ef38a2399ac14dc085b2680a7eb43ff1ae/RadarSDK/RadarAPIClient.m#L450. I changed to `if (token)` (may not be necessary but might as well be extra safe and not deliver `nil` tokens to the delegate)

Example/Example/AppDelegate.swift
nickpatrick: We could. I actually didn't add a call to `startTrackingVerified()` to the example app since there's so much going on in there. But I did want to add the `RadarVerifiedDelegate` as an extra sanity check that things compiled and in case we wanted to add it later

PR #281, opened 2023-10-30
Ramp up / ramp down

lmeier:


RadarSDK/RadarLocationManager.m
```diff
@@ -223,7 +242,7 @@ - (void)startUpdates:(int)interval {
             [self.timer invalidate];
         }
 
-        self.timer = [NSTimer scheduledTimerWithTimeInterval:interval
+        self.timer = [NSTimer scheduledTimerWithTimeInterval:(interval + 0.1)
```
lmeier, line 245: Gonna pull this into a separate PR — doing this b/c I'm seeing lots of 9.95 intervals skipped

RadarSDK/RadarLocationManager.m
```diff
@@ -461,12 +586,18 @@ - (void)replaceSyncedGeofences:(NSArray<RadarGeofence *> *)geofences {
...
+    RadarTrackingOptions *trackingOptions = [Radar getTrackingOptions];
+
+
     for (int i = 0; i < numGeofences; i++) {
         RadarGeofence *geofence = [geofences objectAtIndex:i];
         NSString *geofenceId = geofence._id;
         NSString *identifier = [NSString stringWithFormat:@"%@%@", kSyncGeofenceIdentifierPrefix, geofenceId];
         RadarCoordinate *center;
         double radius = 100;
+        bool rampUpGeofenceAndDifferentFromOriginalGeofence = NO;
```
KennyHuRadar, line 600: A little confused by the naming of this variable? Can you share a little more context? Thanks!

lmeier: Yea, `rampUpGeofenceAndDifferentFromOriginalGeofence` is meant to track whether or not we should be populating local notifications onto the synced geofence. If the geofence has the geometry of the original geofence synced, the inverse of which `rampUpGeofenceAndDifferentFromOriginalGeofence` is meant to signify, then we can sync the notifications, else we need to skip.

lmeier: I'm gonna rename to something like `skipSyncingLocalNotifications`

RadarSDK/RadarLocationManager.m
```diff
@@ -411,6 +474,68 @@ - (void)updateTrackingFromMeta:(RadarMeta *_Nullable)meta {
 
 }
 
+
+// Call this method when the tracking state changes
+- (void)changeTrackingState:(BOOL)rampedUp {
```
lmeier, line 479: this name can be improved

RadarSDK/RadarLocationManager.m
```diff
@@ -892,6 +1101,11 @@ - (void)locationManager:(CLLocationManager *)manager didEnterRegion:(CLRegion *)
         return;
     }
 
+    if ([region.identifier hasPrefix:kRampUpGeofenceIdentifierPrefix]) {
+
+        return;
+    }
+
```
lmeier, line 1108: Hmmm, do we actually want this?

Guessing this was done to avoid the issue with notifications, which have since been handled

RadarSDK/RadarTrackingOptions.m
```diff
@@ -111,6 +114,34 @@ + (RadarTrackingOptions *)presetEfficient {
     options.useVisits = YES;
     options.useSignificantLocationChanges = NO;
     options.beacons = NO;
+    options.rampUpRadius = 200;
+
+    return options;
+}
+
++ (RadarTrackingOptions *)rampedUpOptions {
```
tjulien, line 122: Is `continuous` not good enough?  If not, wondering if we reuse it but change 1-2 of the options via code instead of creating a new preset.  If we did want to create a new preset we would probably want to name it `presetXxx` and also think of a more descriptive name then "ramped up".

lmeier: think essentially we want continuous but with more frequent updates / faster sync interval.

IMO the intended tracking options are what works for microdetection / DD use case

tjulien: Makes sense.  `RampingOption` and `StateChange` seem to add alot of complexity that will be hard to get right I think.  Two things I'm thinking about:

1) Would it help if `Radar.startTripWithOptions()` accepted a second `trackingOptions` param that if set kicks in when within the radius?  And we would expand the "previous tracking options" concept from 1 to a list, maybe.
2) Would it help if a single bundle of tracking options had a few more fields representing the ramped up needs?  It might only be the update intervals and sync intervals.  It would be less generic bc you couldn't change anything else when ramping, but the code to handle those values would be alot simpler I suspect.

lmeier: I've taken a stab at number 2, cutting a waypoint build now: https://github.com/radarlabs/radar-sdk-ios/pull/281/commits/e68d2f5c0bf9e14679cde8249eb90a199056eeaf

tjulien: tactical thing, I think the force pushes are making it hard to see what changed since I looked at this last.  Everything looks like it was pushed 3 days ago. I'll fiddle around in github, see if I can find a way.  I guess lacking that, what changed to implement #2?

But more generally, the idea behind my comment was to get rid of `RampingOption` and `StateChange` because I think they add too much complexity.  I'll spend the day tomorrow trying to understand how those work and try to form a more detailed opinion.

lmeier: Oh yikes, that makes sense. Didn't mean to be force pushing. Will watch out for that.


What I did was remove the notion of there being any difference between "ramped up" tracking options and "normal / ramped down" tracking options. Instead, the tracking options are fixed — eliminating some of the state complexity you pointed out, not having to worry about keeping track of the tracking options over time.

Still, we need to keep track of past ramped state so that we can timeout of being ramped up. This is how we address the problem of being persistently (or for a long time) close to a ramp up geofence. 

We might also choose to ramp down when `stopped: true` (GK suggested this today).

lmeier: Look forward to hearing the more detailed opinion that you form

tjulien: Got it. Are`RampingOption` and `StateChange` primarily needed to be able to timeout? Or are they important for other reasons? Ramping down or shutting down when stopped is an interesting idea. 

How long is the timeout?

RadarSDK/Include/RadarTrackingOptions.h
```diff
@@ -154,6 +154,18 @@ typedef NS_ENUM(NSInteger, RadarTrackingOptionsSyncLocations) {
  */
 @property (nonatomic, assign) BOOL beacons;
 
+/**
+ Determines a radius around the center of each geofence under which tracking should ramp up.
+ */
+
+@property (nonatomic, assign) int rampUpRadius;
```
tjulien, line 161: thinking a bit about the naming here, I think we may want to talk about this as "nearby tracking" and name these `nearbyTrackingDistance` and `nearbyTrackingInterval`

RadarSDK/Include/RadarTripOptions.h
```diff
@@ -52,6 +52,11 @@ NS_ASSUME_NONNULL_BEGIN
  */
 @property (nullable, nonatomic, copy) NSDate *scheduledArrivalAt;
 
+/**
+ Optional radius in meters at which to begin ramping up tracking. Takes precedence over both geofence specific and tracking option ramp-up radii.
+ */
+@property (nonatomic, assign) int rampUpRadius;
```
tjulien, line 58: should this also have `rampUpInterval`?

RadarSDK/RadarSettings.h
```diff
@@ -33,6 +33,8 @@ NS_ASSUME_NONNULL_BEGIN
 + (void)setAnonymousTrackingEnabled:(BOOL)enabled;
 + (BOOL)tracking;
 + (void)setTracking:(BOOL)tracking;
++ (BOOL)rampedUp;
```
tjulien, line 36: think we can call these `nearbyTracking`

RadarSDK/RadarTrackingOptions.m
```diff
@@ -77,7 +81,7 @@ + (RadarTrackingOptions *)presetResponsive {
     options.startTrackingAfter = nil;
     options.stopTrackingAfter = nil;
     options.syncLocations = RadarTrackingOptionsSyncAll;
-    options.replay = RadarTrackingOptionsReplayStops;
+    options.replay = RadarTrackingOptionsReplayAll;
```
tjulien, line 84: don't forget to revert before merging

RadarSDK/Include/RadarTrackingOptions.h
lmeier: Interesting... open to it, let's discuss in our 1 on 1

RadarSDK/Radar.m
```diff
@@ -564,11 +564,12 @@ + (void)startTripWithOptions:(RadarTripOptions *)tripOptions
                                              if (status == RadarStatusSuccess) {
                                                  [RadarSettings setTripOptions:tripOptions];
 
-                                                 if (Radar.isTracking) {
+                                                 if (Radar.isTracking && ![RadarSettings rampedUp]) {
```
tjulien, line 567: Is this so that ramped up tracking overrides trackingOptions that are used when a trip starts?  If so we may want to take the approach of having people set the ramp up options on both on trip and off trip tracking options to keep this simple.  Then I think we could remove this change?

RadarSDK/Include/RadarTripOptions.h
lmeier: No, we don't want a `rampUpInterval` here actually. We check for the trip option `rampUpRadius` in our `replaceSyncedGeofences` _iff_ the geofence in question is the trip destination geofence.

The `rampUpInterval` comes up in `handleLocation`, and applies regardless of which geofence has caused the ramp up.

If one wanted to have a different `rampUpInterval` when on trips, they could achieve that via passing in tracking options with the desired `rampUpInterval`.

RadarSDK/RadarTrackingOptions.m
lmeier: Timeout is 20min in 1 hour OR 120 minutes in 12 hours

RadarSDK/RadarLocationManager.m
```diff
@@ -411,12 +451,82 @@ - (void)updateTrackingFromMeta:(RadarMeta *_Nullable)meta {
 
 }
 
+
+// Call this method when the tracking state changes
+- (void)changeRampedState:(BOOL)rampedUp {
```
tjulien, line 456: Things I'm thinking about:
1) Should we just not support a timeout in v1 to avoid the complexity of `StateChange`?
2) Can we get away with a single variable to track the running total of time spent ramped up instead of a list of `StateChange`s?
3) Should we make the timeout configurable on tracking options, something like 'nearbyTrackingTimeout`?

RadarSDK/RadarLocationManager.m
```diff
@@ -411,12 +451,82 @@ - (void)updateTrackingFromMeta:(RadarMeta *_Nullable)meta {
...
+    [RadarSettings setRampedUp:rampedUp];
+    
+    StateChange *change = [[StateChange alloc] init];
+    change.timestamp = [NSDate date]; // Current timestamp
+    change.rampedUp = rampedUp;
+    [self.stateChanges addObject:change];
+}
+
+- (NSDictionary *)calculateRampedUpTimesAndCleanup {
```
tjulien, line 465: Things I'm thinking about:
1) Should we just not support a timeout in v1 to avoid the complexity of `StateChange`?
2) Can we get away with a single variable to track the running total of time spent ramped up instead of a list of `StateChange`s?
3) Should we make the timeout configurable on tracking options, something like 'nearbyTrackingTimeout`?

RadarSDK/Radar.m
lmeier: Ahhh, with the changes to collapse everything into the existing tracking options (no new `rampedUpTrackingOption` preset), we can revert this change, nice catch.

RadarSDK/RadarLocationManager.m
```diff
@@ -411,12 +451,82 @@ - (void)updateTrackingFromMeta:(RadarMeta *_Nullable)meta {
...
+}
+
+
 - (void)restartPreviousTrackingOptions {
     RadarTrackingOptions *previousTrackingOptions = [RadarSettings previousTrackingOptions];
     [[RadarLogger sharedInstance] logWithLevel:RadarLogLevelDebug message:@"Restarting previous tracking options"];
 
     if (previousTrackingOptions) {
         [Radar startTrackingWithOptions:previousTrackingOptions];
+        if ([RadarSettings rampedUp]) {
```
tjulien, line 524: this is block here is pretty complicated (in a part of the code that's already kinda tricky)- don't think we want to be special casing this this much

lmeier: Yes, good call. This can and should be removed.

RadarSDK/RadarLocationManager.m
lmeier: Hmmm, I like the idea of having this value be configurable via the tracking options.

Note that it's less of a "timeout" and more of a "maximum allowed running time", because you can ramp up and ramp down repeatedly if you're near or transitioning through the edge of the ramp up radius.

Need to think it through a bit, but I'm pretty sure that if we can go with just a single "timeout"/"max running time", we can just store and intelligently increment a timestamp that gets compared. I'll take a stab

RadarSDK/RadarLocationManager.m
```diff
@@ -90,8 +95,8 @@ - (instancetype)init {
         _locationManager.allowsBackgroundLocationUpdates = [RadarUtils locationBackgroundMode] && [CLLocationManager authorizationStatus] == kCLAuthorizationStatusAuthorizedAlways;
 
         _lowPowerLocationManager = [CLLocationManager new];
-        _lowPowerLocationManager.desiredAccuracy = kCLLocationAccuracyThreeKilometers;
-        _lowPowerLocationManager.distanceFilter = 3000;
+        _lowPowerLocationManager.desiredAccuracy = kCLLocationAccuracyHundredMeters;
+        _lowPowerLocationManager.distanceFilter = kCLDistanceFilterNone;
```
lmeier, line 99: `lowPowerLocationManager` changes a la https://github.com/radarlabs/radar-sdk-ios/pull/303

lmeier:
Closing as we're opting to do this server side

PR #287, opened 2023-12-07
Add buffer to sync interval

lmeier:
When setting `desiredStoppedUpdateInterval`, `desiredMovingUpdateInterval` and `desiredSyncInterval` to the same value (e.g. 10), I've been observing cases where we skip a sync because the timer fired just slightly prematurely, due to the difference in when we start the `track` and when we update `lastSentAt`. This buffer prevents that edge case and allows for expected behavior when setting sync intervals as one would normally.


KennyHuRadar (approved):
lgtm

PR #288, opened 2023-12-07
[MAPS-694] remove instances of `groups` from places search

j949wang:


docs/docsets/RadarSDK.docset/Contents/Resources/Documents/search.json
```diff
@@ -1 +1,1747 @@
...
\ No newline at end of file
+{
```
j949wang, line 1: Oops prettier made in not 1 line

lmeier:
Are we completely eliminating `groups` from the API at this time or just deprecating it, keeping up a version for backwards compatibility?

Ideally we'd have multiple function definitions here, one that's deprecated and one that isn't, so that we can make this a non-breaking change.

j949wang:
I think the goal is to eliminate `groups` entirely since its no longer documented. This is sort of the tail end of clean up work post geocoder cutover for places. I think conceivably we could have multiple function definitions to start while the server clean up goes through and then eliminate them here afterwards.

PR #293, opened 2024-01-08
add event notification prefix

KennyHuRadar:
The notification identifier is currently just the `eventId`. This can lead to developer error as consumers of the SDK may also be listening to `radarEvents` to generate notifications. In particular, it may seem intuitive to the developer to use the `eventId` as the identifier of the notification. However, we are also doing the same thing. Registering 2 notifications with the same identifier at the same time will the cause the first notification to be dropped silently. 

This PR fixes this issue by prefixing the identifier with another string to avoid such conflicts.  

QA: successfully generate notifications with waypoint.
image


lmeier (commented):
bump version

RadarSDK.xcodeproj/project.pbxproj
```diff
@@ -981,7 +981,7 @@
 				GCC_WARN_UNUSED_FUNCTION = YES;
 				GCC_WARN_UNUSED_VARIABLE = YES;
 				IPHONEOS_DEPLOYMENT_TARGET = 10.0;
-				MARKETING_VERSION = 3.8.13;
+				MARKETING_VERSION = 3.8.14;
```
lmeier, line 984: fix this

RadarSDK.xcodeproj/project.pbxproj
```diff
@@ -1039,7 +1039,7 @@
...
```
lmeier, line 1042: fix this

PR #294, opened 2024-01-09
expandUnits -> mailable

york-wei:
This PR changes the `expandUnits` param for autocomplete to `mailable`

### DIFF
--- a/MIGRATION.md
+++ b/MIGRATION.md
@@ -1,7 +1,7 @@
 # Migration guides
 
 ## 3.8.x to 3.9.0
-- The `Radar.autocompleteQuery(query:near:layers:limit:country:expandUnits:completionHandler:)` method is now `Radar.autocompleteQuery(query:near:layers:limit:country:mailable:completionHandler:)`.
+- The `Radar.autocomplete(query:near:layers:limit:country:expandUnits:completionHandler:)` method is now `Radar.autocomplete(query:near:layers:limit:country:mailable:completionHandler:)`.
 
 ## 3.6.x to 3.7.0
 - Custom events have been renamed to conversions.
--- a/MIGRATION.md
+++ b/MIGRATION.md
@@ -1,5 +1,8 @@
 # Migration guides
 
+## 3.8.x to 3.9.0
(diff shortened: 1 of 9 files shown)


RadarSDK/Radar.m
```diff
@@ -838,14 +838,14 @@ + (void)autocompleteQuery:(NSString *_Nonnull)query
                    layers:(NSArray *_Nullable)layers
                     limit:(int)limit
                   country:(NSString *_Nullable)country
-              expandUnits:(BOOL)expandUnits
+                 mailable:(BOOL)mailable
```
lmeier, line 841: Let's define another `(void)autocompleteQuery` method with these like this, while also keeping the old one that takes an `expandUnits`.

For the old one, we should note in the comment that it's deprecated. We can discard the `expandUnits` (it's always true per Jeff's JS [change](https://github.com/radarlabs/radar-sdk-js/blob/fd0850e004b0952e52e2443caf550bb9676832ea/src/types.ts#L341), and pass `NO` into `mailable`.

The reason we want to do this this way is so that this isn't a breaking change that requires a migration. (People's SDKs often auto-update, for a patch/minor version bump like this here we shouldn't require code changes).

lmeier: (You'll need to have both headings in `Radar.h` as well).

lmeier (changes requested):
Nice start! See comment for a slight diff in how we want to do this to make it non-breaking.

Also, please take a look at the automated tests in CI. Looks like one had broken.

lmeier (commented):
Looks like you previously had changed the example and then reverted that change. Let's get that back in.

Once you do that and we test this, this is good to go.

york-wei:
Verified that the `mailable` param is being sent 👍

MIGRATION.md
```diff
@@ -1,5 +1,8 @@
 # Migration guides
 
+## 3.8.x to 3.9.0
+- The `Radar.autocomplete(query:near:layers:limit:country:expandUnits:completionHandler:)` method is now `Radar.autocomplete(query:near:layers:limit:country:mailable:completionHandler:)`.
```
lmeier, line 4: Call out explicitly that `expandUnits` is deprecated, and that the API now always behaves as if `expandUnits` true.

lmeier (approved):
lgtm, congrats on first pr @york-wei ! 💯 🚀 🥇

PR #296, opened 2024-01-16
Add geometry to beacons

KennyHuRadar:
This PR add a new field that serializes the geometry (coordinates) of the beacons.
It will now be possible for consumers to overlay the location of the beacons on the UI. 
image


KennyHuRadar:
The versioning is not bumped yet as that will cause issues with current waypoint builds needed for testing, will bump before merging.

lmeier (commented):
version # is wrong. Otherwise, lgtm

lmeier (approved):

PR #297, opened 2024-01-16
Ignore 0 values of latitude and longitude on client

york-wei:
- `isValid` check failed to filter out 0 coordinates previously due to `lon` and `lat` values not being precisely 0


lmeier:
This is looking good to me.

Because this is in the critical path, we should do some dogfooding. I propose:
- run a build of Waypoint locally on your simulator. Change location to something like (40.124567, 0.000000) and verify that that fails while (40.124567, 0.001) succeeds
- have the build on your device for ~24 hours, tracking in responsive, and check your user activity page a few times to verify that we're not getting spurious `error_location` errors from this

RadarSDK/CLLocation+Radar.m
```diff
@@ -8,17 +8,23 @@
 @import Foundation;
 #import "CLLocation+Radar.h"
 
+const double DEGREE_EPSILON = 0.00000001;
```
KennyHuRadar, line 11: I believe obj C has this `DBL_EPSILON` global const that we can use, maybe we can consider its use here? I think it can be imported with `float.h`.

KennyHuRadar (changes requested):

RadarSDK/CLLocation+Radar.m
york-wei: `DBL_EPSILON` is approximately 2.2e-16 which is too precise I think, it would give similar results to just using the `!=` operator like we did before

york-wei:
@lmeier verified both :+1:

> This is looking good to me.
> 
> Because this is in the critical path, we should do some dogfooding. I propose:
> 
> * run a build of Waypoint locally on your simulator. Change location to something like (40.124567, 0.000000) and verify that that fails while (40.124567, 0.000001) succeeds
> * have the build on your device for ~24 hours, tracking in responsive, and check your user activity page a few times to verify that we're not getting spurious `error_location` errors from this
> 
> Once you do that, leave a comment, and we can move this forward.

lmeier (approved):
nice 🔥

PR #299, opened 2024-01-25
Include beacons in handleLocation track call

york-wei:
- `nil` was being passed into the `beacons` param for the `track` call, causing beacon flapping
- Also send up empty array instead of `nil` when nearby beacon search returns no beacons


RadarSDK/RadarLocationManager.m
```diff
@@ -830,39 +830,71 @@ - (void)sendLocation:(CLLocation *)location stopped:(BOOL)stopped source:(RadarL
...
-                    if (beaconUUIDs && beaconUUIDs.count) {
-                        [self replaceSyncedBeaconUUIDs:beaconUUIDs];
-                    } else if (beacons && beacons.count) {
-                        [self replaceSyncedBeacons:beacons];
-                    }
+        
+        [[RadarAPIClient sharedInstance]
+         searchBeaconsNear:location
+         radius:1000
+         limit:10
```
lmeier, line 862: not yet necessarily, but I think at some point I think we'll want this limit set by a `featureSetting` so that we can choose how many beacons & how many geofences we sync down. (Might not matter if we're merely ranging beacons and not storing them as `CLRegion`s.

PR #300, opened 2024-01-26
Feature flagged removal of beacon monitoring

york-wei:
- Removed beacon monitoring to reduce flapping


RadarSDK/RadarBeaconManager.m
```diff
@@ -265,21 +265,13 @@ - (CLBeaconRegion *)regionForUUID:(NSString *)uuid {
 }
 
 - (void)handleBeacons {
-    if (self.beaconUUIDs.count == 0 && self.nearbyBeaconIdentifiers.count + self.failedBeaconIdentifiers.count == self.beacons.count) {
```
york-wei, line 268: Not sure if this is intentional but this check meant that we did not use ranging to determine beacon enters/exits when using UUIDs

lmeier (approved):

PR #301, opened 2024-01-29
FENCE-1758: Add hostname to init

york-wei:


nickpatrick:
@york-wei @lmeier Don't have context on this work, but I don't think this is something we want to expose to developers as a public API. I know we change the hostname to staging for our own development, but customers shouldn't be doing this. Historically we've just modified `UserDefaults` directly (I realize that's going away)

lmeier:
@york-wei mind exposing this as a `region` param which then corresponds to the host? Tagged you in a slack thread explicating the details. Thanks.

RadarSDK/Radar.m
```diff
@@ -71,6 +71,12 @@ + (void)initializeWithPublishableKey:(NSString *)publishableKey {
     
 }
 
++ (void)initializeWithPublishableKey:(NSString *)publishableKey
+                              region:(RadarHostRegion)region {
+    [RadarSettings setHostRegion:region];
+    [Radar initializeWithPublishableKey:publishableKey];
+}
```
lmeier, line 78: If we use this way of setting the host (region) in Waypoint, it'll break the way we change the host in SU... that's probably alright (we just won't use this in Waypoint)

lmeier: Actually, we can set the `region` in demo app config, and use that to use this if set, otherwise, just using the original one. That will allow us to dogfood this.

Let's wait for the europe / na regions to go live and have this in Waypoint before we merge

PR #303, opened 2024-02-02
FENCE-1850: Feature flagged background responsive mode changes

york-wei:


lmeier:
@york-wei mind cutting a Waypoint pointing to this and circulating?

lmeier:
We might also want to revert this commit as a part of testing this: https://github.com/radarlabs/radar-sdk-ios/commit/675cff177f71170e99d0696102f4e79406da701f

lmeier:
https://linear.app/radarlabs/issue/FENCE-1850/low-power-location-manager-changes

RadarSDK/RadarLocationManager.m
```diff
@@ -727,7 +727,7 @@ - (void)handleLocation:(CLLocation *)location source:(RadarLocationSource)source
                 duration = -[location.timestamp timeIntervalSinceNow];
             }
             BOOL arrival = source == RadarLocationSourceVisitArrival;
-            stopped = (distance <= options.stopDistance && duration >= options.stopDuration) || arrival;
+            stopped = (distance <= options.stopDistance && duration >= options.stopDuration);
```
york-wei, line 730: Removed stop detection using arrival, can feature flag this too if needed

lmeier: I think we should feature flag such that if it's default (3000 & 3000) we do count arrivals as stop (old impl) and if it's non default we don't.

PR #306, opened 2024-02-13
Include cross-platform info in requests

york-wei:


RadarSDK/RadarAPIClient.m
```diff
@@ -243,6 +250,10 @@ - (void)trackWithLocation:(CLLocation *_Nonnull)location
     params[@"country"] = [RadarUtils country];
     params[@"timeZoneOffset"] = [RadarUtils timeZoneOffset];
     params[@"source"] = [Radar stringForLocationSource:source];
+    if ([RadarSettings crossPlatform]) {
+        params[@"xPlatformType"] = [RadarSettings crossPlatformSDKType];
+        params[@"xPlatformSdkVersion"] = [RadarSettings crossPlatformSDKVersion];
+    }
```
lmeier, line 256: let's else a `params[@"xPlatformType"] = @("native");` to make it clear on the user

lmeier (commented):
1 small thing

nickpatrick:
Couple small stylistic things we may want to discuss:
- xPlatform or crossPlatform? be consistent everywhere?
- ___SdkVersion or SDKVersion? fwiw Android tends to style like the first, iOS like the second
- native or Native? fwiw we send deviceType = Web, iOS, Android which maybe implies we send Native, Capacitor, Cordova (capitalized)

lmeier (approved):
thanks for getting this in quickly ⚡ 

lgtm

PR #313, opened 2024-03-07
Driving detection & motion data

york-wei:


RadarSDK/RadarLocationManager.m
```diff
@@ -94,6 +94,58 @@ - (instancetype)init {
...
+                        @"driving" : @(activity.automotive),
+                        @"cycling" : @(activity.cycling),
+                        @"unknown" : @(activity.unknown),
+                        @"timestamp" : @([activity.startDate timeIntervalSince1970]),
+                        @"confidence" : @(activity.confidence)
+                    }];
+                }
+            }];
+        
+            _motionManager = [CMMotionManager new];
```
KennyHuRadar, line 117: We are creating an instance of CMMotionManager on the main thread, apparently the init of the manager is slow and we should do this on another thread.
https://stackoverflow.com/questions/55629406/ios-12-2-cmmotionmanager-is-blocking-the-main-thread

KennyHuRadar:
There are anecdotal evidence that CMMotionManager is hard on the battery. I don't think there are any PR specific action items for us to take here but when we integrate it to waypoint or any other example app we should examine the battery impact.

We may end up handling this downside like how we handle constant high frequency gps tracks via tracking settings and including it only in continuous mode.

RadarSDK/RadarLocationManager.m
```diff
@@ -94,6 +101,43 @@ - (instancetype)init {
         _lowPowerLocationManager.distanceFilter = 3000;
         _lowPowerLocationManager.allowsBackgroundLocationUpdates = [RadarUtils locationBackgroundMode];
 
+        if ([RadarSettings useLocationMetadata]) {
```
KennyHuRadar, line 104: I think this chunk should not be called here as the `RadarLocationManager` is actually initialized before our call to `getConfigForUsage`.  I don't think server-side configurations of `useLocationMetaData` will be applied. I think we should move it to `updateTracking`

RadarSDK/RadarLocationManager.m
```diff
@@ -204,13 +249,18 @@ - (void)startTrackingWithOptions:(RadarTrackingOptions *)trackingOptions {
...
     [RadarSettings setTracking:NO];
+
     
     RadarSdkConfiguration *sdkConfiguration = [RadarSettings sdkConfiguration];
     if (sdkConfiguration.extendFlushReplays) {
         [[RadarLogger sharedInstance] logWithLevel:RadarLogLevelInfo type:RadarLogTypeSDKCall message:@"Flushing replays from stopTracking()"];
         [[RadarReplayBuffer sharedInstance] flushReplaysWithCompletionHandler:nil completionHandler:nil];
     }
 
+    if (sdkConfiguration.useLocationMetadata) {
```
KennyHuRadar, line 260: Do we also need to call this inside `updateTracking`? I don't think this method gets called by `updateTracking`

Example/Example/Example-Bridging-Header.h
```diff
@@ -6,3 +6,4 @@
 //
 
 #import <RadarSDK/RadarSDK.h>
+#import <RadarMotion/RadarSDKMotion.h>
```
lmeier, line 9: Hmmmm, given the new 1 line setup of this, do we need this import

RadarSDK.podspec
```diff
@@ -1,6 +1,6 @@
 Pod::Spec.new do |s|
   s.name                  = 'RadarSDK'
-  s.version               = '3.16.0'
+  s.version               = '3.16.1'
```
lmeier, line 3: Probably worth making this a `3.17.0` given the size of this.

RadarSDK/Radar.m
```diff
@@ -43,6 +43,12 @@ + (id)sharedInstance {
 
 + (void)initializeWithPublishableKey:(NSString *)publishableKey {
     [[RadarLogger sharedInstance] logWithLevel:RadarLogLevelInfo type:RadarLogTypeSDKCall message:@"initialize()"];
+    
+    Class RadarSDKMotion = NSClassFromString(@"RadarSDKMotion");
+    if (RadarSDKMotion) {
+        id radarSDKMotion = [[RadarSDKMotion alloc] init];
+        [RadarActivityManager sharedInstance].radarSDKMotion = radarSDKMotion;
+    }
```
lmeier, line 51: Thinking out loud about someone deciding to turn this on: basically, in the first build that they import `RadarSDKMotion`, they'll have to add the new permission.

Given that we have this check above, we'll not need to do any version gating on `useLocationMetadata`, right?

RadarSDK/RadarAPIClient.m
```diff
@@ -326,6 +326,34 @@ - (void)trackWithLocation:(CLLocation *_Nonnull)location
...
     }
     params[@"appId"] = [[NSBundle mainBundle] bundleIdentifier];
+    RadarSdkConfiguration *sdkConfiguration = [RadarSettings sdkConfiguration];
+    if (sdkConfiguration.useLocationMetadata) { 
+        NSMutableDictionary *locationMetadata = [NSMutableDictionary new];
+        locationMetadata[@"motionActivityData"] = [RadarState lastMotionActivityData];
+        locationMetadata[@"accelerometerData"] = [RadarState lastAccelerometerData];
+        locationMetadata[@"gyroData"] = [RadarState lastGyroData];
+        locationMetadata[@"magnetometerData"] = [RadarState lastMagnetometerData];
+        locationMetadata[@"heading"] = [RadarState lastHeadingData];
```
lmeier, line 336: How do we want to send these up if they're not present (e.g. if motion activity permission wasn't granted or, in a build before `RadarSDKMotion` was added to the client app)

KennyHuRadar: this is actually a good point, I do feel that given the labels we are applying to these values, we should ensure that these values are actually "fresh". I think we should actually return null values if the data is more than 1 min old or something like that.

KennyHuRadar: Right now if the values are not available, we are assigning null to the field and the serialization gets rid of the field, which I think is the right thing to do.

Thinking about json and webhookland, having a field with value "null" just to keep the shape might not be the best pattern. It may be more user friendly to just not have the field since `{}.nonExistentField` evaluates to the null value in most languages anyways. Having a non null value of the string "null" may throw people off more, especially since that is not a pattern we have elsewhere.

RadarSDK/Radar.m
KennyHuRadar: For this chunk of code, we are not concerned about if the flags are set to allow this feature, we are just loading a side component if we are able too. We also do not even have the feature flags at this point.

Right now, its the responsibility of ActivityManager and LocationManager to decide if we want to perform these extra data collection based on the feature flags.

RadarSDK/RadarState.m
```diff
@@ -155,6 +159,19 @@ + (void)setBeaconIds:(NSArray<NSString *> *)beaconIds {
     [[NSUserDefaults standardUserDefaults] setObject:beaconIds forKey:kBeaconIds];
 }
 
++ (void) setTimeStamp:(NSString *)key {
```
lmeier, line 162: extra space b/t `(void)` and `set..`

lmeier (approved):

PR #314, opened 2024-03-11
Add new fraud events and fields

nickpatrick:


RadarSDK/RadarEvent.m
```diff
@@ -379,6 +388,8 @@ + (NSString *)stringForType:(RadarEventType)type {
         return @"user.approaching_trip_destination";
     case RadarEventTypeUserArrivedAtTripDestination:
         return @"user.arrived_at_trip_destination";
+    case RadarEventTypeUserFailedFraud:
+        return @"user.failed_fraud";
```
lmeier, line 392: Do we want a `user.passed_fraud` as well? Imagining impl's that only use `startTrackingVerified` and just have an event listener waiting for a `passed_fraud` event to set some `allow_functionality: BOOL`

nickpatrick: Something we could add later. If we add it, sort of duplicative of locations, though. Potentially useful for webhooks, but then again we could just send all locations to webhooks

lmeier (approved):

PR #317, opened 2024-03-13
Fix for adding fraud field to RadarEvents

KennyHuRadar:


nickpatrick:
Good catch, thx for this. Can we make the comment/placement match Android? (except sub nil for null) https://github.com/radarlabs/radar-sdk-android/blob/master/sdk/src/main/java/io/radar/sdk/model/RadarEvent.kt

nickpatrick:
Small typo: nill to nil

Nit: Let's also move this between trip and alternatePlaces everywhere. Makes it easier to cross-reference between iOS and Android

PR #318, opened 2024-03-19
Add coalescing trackOnce calls to initialize, resume, and location auth listener

york-wei:


york-wei:
Waypoint testing:
- on app start, Waypoint triggers two `trackOnce` calls while the SDK triggers two `trackOnce` calls via auth listener and initialize. Three calls are coalesced and last `trackOnce` call by Waypoint fires a second after (due sleep semaphore)
Screenshot 2024-03-19 at 5 00 00 PM
- on app resume, Waypoint triggers a `trackOnce` call and the SDK triggers a `trackOnce` call via resume callback. The two calls are coalesced together
Screenshot 2024-03-19 at 5 03 26 PM

"reached track completion" is only logged when a track api call has been made

PR #320, opened 2024-03-29
Make geofence geometry optional

kochis:
Interface change:

We modify the interface of `searchGeofences` to the following

```
/**
Gets the device's current location, then searches for geofences near that location, sorted by distance.

 @see https://radar.com/documentation/api#search-geofences
 */
+ (void)searchGeofences:(RadarSearchGeofencesCompletionHandler)completionHandler NS_SWIFT_NAME(searchGeofences(tags:metadata:completionHandler:));

/**
 Searches for geofences near a location, sorted by distance.

 @param near The location to search. Use null to search near the device's current location.
 @param radius The radius to search, in meters. A number between 100 and 10000. If -1 is entered, the server defaults to using unlimited radius.
 @param tags An array of tags to filter. See https://radar.com/documentation/geofences
 @param metadata A dictionary of metadata to filter. See https://radar.com/documentation/geofences
 @param limit The max number of geofences to return. A number between 1 and 1000. Recommend to be set to 100.
 @param includeGeometry A boolean indicating whether to include geometry in the response. Recommend to be set to false.

 @see https://radar.com/documentation/api#search-geofences
 */
 + (void)searchGeofencesNear:(CLLocation *_Nullable)near
                     radius:(int)radius
                       tags:(NSArray<NSString *> *_Nullable)tags
                   metadata:(NSDictionary *_Nullable)metadata
                      limit:(int)limit
            includeGeometry:(BOOL)includeGeometry
          completionHandler:(RadarSearchGeofencesCompletionHandler)completionHandler NS_SWIFT_NAME(searchGeofences(near:radius:tags:metadata:limit:includeGeometry:completionHandler:));


```
We essentially have 2 methods, a bare bones one that should be easy to use, and a maximal one with optional named arguments for more advanced use-cases. 

QA:
We were able to observe successful calls with the example app code snippet to both `getContext` and `searchGeofence`
```
let location = CLLocation(latitude: 49.68956683045484, longitude: -107.27634218307078)
Radar.getContext(location:location) { (status, location, context) in
  print("Context: status = \(Radar.stringForStatus(status)); location = \(String(describing: location)); context?.geofences = \(String(describing: context?.geofences)); context?.place = \(String(describing: context?.place)); context?.country = \(String(describing: context?.country))")
        }
Radar.searchGeofences() { (status, location, geofences) in
            print("Search geofences: status = \(Radar.stringForStatus(status)); geofences = \(String(describing: geofences))")
        }
        
Radar.searchGeofences(near: location, radius: -1, tags: nil, metadata: nil, limit: 100, includeGeometry: true) { (status, location, geofences) in
            print("Search geofences 2: status = \(Radar.stringForStatus(status)); geofences = \(String(describing: geofences))")
        }
```

with the printed output
image


RadarSDK/RadarGeofence.m
```diff
@@ -127,52 +127,49 @@ - (instancetype _Nullable)initWithObject:(id)object {
...
-
-            id coordinatesObj = ((NSDictionary *)geometryObj)[@"coordinates"];
-            if (![coordinatesObj isKindOfClass:[NSArray class]]) {
-                return nil;
-            }
-
-            NSArray *coordinatesArr = (NSArray *)coordinatesObj;
-            if (coordinatesArr.count != 1) {
-                return nil;
+            if ([geometryObj isKindOfClass:[NSDictionary class]]) {
```
lmeier, line 130: Hmmm, do you have an example of the new payload?

What we need to think through: presently the `geometry` object is where we maintain the coordinates of the `RadarGeofence` object. We're not maintaining it independently of the geometry for geofences (we _do_ maintain it independently for `RadarPlace`)

So either we need to somehow take the response and create a faux geometry where we keep the coordinates (perhaps defaulting to the above radius `0.0`) or we mirror the approach of `RadarPlace`. I think I prefer the former, maintaining a shell geometry — perhaps of a new type `Point`.

(A third option I don't think we'd opt for is having no coordinates on these geofences).

lmeier (commented):
Need to think through how we're gonna maintain coordinates for `RadarGeofence`.

Also, did we settle on `includeGeometry` being the optional param for `searchGeofences`? If so, we'll want to add that parameter to the function / api call in this PR as well.

Once we resolve the approach, @KennyHuRadar can take this to completion (and do Android & x-platform changes)

RadarSDK/RadarGeofence.m
KennyHuRadar: As it is `RadarPolygonGeometry` has a center, radius and coordinates (which is nullable). I think we can just pass in a null into the coordinates field when it is nor being returned from the server.

RadarSDK/RadarAPIClient.m
```diff
@@ -777,6 +780,9 @@ - (void)searchGeofencesNear:(CLLocation *_Nonnull)near
             [queryString appendFormat:@"&metadata[%@]=%@", key, metadata[key]];
         }
     }
+    
+    [queryString appendFormat:@"&includeGeometry=%@", includeGeometry? @"true" : @"false"];
```
lmeier, line 784:
```suggestion
    [queryString appendFormat:@"&includeGeometry=%@", includeGeometry ? @"true" : @"false"];
```

RadarSDK/RadarAPIClient.m
```diff
@@ -767,7 +768,9 @@ - (void)searchGeofencesNear:(CLLocation *_Nonnull)near
 
     NSMutableString *queryString = [NSMutableString new];
     [queryString appendFormat:@"near=%.06f,%.06f", near.coordinate.latitude, near.coordinate.longitude];
-    [queryString appendFormat:@"&radius=%d", radius];
+    if(radius > 0){
```
lmeier, line 771:
```suggestion
    if (radius > 0) {
```

RadarSDK/Include/Radar.h
```diff
@@ -845,27 +845,69 @@ logConversionWithNotification
        completionHandler:(RadarSearchPlacesCompletionHandler)completionHandler NS_SWIFT_NAME(searchPlaces(near:radius:chains:chainMetadata:categories:groups:limit:completionHandler:));
 
 /**
- Gets the device's current location, then searches for geofences near that location, sorted by distance.
+Gets the device's current location, then searches for geofences near that location, sorted by distance.
 
- @param radius The radius to search, in meters. A number between 100 and 10000.
+ @param radius The radius to search, in meters. A number between 100 and 10000. If 0 is entered, the server use its default value.
```
lmeier, line 850:
```suggestion
 @param radius The radius to search, in meters. A number between 100 and 10000. If -1 is supplied, the search is performed without a radius limit.
```

lmeier: I'm tempted to say that we actually should provide a function sans `radius` (and maybe sans the rest of the optionals) for simplicity's sake. Passing in -1 is a bit weird to get the "default" unlimited behavior

RadarSDK/Include/Radar.h
```diff
@@ -845,27 +845,69 @@ logConversionWithNotification
...
+                             tags:(NSArray<NSString *> *_Nullable)tags
+                         metadata:(NSDictionary *_Nullable)metadata
+                            limit:(int)limit
+                  includeGeometry:(BOOL)includeGeometry
+                completionHandler:(RadarSearchGeofencesCompletionHandler)completionHandler NS_SWIFT_NAME(searchGeofences(radius:tags:metadata:limit:includeGeometry:completionHandler:));
+
+/**
+  @deprecated Gets the device's current location, then searches for geofences near that location, sorted by distance.
+
+ @param radius The radius to search, in meters. A number between 100 and 10000. If 0 is entered, the server use its default value.
```
lmeier, line 870: match above suggestion

lmeier (commented):
there's a few things to change here @KennyHuRadar

RadarSDK/Include/Radar.h
KennyHuRadar: Right now the sdk is just passing that int from the user to the server, the server actually does another check to ensure that the radius is more than 100, and sets it to 100 if we do not provide a valid value.

In this case I believe that our SDK can be thought of as the client end stub of a RPC. Since the input verification logic is already handled by the server, we should just pass the value along and pass the warnings back.

However, our SDK does not currently parse the warnings from the server, is this a new pattern we want to introduce?

KennyHuRadar: @lmeier I do agree that passing in -1 may be a little weird for default unlimited behavior. But I also think that growing the number of interface a function of 2x each time we add in a new optional can get ugly in the future. I feel we should bite the bullet and introduce NSNumbers into our interface to support optional bools and ints.

kochis: As it's currently written, `-1` will throw an error if passed. It's either optional, or will support 1 to 10,000.

(we will warn on anything < 100 to preserve backwards compatibility for existing API calls with a small radius)

https://github.com/radarlabs/server/pull/5689/files#diff-6a70e98a5019dec506cc2ac52d75fac8981c3796151166b54e6de676ac5b61cfR365-R369

KennyHuRadar: the sdk won't pass the -1 to the server, it will just ignore that parameter

RadarSDK/Include/Radar.h
```diff
@@ -845,40 +845,34 @@ logConversionWithNotification
...
-                completionHandler:(RadarSearchGeofencesCompletionHandler)completionHandler NS_SWIFT_NAME(searchGeofences(radius:tags:metadata:limit:completionHandler:));
++ (void)searchGeofences:(RadarSearchGeofencesCompletionHandler)completionHandler NS_SWIFT_NAME(searchGeofences(completionHandler:));
 
 /**
  Searches for geofences near a location, sorted by distance.
 
- @param near The location to search.
- @param radius The radius to search, in meters. A number between 100 and 10000.
+ @param near The location to search. Use null to search near the device's current location.
+ @param radius The radius to search, in meters. A number between 100 and 10000. If -1 is entered, the server defaults to using unlimited radius.
```
lmeier, line 859: Let's just point to the version of this w/o the `radius` param if they want the unlimited radius search. `-1` is weird here

RadarSDK/Include/Radar.h
```diff
@@ -845,40 +845,34 @@ logConversionWithNotification
...
  @param tags An array of tags to filter. See https://radar.com/documentation/geofences
  @param metadata A dictionary of metadata to filter. See https://radar.com/documentation/geofences
- @param limit The max number of geofences to return. A number between 1 and 100.
+ @param limit The max number of geofences to return. A number between 1 and 1000. Recommend to be set to 100.
+ @param includeGeometry A boolean indicating whether to include geometry in the response. Recommend to be set to false.
```
lmeier, line 863: let's conform the `includeGeometry` description to the one in the docs and proofread it: https://github.com/radarlabs/documentation/pull/473/files

RadarSDK/Radar.m
```diff
@@ -706,57 +706,61 @@ + (void)searchPlacesNear:(CLLocation *_Nonnull)near
...
-                                                    tags:tags
-                                                metadata:metadata
-                                                   limit:limit
-                                       completionHandler:^(RadarStatus status, NSDictionary *_Nullable res, NSArray<RadarGeofence *> *_Nullable geofences) {
-                                           if (completionHandler) {
-                                               [RadarUtils runOnMainThread:^{
-                                                   completionHandler(status, near, geofences);
-                                               }];
-                                           }
+    if (near == nil){
```
lmeier, line 721: spacing

RadarSDK/Include/Radar.h
KennyHuRadar: The version without the `radius` param just the minimal one, which calls this particular function. If we want to not use `-1`, we will need to either deviate from just having 2 methods or start using nullable NSNumbers.

RadarSDKTests/RadarSDKTests.m
```diff
@@ -823,7 +823,7 @@ - (void)test_Radar_getContext_errorLocation {
         [expectation fulfill];
     }];
 
-    [self waitForExpectationsWithTimeout:30
+    [self waitForExpectationsWithTimeout:60
```
KennyHuRadar, line 826: adding extra time here as the gh action keeps timing out. This is probably made worst by us needing to use a newer  (and probably more demanding) iphone simulator as support for older simulators are being dropped.

RadarSDK/Include/Radar.h
```diff
@@ -845,40 +845,34 @@ logConversionWithNotification
...
 
- @param near The location to search.
- @param radius The radius to search, in meters. A number between 100 and 10000.
+ @param near The location to search. Use null to search near the device's current location.
+ @param radius The radius to search, in meters. A number between 100 and 10000. If -1 is entered, the server defaults to using unlimited radius.
  @param tags An array of tags to filter. See https://radar.com/documentation/geofences
  @param metadata A dictionary of metadata to filter. See https://radar.com/documentation/geofences
- @param limit The max number of geofences to return. A number between 1 and 100.
+ @param limit The max number of geofences to return. A number between 1 and 1000. Defaults to 100.
+ @param includeGeometry Include geofence geometries in the response.. Recommend to be set to false. To retrieve more than 100 results, `includeGeometries` must be set to `false`.
```
lmeier, line 863:
```suggestion
 @param includeGeometry Include geofence geometries in the response. Recommended to be set to false in less you specifically need the geometries. To retrieve more than 100 results, `includeGeometries` must be set to `false`.
```

lmeier (approved):

RadarSDK/Include/Radar.h
```diff
@@ -845,40 +845,34 @@ logConversionWithNotification
...
 
- @param near The location to search.
- @param radius The radius to search, in meters. A number between 100 and 10000.
+ @param near The location to search. Use null to search near the device's current location.
+ @param radius The radius to search, in meters. A number between 100 and 10000. If -1 is entered, the server defaults to using unlimited radius.
  @param tags An array of tags to filter. See https://radar.com/documentation/geofences
  @param metadata A dictionary of metadata to filter. See https://radar.com/documentation/geofences
- @param limit The max number of geofences to return. A number between 1 and 100.
+ @param limit The max number of geofences to return. A number between 1 and 1000. Defaults to 100.
+ @param includeGeometry Include geofence geometries in the response. Recommended to be set to false in less you specifically need the geometries. To retrieve more than 100 results, `includeGeometries` must be set to `false`.
```
lmeier, line 863: Let's get a quick fix in for changing `includeGeometry` instead of `includeGeometries`.

lmeier: I can just do this in mine

KennyHuRadar: i'll fix it in the release pr

PR #329, opened 2024-04-18
Add user _id to createTrip

lmeier:
We actually don't need any change for setting the trip to the one returned as we already have that: https://github.com/radarlabs/radar-sdk-ios/blob/liammeier/fence-1902-send-user-on-starttrip-and-update-tripoptions-with-trip/RadarSDK/Radar.m#L491-L492

### DIFF
--- a/RadarSDK/Include/RadarTripOptions.h
+++ b/RadarSDK/Include/RadarTripOptions.h
@@ -25,7 +25,9 @@ NS_ASSUME_NONNULL_BEGIN
 - (instancetype)initWithExternalId:(NSString *_Nonnull)externalId
             destinationGeofenceTag:(NSString *_Nullable)destinationGeofenceTag
      destinationGeofenceExternalId:(NSString *_Nullable)destinationGeofenceExternalId
-                scheduledArrivalAt:(NSDate *_Nullable)scheduledArrivalAt;
+                scheduledArrivalAt:(NSDate *_Nullable)scheduledArrivalAt
+                          metadata:(NSDictionary *_Nullable)metadata
+                              mode:(RadarRouteMode)mode;
 
 /**
  A stable unique ID for the trip.
--- a/RadarSDK/Radar.m
+++ b/RadarSDK/Radar.m
@@ -489,11 +489,13 @@
     [[RadarAPIClient sharedInstance] createTripWithOptions:tripOptions
                                          completionHandler:^(RadarStatus status, RadarTrip *trip, NSArray<RadarEvent *> *events) {
                                              if (status == RadarStatusSuccess) {
-                                                 RadarTripOptions *newTripOptions = [[RadarTripOptions alloc] initWithExternalId:trip.externalId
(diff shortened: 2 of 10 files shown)


york-wei:
@lmeier it looks like trip options is just set to the one passed in the `startTrip`, I think we need to create a new one with the returned trip's `externalId`

```
+ (void)startTripWithOptions:(RadarTripOptions *)tripOptions
             trackingOptions:(RadarTrackingOptions *)trackingOptions
           completionHandler:(RadarTripCompletionHandler)completionHandler {
    [[RadarLogger sharedInstance] logWithLevel:RadarLogLevelInfo type:RadarLogTypeSDKCall message:@"startTrip()"];
    [[RadarAPIClient sharedInstance] createTripWithOptions:tripOptions
                                         completionHandler:^(RadarStatus status, RadarTrip *trip, NSArray<RadarEvent *> *events) {
                                             if (status == RadarStatusSuccess) {
                                                 [RadarSettings setTripOptions:tripOptions];
```

york-wei (changes requested):
Looks like we need to update the `RadarTrip` class to include `scheduledArrivalAt` and parse it from the JSON payload

RadarSDK/Radar.m
```diff
@@ -518,7 +523,11 @@ + (void)updateTripWithOptions:(RadarTripOptions *)options status:(RadarTripStatu
                                                     status:status
                                          completionHandler:^(RadarStatus status, RadarTrip *trip, NSArray<RadarEvent *> *events) {
                                              if (status == RadarStatusSuccess) {
```
york-wei, line 525: We should add a nil check here for `trip` as well

RadarSDK/Include/RadarTripOptions.h
```diff
@@ -25,7 +25,9 @@ NS_ASSUME_NONNULL_BEGIN
 - (instancetype)initWithExternalId:(NSString *_Nonnull)externalId
             destinationGeofenceTag:(NSString *_Nullable)destinationGeofenceTag
      destinationGeofenceExternalId:(NSString *_Nullable)destinationGeofenceExternalId
-                scheduledArrivalAt:(NSDate *_Nullable)scheduledArrivalAt;
+                scheduledArrivalAt:(NSDate *_Nullable)scheduledArrivalAt
```
KennyHuRadar, line 28: This seems like a breaking change, would we need migration docs?

A follow up question is, do we want to keep on exposing these "init by fields" functions when we already expose the "init by dictionary" functions as we revamp the SDK for one line implementation? The upside would be that we will no longer be collecting deprecated interfaces. The downside would be that initializing from dictionary in native is kinda annoying and not developer friendly. However, one may also claim that with one line implementation all of these options that developers are currently building and loading into the client should be loaded in from the server.

lmeier: Correct, we shouldn't break this here. I'll either introduce another option with the 2 additional params or create a dictionary based initialization.

PR #331, opened 2024-04-26
Update SPM repo

KennyHuRadar:
QA:
We bump the version to 3.9.15-beta.1 and cut a release
Our new dispatch is correctly hitting our github action listener in the SPM repository and creating the correct PR that is automatically merged while a release is automatically created.

image
image
image
image

Using the SPM on an example app indicates that the automated update was successful.
image


.github/workflows/release-sdk.yml
```diff
@@ -43,3 +38,22 @@ jobs:
...
+        run: echo "::set-output name=checksum::$(shasum -a 256 RadarSDK.xcframework.zip | cut -d ' ' -f 1)"
+
+      - name: Repository Dispatch
+        if: ${{ !github.event.release.prerelease }}
+        uses: peter-evans/repository-dispatch@v3
+        with:
+          token: ${{ secrets.SPM_REPO_KEY }}
+          repository: radarlabs/radar-sdk-ios-spm
+          event-type: update-xcframework
+          client-payload: '{"release": "${{ github.event.release.tag_name }}", "checksum": "${{ steps.checksum.outputs.checksum }}"}'
```
lmeier, line 54: Can we auto-create the release on `radar-sdk-ios-spm`? Ideally we don't create a manual step of having to merge a PR in that repo with this.

lmeier (commented):
Are we basing this approach on prior art? I'd love to see a pointer to that approach as I evaluate this PR

.github/workflows/release-sdk.yml
KennyHuRadar: This is mainly based off example snippets from: https://github.com/peter-evans/repository-dispatch

We do something similar in https://github.com/radarlabs/radar-sdk-ios/blob/master/.github/workflows/downstream-version-bump.yml, but if you look at the our github action history that action has been failing for a year now. I personally don't think that automation is useful but that is another topic.

KennyHuRadar: And yes, we can automate merging and creating releases.

RadarSDK/RadarUtils.m
```diff
@@ -45,7 +45,7 @@ + (NSNumber *)timeZoneOffset {
 }
 
 + (NSString *)sdkVersion {
-    return @"3.9.14";
+    return @"3.9.15-beta.3";
```
lmeier, line 48: Let's remove this (and from above as well)

lmeier (approved):
remove the version # change, then this is g2g

lmeier:
Looks like you might need a merge as well

PR #333, opened 2024-04-29
Config fail guards

lmeier:


RadarSDK/RadarLocationManager.m
```diff
@@ -871,6 +871,9 @@ - (void)sendLocation:(CLLocation *)location stopped:(BOOL)stopped source:(RadarL
                                      completionHandler:^(RadarStatus status, NSDictionary *_Nullable res, NSArray<RadarEvent *> *_Nullable events, RadarUser *_Nullable user,
                                                          NSArray<RadarGeofence *> *_Nullable nearbyGeofences, RadarConfig *_Nullable config, NSString *_Nullable token) {
                                          self.sending = NO;
+                                         if (status != RadarStatusSuccess) {
```
KennyHuRadar, line 874:
```suggestion
                                         if (status != RadarStatusSuccess || !config) {
```
Is there a reason why we are not checking here?

lmeier: Ah, indeed can check there.

KennyHuRadar (approved):

PR #334, opened 2024-04-30
Increase timeout for flakey GH tests

KennyHuRadar:
The `getContext` seem to randomly timing out on github action. 
image
image
This is likely an issue with github action because
- This does not occur on circle CI
- This does not occur on local testing
As a sanity check, the SDK endpoint for `getContext` was called on waypoint with no noticeable issue.

### DIFF
--- a/RadarSDKTests/RadarSDKTests.m
+++ b/RadarSDKTests/RadarSDKTests.m
@@ -823,7 +823,7 @@ static NSString *const kPublishableKey = @"prj_test_pk_0000000000000000000000000
         [expectation fulfill];
     }];
 
-    [self waitForExpectationsWithTimeout:60
+    [self waitForExpectationsWithTimeout:120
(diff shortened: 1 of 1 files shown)


RadarSDKTests/RadarSDKTests.m
```diff
@@ -823,7 +823,7 @@ - (void)test_Radar_getContext_errorLocation {
         [expectation fulfill];
     }];
 
-    [self waitForExpectationsWithTimeout:60
+    [self waitForExpectationsWithTimeout:120
```
lmeier, line 826: What's the explanation for why we weren't seeing this before? Or were we?

KennyHuRadar: we were, they come sporadically after we upgraded the github action to simulate an iphone 15 pro

PR #339, opened 2024-05-09
Better abstraction for permissions handling

KennyHuRadar:
Work in progress PR for better location permissions handling with the SDK. We expose a state that can be observed for easier/better UI implementation of location permissions gathering. This PR:
- Defines `RadarLocationPermissionsStatus` to describe the permission status of the device
- Manages permissions gathering and state management with `RadarLocationPermissionsManager`
- Exposes `Radar.requestForegroundLocationPermissions`, `Radar.requestBackgroundLocationPermissions`, `Radar.getLocationPermissionsStatus` and `RadarDelegate.didUpdateLocationPermissionsStatus`


RadarSDK/RadarLocationPremissionsManager.m
```diff
@@ -0,0 +1,167 @@
...
+    self.status = status;
+    [RadarLocationPermissionsStatus store:status];
+    [self sendDelegateUpdate];
+}
+
+- (void)sendDelegateUpdate {
+    if (@available(iOS 14.0, *)) {
+        [[RadarDelegateHolder sharedInstance] didUpdateLocationPermissionsStatus:self.status];
+    }
+}
```
lmeier, line 76: I think you can just pull this into the above, no need to have a function `sendDelegateUpdate`

RadarSDK/RadarLocationPremissionsManager.m
```diff
@@ -0,0 +1,167 @@
...
+                                                                                          inForegroundPopup:self.status.inForegroundPopup
+                                                                          userRejectedBackgroundPermissions: self.status.userRejectedBackgroundPermissions];
+            [self updateStatus:status];
+        }
+
+        // We set a flag that request has been made and start a timer. If we resign active we unset the timer.
+        // When the timer fires and the flag has not been unset, we assume that app never resigned active.
+        // Usually this means that the user has previously rejected the background permissions.
+        dispatch_after(dispatch_time(DISPATCH_TIME_NOW, (int64_t)(0.5 * NSEC_PER_SEC)), dispatch_get_main_queue(), ^{
+            if (self->danglingBackgroundPermissionsRequest) {
```
lmeier, line 97: why aren't we doing `self.danglingBackgroundPermissionsRequest`? We don't use this pointer notation elsewhere

RadarSDK/RadarLocationPremissionsManager.m
```diff
@@ -0,0 +1,167 @@
...
+                                                                                          inForegroundPopup:YES
+                                                                          userRejectedBackgroundPermissions: self.status.userRejectedBackgroundPermissions];
+            [self updateStatus:status];
+        }
+    }
+}
+
+- (void)applicationDidBecomeActive {
+    // we need to handle the case of double updates, we onyl want to update the status if and only if we are coming back from a popup and the status has changed.
+    if (inBackgroundLocationPopUp) {
```
lmeier, line 123: `if (self.inBackgroundLocationPopUp) {` ?

RadarSDK/RadarLocationPremissionsManager.m
```diff
@@ -0,0 +1,167 @@
...
+                                                                                   backgroundPopupAvailable:self.status.backgroundPopupAvailable
+                                                                                          inForegroundPopup:YES
+                                                                          userRejectedBackgroundPermissions: self.status.userRejectedBackgroundPermissions];
+            [self updateStatus:status];
+        }
+    }
+}
+
+- (void)applicationDidBecomeActive {
+    // we need to handle the case of double updates, we onyl want to update the status if and only if we are coming back from a popup and the status has changed.
```
lmeier, line 122: "only"

RadarSDK/RadarLocationPremissionsManager.m
```diff
@@ -0,0 +1,167 @@
...
+        }
+    }
+
+    inBackgroundLocationPopUp = NO;
+}
+
+
+- (void)applicationWillResignActive {
+    if (danglingBackgroundPermissionsRequest) {
+        inBackgroundLocationPopUp = YES;
```
lmeier, line 144: `self.inBackgroundLocationPopUp`?

RadarSDK/RadarUtils.m
```diff
@@ -45,7 +45,7 @@ + (NSNumber *)timeZoneOffset {
 }
 
 + (NSString *)sdkVersion {
-    return @"3.10.1";
+    return @"3.10.2-beta.1";
```
lmeier, line 48: just fyi I've already used this beta version

RadarSDK/RadarLocationPremissionsStatus.m
```diff
@@ -0,0 +1,154 @@
...
+        case ForegroundPermissionsPending:
+            return @"ForegroundPermissionsPending";
+        case BackgroundPermissionsGranted:
+            return @"BackgroundPermissionsGranted";
+        case BackgroundPermissionsRejected:
+            return @"BackgroundPermissionsRejected";
+        case BackgroundPermissionsPending:
+            return @"BackgroundPermissionsPending";
+        case PermissionsRestricted:
+            return @"PermissionsRestricted";
```
lmeier, line 117: do we have the "allow once" case covered? by which?

lmeier (commented):
some comments

Example/Example/AppDelegate.swift
```diff
@@ -248,4 +248,8 @@ class AppDelegate: UIResponder, UIApplicationDelegate, UNUserNotificationCenterD
         
     }
     
+    func didUpdateClientLocationPermissionsStatus(status: RadarLocationPermissionsStatus){
+
```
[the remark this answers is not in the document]
KennyHuRadar: It really depends on how we want to integrate this new feature into our example app. I think building out the whole flow will be overkill and will make up a huge proportion of the example app. Maybe we can just print the state changes?

RadarSDK/RadarLocationPremissionsManager.m
```diff
@@ -0,0 +1,167 @@
...
+        }
+       
+        [[NSNotificationCenter defaultCenter] addObserver:self
+                                                 selector:@selector(applicationWillResignActive)
+                                                     name:UIApplicationWillResignActiveNotification
+                                                   object:nil];
+        [[NSNotificationCenter defaultCenter] addObserver:self
+                                                 selector:@selector(applicationDidBecomeActive)
+                                                     name:UIApplicationDidBecomeActiveNotification
+                                                   object:nil];
```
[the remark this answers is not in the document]
KennyHuRadar: Yes this does not cause conflicts, we should be allowed to register as many listeners as we want and they should not step on each other.

RadarSDK/RadarLocationPremissionsManager.m
KennyHuRadar: I made the decision to not expose `danglingBackgroundPermissionsRequest` as a property on the interface as it is an implementation detail. Hence it only exists as a instance variable and we cannot access it with `.`, only `->`. We can technically drop the `self->` but xcode's linter will be angry at us as this is within a `block`.

RadarSDK/RadarLocationPremissionsManager.m
KennyHuRadar: `self` is implicit in instance methods. it will either be `inBackgroundLocationPopUp ` or `self -> inBackgroundLocationPopUp `

RadarSDK/RadarLocationPremissionsManager.m
KennyHuRadar: see above

RadarSDK/RadarLocationPremissionsStatus.m
KennyHuRadar: So the `allow once` case has 2 "phases". On the first session, it is functionally similar to the `granted foreground` state, with the exception that we cannot launch the request background pop-up (which is handled by us detecting that no pop-up is launched and sending user to recovery state).  On the subsequent session, it is similar to `no permissions granted` and we handle it simply by detecting changes in permissions.

Example/Example/AppDelegate.swift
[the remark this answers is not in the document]
lmeier: Yea, with a link to the full tutorial as well

RadarSDK/RadarLocationPremissionsManager.m
KennyHuRadar: actually changed my mind, apparently its not a pattern we use in the rest of the code, we seem to prefer just using interface and being more "heavy handed" and making stuff properties. Ultimately it's more a stylistic choice, defaulting to existing patterns in this case.

RadarSDK/Include/RadarLocationPermissionStatus.h
```diff
@@ -0,0 +1,55 @@
...
+    NoPermissionsGranted,
+    ForegroundPermissionsGranted,
+    ForegroundPermissionsRejected,
+    ForegroundPermissionsPending,
+    BackgroundPermissionsGranted,
+    BackgroundPermissionsRejected,
+    BackgroundPermissionsPending,
+    PermissionsRestricted,
+    Unknown
+};
```
lmeier, line 30: These should be made singular

RadarSDK/RadarDelegateHolder.m
```diff
@@ -86,4 +87,15 @@ - (void)didUpdateToken:(NSString *)token {
...
 
+- (void)didUpdateLocationPermissionStatus:(RadarLocationPermissionStatus*)status {
+    if (self.delegate) {
+        [self.delegate didUpdateLocationPermissionStatus:status];
+    }
+    
+    NSDictionary *statusDict = [status dictionaryValue];
+    
+    NSString *jsonString = [RadarUtils dictionaryToJson:statusDict];
+    [[RadarLogger sharedInstance] logWithLevel:RadarLogLevelInfo message:[NSString stringWithFormat:@"📍 Radar location permissions updated | status = %@", jsonString]];
```
lmeier, line 98: Can you attach for QA purposes screenshots of these logs in the dashboard?

RadarSDK/RadarLocationPremissionManager.m
```diff
@@ -0,0 +1,190 @@
...
+                self.status = [[RadarLocationPermissionStatus alloc] initWithStatus:self.locationManager.authorizationStatus
+                                                            backgroundPopupAvailable:YES
+                                                                   inForegroundPopup:NO
+                                                   userRejectedBackgroundPermissions:NO];
+            } else {
+                // just a dummy value, this component will not communicate with the rest of the SDK if the version is below 14.0
+                self.status = [[RadarLocationPermissionStatus alloc] initWithStatus: kCLAuthorizationStatusAuthorizedAlways
+                                                            backgroundPopupAvailable:NO
+                                                                   inForegroundPopup:NO
+                                                   userRejectedBackgroundPermissions:NO];
```
lmeier, line 53: Thinking ahead to the analytics, is this the right way of create a dummy value for when it's `unknown`?

lmeier (commented):
few comments

RadarSDK/RadarDelegateHolder.m
KennyHuRadar: image

RadarSDK/RadarLocationPremissionManager.m
KennyHuRadar: I'm just going to make this a nullable property and leave it as null if the iOS version is too low.

Example/Example/AppDelegate.swift
```diff
@@ -248,4 +248,8 @@ class AppDelegate: UIResponder, UIApplicationDelegate, UNUserNotificationCenterD
         
     }
     
+    func didUpdateClientLocationPermissionStatus(status: RadarLocationPermissionStatus){
```
lmeier, line 251: Why `didUpdateClientLocationPermissionStatus` instead of `didUpdateLocationPermissionStatus`? I feel like the latter is more concise. Is there any lack of clarity that removing `client` brings?

KennyHuRadar: Your right, I'll remove it

lmeier (approved):
🚀

PR #342, opened 2024-05-20
add server side logging options

ShiCheng-Lu:
waypoint build

https://expo.dev/accounts/radarlabs/projects/waypoint/builds/b4084b22-731a-407f-8fe2-1039c36ae347


RadarSDK/RadarSettings.m
```diff
@@ -227,6 +228,19 @@ + (void)setFeatureSettings:(RadarFeatureSettings *)featureSettings {
...
+    sdkConfiguration.logLevel = [RadarSettings logLevel];
+    return sdkConfiguration;
+}
+
++ (void)setSDKConfiguration:(RadarSDKConfiguration *)sdkConfiguration {
+    [[RadarLogger sharedInstance] logWithLevel:RadarLogLevelDebug 
+        message:[NSString stringWithFormat:@"Setting SDK Configuration | sdkConfiguration = %@", sdkConfiguration]];
+    if (sdkConfiguration) {
+        [RadarSettings setLogLevel:sdkConfiguration.logLevel];
+    }
```
lmeier, line 242: Hmmmm, I think instead of writing to log level like `[RadarSettings setLogLevel:...]`, maybe we should just write to a `sdkConfiguration` object stored in `NSUserDefaults` (like we do for feature settings), and then in `[RadarSettings logLevel]`, we implement as the second conditional check (`RadarSettings userDebug` should remain the first) where we use the `sdkConfiguration.logLevel` if it exists, otherwise falling back to the local like we do now.

RadarSDK/RadarSdkConfiguration.m
```diff
@@ -0,0 +1,43 @@
...
+- (instancetype)initWithLogLevel:(RadarLogLevel)logLevel {
+    if (self = [super init]) {
+        _logLevel = logLevel;
+    }
+    return self;
+}
+
++ (RadarSDKConfiguration *_Nullable)sdkConfigurationFromDictionary:(NSDictionary *)dict {
+    if (!dict) {
+        return NULL;
```
lmeier, line 23:
```suggestion
        return nil;
```

lmeier: (It'll be helpful to check the codebase for how to implement certain patterns when you're writing SDK code — I know objective C is a bit unfamiliar, lean on the prior art)

RadarSDK/RadarSdkConfiguration.m
lmeier, line 1: Given that this is your first time writing significant SDK code, I think it'll be good to get some basic tests of these things in our testing suite.

RadarSDK/RadarLog.m
```diff
@@ -43,6 +43,22 @@ + (NSString *)stringForLogLevel:(RadarLogLevel)level {
...
+        return RadarLogLevelNone;
+    } else if ([string isEqualToString:@"error"]) {
+        return RadarLogLevelError;
+    } else if ([string isEqualToString:@"warning"]) {
+        return RadarLogLevelWarning;
+    } else if ([string isEqualToString:@"info"]) {
+        return RadarLogLevelInfo;
+    } else if ([string isEqualToString:@"debug"]) {
+        return RadarLogLevelDebug;
+    } else { // default to info
```
lmeier, line 57: don't need this comment

RadarSDK/RadarLogger.m
```diff
@@ -49,7 +49,12 @@ - (void)logWithLevel:(RadarLogLevel)level type:(RadarLogType)type message:(NSStr
 }
 
 - (void)logWithLevel:(RadarLogLevel)level type:(RadarLogType)type message:(NSString *)message includeDate:(BOOL)includeDate includeBattery:(BOOL)includeBattery append:(BOOL)append{
-     NSString *dateString = [self.dateFormatter stringFromDate:[NSDate date]];
+    RadarLogLevel logLevel = [RadarSettings logLevel];
+    if (level > logLevel) {
+        return;
+    }
```
lmeier, line 55: Hmmm, do we need to think through other effects of this?

RadarSDK/RadarSdkConfiguration.h
```diff
@@ -0,0 +1,44 @@
...
+/**
+ Represents server-side sdk configuration.
+ 
+ @see https://radar.com/documentation/sdk/ios
+ */
+@interface RadarSDKConfiguration : NSObject
+
+/**
+ 
+ */
```
lmeier, line 22: Think an empty comment snuck in here

RadarSDK/RadarSdkConfiguration.m
ShiCheng-Lu: this is because I'm professional C programmer, and NULL didn't raising any errors so... (it shall be fixed)

RadarSDK/RadarLogger.m
ShiCheng-Lu: this check used to be performed inside dispatch_async after Radar sendLog, which means the logs are still sent to server. moving it up here prevents logs with higher value (less important) than the set level from being sent. This matches android logging behaviour.

.gitignore
```diff
@@ -11,4 +11,4 @@ build/
 docs/undocumented.json
 *.xcarchive
 IntegrationExamples
-
+.build/
```
lmeier, line 14: remove

RadarSDK/RadarSettings.m
```diff
@@ -226,6 +230,32 @@ + (void)setFeatureSettings:(RadarFeatureSettings *)featureSettings {
...
++ (void)updateSdkConfigurationFromServer {
+    [[RadarAPIClient sharedInstance] getConfigForUsage:@"sdkConfigUpdate"
+                                              verified:NO
+                                     completionHandler:^(RadarStatus status, RadarConfig *config) {
+                                         if (status != RadarStatusSuccess || !config) {
+                                            return;
+                                         }
+                                         [RadarSettings setSdkConfiguration:config.meta.sdkConfiguration];
+                                     }];
+}
```
lmeier, line 242: We don't call the API client anywhere else in `RadarSettings` — I think we can find a better home for this call.

I'd also say that the name `updateSdkConfigurationFromServer` is a clue that isn't correctly placed (I don't think we a want function named `update_____FromServer`)

ShiCheng-Lu:
I will remove the key for testing if everything else looks good

RadarSDK/RadarSdkConfiguration.h
```diff
@@ -0,0 +1,43 @@
...
+ @see https://radar.com/documentation/sdk/ios
+ */
+@interface RadarSdkConfiguration : NSObject
+
+@property (nonatomic, assign) RadarLogLevel logLevel;
+
+/**
+ Initializes a new RadarSdkConfiguration object with given value.
+ */
+- (instancetype)initWithLogLevel:(RadarLogLevel)logLevel;
```
KennyHuRadar, line 25: Do we want to call it this? It seems to me that the log level is unlikely to remain the most important field in this obj when we are done.

RadarSDK/RadarSettings.m
```diff
@@ -226,6 +229,22 @@ + (void)setFeatureSettings:(RadarFeatureSettings *)featureSettings {
...
+    } 
+    return sdkConfigurationDict;
+}
+
++ (void)setSdkConfiguration:(RadarSdkConfiguration *)sdkConfiguration {
+    [[RadarLogger sharedInstance] logWithLevel:RadarLogLevelDebug
+        message:[NSString stringWithFormat:@"Setting SDK configuration | sdkConfiguration = %@",
+                            [RadarUtils dictionaryToJson:[sdkConfiguration dictionaryValue]]]];
+    if (sdkConfiguration) {
+        [[NSUserDefaults standardUserDefaults] setInteger:(int)sdkConfiguration.logLevel forKey:kLogLevel];
```
KennyHuRadar, line 245: why are we storing just an int here? Why not the whole `RadarSdkConfiguration` obj

RadarSDK/RadarSettings.m
```diff
@@ -240,8 +259,18 @@ + (RadarLogLevel)logLevel {
 }
 
 + (void)setLogLevel:(RadarLogLevel)level {
-    NSInteger logLevelInteger = (int)level;
-    [[NSUserDefaults standardUserDefaults] setInteger:logLevelInteger forKey:kLogLevel];
+    NSMutableDictionary *sdkConfiguration = [[RadarSettings clientSdkConfiguration] mutableCopy];
```
KennyHuRadar, line 262: I do not think it is appropriate to sync the log level here. I believe that the responsibility of  `RadarSettings` should be a simple wrapper around NSUserdefault, like our own key value store. Such syncing logic probably belongs to `RadarLogger` instead. (granted this will result in more lines of code changed)

ShiCheng-Lu: how this works is we have a client sdk config (what the client wants), then a server sdk config (what the server wants). and setLogLevel sets what the client wants, then that gets sent to the server and server can override a client config and send the config that will be used back down. 

RadarLogger is not going to be syncing the config every single time it gets the log level. The logic here is to minimize sync config calls if the target log level is already set. 

I think its fine to keep this here as this is the settings managing logic.(RadarSettings updateSessionId also does some logic to flush replays, getLogLevel does some logic to determine log level based on if the user is a debug user.

PR #349, opened 2024-06-14
Shicheng/fence 1995 start tracking on initialize

ShiCheng-Lu:
tested with example app, verified via logs startTracking() is called when the startTrackingOnInitialize variable is set to true. Also only called when tracking is not started.

Will remove the testing stuff in AppDelegate if everything else looks good.

### DIFF
--- a/Example/Example/AppDelegate.swift
+++ b/Example/Example/AppDelegate.swift
@@ -26,11 +26,9 @@ class AppDelegate: UIResponder, UIApplicationDelegate, UNUserNotificationCenterD
         Radar.stopTracking()
         
         // Replace with a valid test publishable key
-        Radar.initialize(publishableKey: "prj_test_pk_d9020be31f7cd2357d59a728008f469b1395ee4e")
+        Radar.initialize(publishableKey: "prj_test_pk_0000000000000000000000000000000000000000")
         Radar.setDelegate(self)
         Radar.setVerifiedDelegate(self)
-        
-        return true
 
         if UIApplication.shared.applicationState != .background {
             Radar.getLocation { (status, location, stopped) in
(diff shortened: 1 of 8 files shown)


Example/Example/AppDelegate.swift
```diff
@@ -21,10 +21,16 @@ class AppDelegate: UIResponder, UIApplicationDelegate, UNUserNotificationCenterD
         locationManager.delegate = self
         self.requestLocationPermissions()
 
+        UserDefaults.standard.set("https://api-shicheng.radar-staging.com", forKey:"radar-host")
+        
+        Radar.stopTracking()
+        
         // Replace with a valid test publishable key
-        Radar.initialize(publishableKey: "prj_test_pk_0000000000000000000000000000000000000000")
+        Radar.initialize(publishableKey: "prj_test_pk_d9020be31f7cd2357d59a728008f469b1395ee4e")
```
KennyHuRadar, line 29: lets not commit this, also you kinda leaked your pub key. Let's not have this in the PR and you may want to rotate your staging key.

RadarSDK/RadarSettings.m
```diff
@@ -243,9 +247,15 @@ + (void)setSdkConfiguration:(RadarSdkConfiguration *)sdkConfiguration {
                             [RadarUtils dictionaryToJson:[sdkConfiguration dictionaryValue]]]];
     if (sdkConfiguration) {
         [[NSUserDefaults standardUserDefaults] setInteger:(int)sdkConfiguration.logLevel forKey:kLogLevel];
+        [[NSUserDefaults standardUserDefaults] setObject:[sdkConfiguration dictionaryValue] forKey:kSdkConfiguration];
     }
 }
 
++ (RadarSdkConfiguration *_Nullable)sdkConfiguration {
+    NSDictionary *sdkConfigurationDict = [[NSUserDefaults standardUserDefaults] dictionaryForKey:kSdkConfiguration];
```
KennyHuRadar, line 255: Is this a safe operation?

ShiCheng-Lu: will return nil if the value doesn't exist, then RadarSdkConfiguration from dictionary also returns nil

PR #352, opened 2024-06-20
Shicheng/fence 1979 track once on initialize and app resume

ShiCheng-Lu:

### DIFF
--- a/RadarSDK/Radar.m
+++ b/RadarSDK/Radar.m
@@ -79,6 +79,9 @@
                                          if (sdkConfiguration.startTrackingOnInitialize && ![RadarSettings tracking]) {
                                             [Radar startTrackingWithOptions:[RadarSettings trackingOptions]];
                                          }
+                                         if (sdkConfiguration.trackOnceOnInitialize) {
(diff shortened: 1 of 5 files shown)


RadarSDK/Radar.m
```diff
@@ -79,6 +79,9 @@ + (void)initializeWithPublishableKey:(NSString *)publishableKey {
                                          if (sdkConfiguration.startTrackingOnInitialize && ![RadarSettings tracking]) {
                                             [Radar startTrackingWithOptions:[RadarSettings trackingOptions]];
                                          }
+                                         if (sdkConfiguration.trackOnceOnInitialize) {
```
KennyHuRadar, line 82: how does this perform in practice? iOS does not allow us to request location multiple times at once (one of the location requests may be dropped). I think our implementation is resilient to that but lets make sure.

ShiCheng-Lu: for now, we keep this. there is this coalescing track feature https://github.com/radarlabs/radar-sdk-ios/pull/318 that may be used in the future.

PR #354, opened 2024-06-24
Fix parsing of geofence coordinates

KennyHuRadar:
I was under the impression that this issue was fixed and merged previously but apparently it is not. 

@lmeier the P1 fix that you made in waypoint addresses this issue in iOS but breaks android. This PR is the more appropriate  fix for the issue.


ShiCheng-Lu (approved):
lgtm

PR #355, opened 2024-06-25
[WIP] Location permission improvements

KennyHuRadar:
- Includes enumerations and logic to handle approximate locations.
- changes internal naming.
- Fixes typos.


KennyHuRadar:
I had to increase timeouts for some of the test to pass the github CI :(

RadarSDK/Include/RadarLocationPermissionStatus.h
```diff
@@ -20,9 +20,11 @@
 typedef NS_ENUM(NSInteger, RadarLocationPermissionState) {
     NoPermissionGranted,
     ForegroundPermissionGranted,
+    ApproximateForegroundPermissionGranted,
     ForegroundPermissionRejected,
     ForegroundPermissionPending,
     BackgroundPermissionGranted,
+    ApproximateBackgroundPermissionGranted,
```
ShiCheng-Lu, line 27: has these been added by apple? if not, then we should leave them out even if android has them.

RadarSDK/Include/RadarLocationPermissionStatus.h
```diff
@@ -18,33 +18,39 @@
...
+    ForegroundAuthorized,
+    ForegroundFullAccuracyDenied,
+    ForegroundAuthorizationDenied,
+    ForegroundAuthorizationRequestInProgress,
+    BackgroundAuthorized,
+    BackgroundFullAccuracyDenied,
+    BackgroundAuthorizationDenied,
+    BackgroundAuthorizationRequestInProgress,
+    AuthorizationRestricted,
     Unknown
```
KennyHuRadar, line 31: @lmeier here is my proposed renaming scheme to make our vocabulary more similar to that of Apples.
For reference here is Apple's docs for the new fields: https://developer.apple.com/documentation/corelocation/clservicesessiondiagnostic?changes=latest_minor&language=objc

KennyHuRadar:
@lmeier As per discussed offline:

- We prefer to keep our `foreground` and `background` prefix.
- We will be going with the more colloquial term `permission` vs apple's term of `authorization`
- We will prefer `precise` and maybe `coarse` vs Apple's term `fullAccuracy` 
- We will adopt Apple's phrasing `inProgress` and  `request`
- We will use Apple's term `denied`, which will introduce its counterpart `granted`

lmeier:
@KennyHuRadar looks like you haven't yet pushed here.

lmeier:
Closing for now

PR #357, opened 2024-06-28
Initialize changes for 1 line imp

ShiCheng-Lu:
- allows for `trackOnceOnAppOpen` to be set server side
- allows for `startTrackingOnInitialize`
- allows for log level to be set server side

waypoint build:
https://expo.dev/accounts/radarlabs/projects/waypoint/builds/c06980ba-46fa-45b2-80b8-6d3c316f1f78


RadarSDK/Radar.m
```diff
@@ -62,22 +71,36 @@ + (void)initializeWithPublishableKey:(NSString *)publishableKey {
         [RadarSettings updateSessionId];
     }
 
-    [RadarLocationPermissionManager sharedInstance];
```
ShiCheng-Lu, line 65: is this construction?

RadarSDK.podspec
```diff
@@ -1,6 +1,6 @@
 Pod::Spec.new do |s|
   s.name                  = 'RadarSDK'
-  s.version               = '3.13.5'
+  s.version               = '3.14.0'
```
KennyHuRadar, line 3: @lmeier I recall us wanting to make this a major version update?

RadarSDK/Radar.m
```diff
@@ -65,19 +74,36 @@ + (void)initializeWithPublishableKey:(NSString *)publishableKey {
...
+                                         if (sdkConfiguration.trackOnceOnInitialize) {
+                                            [Radar trackOnceWithCompletionHandler:nil];
                                          }
-                                         [[RadarLocationManager sharedInstance] updateTrackingFromMeta:config.meta];
-                                         [RadarSettings setFeatureSettings:config.meta.featureSettings];
+
                                          [self flushLogs];
                                      }];
-    
+    if (options.userId != nil) {
```
KennyHuRadar, line 97: Is there a reason we are setting the user ID and meta data after the initial sync? Do we want to send that information up to the server during initialize right away?

RadarSDK/RadarInitializeOptions.m
```diff
@@ -0,0 +1,13 @@
+//
```
KennyHuRadar, line 1: we need some to dict and from dict helper methods for x-platform compatibility.

RadarSDK/RadarSettings.m
```diff
@@ -226,6 +234,28 @@ + (void)setFeatureSettings:(RadarFeatureSettings *)featureSettings {
     }
 }
 
++ (NSDictionary *)clientSdkConfiguration {
```
KennyHuRadar, line 237: Can you add a comment on why isn't there a setter for this? Will we need one in the future?

RadarSDKTests/RadarSDKTests.m
```diff
@@ -1489,5 +1489,47 @@ - (void)test_RadarLogBuffer_purge {
     [[RadarLogBuffer sharedInstance]clearBuffer];
 }
 
+- (void)test_RadarSdkConfiguration {
```
KennyHuRadar, line 1492: I feel that having a test using initialize might be appropriate here to give better guarantees about e2e behaviors.  Granted we need to mock a bunch of stuff...

RadarSDK/Radar.m
ShiCheng-Lu: I don't think those get sent during initialize, normally in code they are set after the init call, I just placed them at the end for convenience, we can move them up in case they are being used.

RadarSDK/RadarInitializeOptions.m
ShiCheng-Lu: I shall add them 👍

RadarSDK/RadarSettings.m
ShiCheng-Lu: This is what the client want to sdkConfiguration to be, currently only logLevel is settable within clientSdkConfiguration (it is also the only field in it), which is set via setLogLevel. We might need one in the future if there is other things to set.

RadarSDK/RadarInitializeOptions.m
KennyHuRadar: hmm technically the native side of the x platform sdk can just use the native init method... maybe we can get away with not having any? Thoughts @lmeier

RadarSDK/Radar.m
KennyHuRadar: Hmm its my impression that one of the intention of one line implementation is to automatically sync meta-data both ways during initialization time to reduce the lines of code that is required to integrate our SDK. To me it feels like an omission to not sync the metadata and userid provided directly to the client to the server.

ShiCheng-Lu: server doesn't not need to know about metadata/userId, it is passed up per track call.

RadarSDKTests/RadarSDKTests.m
ShiCheng-Lu: I added userId and metadata check for initialize, It's difficult to verify if trackOnce was called on init, I feel like the api.getConfigForUsage is good enough to cover the same call from init (verified if the variable is set).

lmeier:
Let's fix the failing tests here @ShiCheng-Lu:

```
RadarSDKTests
  test_RadarSdkConfiguration, ((config.meta.sdkConfiguration.trackOnceOnResume) equal to (true)) failed: ("{length = 1, bytes = 0x00}") is not equal to ("1")
  /Users/distiller/project/RadarSDKTests/RadarSDKTests.m:1526
  ```
        
        XCTAssertEqual(config.meta.sdkConfiguration.trackOnceOnResume, true);
        XCTAssertEqual(config.meta.sdkConfiguration.trackOnceOnInitialize, true);
  ```

  test_RadarSdkConfiguration, ((savedSdkConfiguration.trackOnceOnResume) equal to (YES)) failed: ("{length = 1, bytes = 0x00}") is not equal to ("{length = 1, bytes = 0x01}")
  /Users/distiller/project/RadarSDKTests/RadarSDKTests.m:1545
  ```
    RadarSdkConfiguration *savedSdkConfiguration = [RadarSettings sdkConfiguration];
    XCTAssertEqual(savedSdkConfiguration.trackOnceOnResume, YES);
    XCTAssertEqual(savedSdkConfiguration.trackOnceOnInitialize, YES);
  ```
  ```

File: RadarSDK.podspec (hunk 2 above)
Comment (lmeier) [reply RadarSDK.podspec:3]: Given that we're keeping existing behavior, we can keep this a minor. I expect that sometime this fall we'll publicize all these changes a bit more and roll into a `4.0.0`

File: RadarSDK/RadarInitializeOptions.m (hunk 4 above)
Comment (lmeier) [reply RadarSDK/RadarInitializeOptions.m:1]: Hmmm, first instinct is that then the x-platforms are essentially rewriting the `toDict` themselves in the init -- probably nicer to follow the normal abstraction pattern we have established an indeed have both `fromDict` and `toDict` and use those

**Review by lmeier — COMMENTED** [review]
Let's get a Waypoint build with this PR into our hands ASAP — getting some reps in on this will help ensure that impl is correct

File: Example/Example/AppDelegate.swift [hunk 7]
```diff
@@ -20,9 +20,14 @@ class AppDelegate: UIResponder, UIApplicationDelegate, UNUserNotificationCenterD
...
-
+        
+        Radar.stopTracking()
+        
         // Replace with a valid test publishable key
-        Radar.initialize(publishableKey: "prj_test_pk_0000000000000000000000000000000000000000")
+        Radar.initialize(publishableKey: "prj_test_pk_0000000000000000000000000000000000000000", options: RadarInitializeOptions.fromDictionary([
+            "userId": "testUserId",
+            "metadata": [ "foo": "bar" ]
+        ]))
```
lmeier, line 30: @ShiCheng-Lu can we just make the second argument a dictionary instead of `RadarInitializeOptions` object?

RadarSDK/Radar.m
```diff
@@ -44,6 +46,13 @@ + (id)sharedInstance {
 }
 
 + (void)initializeWithPublishableKey:(NSString *)publishableKey {
+    RadarInitializeOptions *options = [[RadarInitializeOptions alloc] init];
+    [Radar initializeWithPublishableKey:publishableKey options:options];
+}
+
+
++ (void)initializeWithPublishableKey:(NSString *)publishableKey
+                             options:(RadarInitializeOptions *)options {
```
lmeier, line 55: I think it's a much nicer interface to just accept a dictionary here instead of `RadarInitializeOptions`

ShiCheng-Lu: could also add a initialize function that takes a function to modify RadarInitializeOptions like how sentry initialization works https://github.com/getsentry/sentry-cocoa/blob/main/Sources/Sentry/SentrySDK.m.

RadarSDK/RadarInitializeOptions.m
ShiCheng-Lu: we have from/to dict now

RadarSDK/Radar.m
lmeier: yes indeed — we'll make a call on this after the Thursday geofencing review

lmeier:
The last thing on my mind here: Nick had the feedback that we have `RadarFeatureSetting`, `RadarSDKConfiguration`, and `RemoteTrackingOptions`.

We should unify at least `RadarFeatureSetting` and `RadarSDKConfiguration` before we ship this. (and maybe the tracking options as well — let's discuss this).

RadarSDK/RadarSettings.m
lmeier, line 1: there remains the `static NSString *const kFeatureSettings = @"radar-featureSettings";` line in this file

RadarSDK/RadarLocationManager.m
```diff
@@ -875,7 +875,6 @@ - (void)sendLocation:(CLLocation *)location stopped:(BOOL)stopped source:(RadarL
                 self.sending = NO;
                 
                 [self updateTrackingFromMeta:config.meta];
-                [RadarSettings setFeatureSettings:config.meta.featureSettings];
```
lmeier, line 878: I think we want to update the sdk configuration here:
``` 
[RadarSettings setSdkConfiguration:config.meta.sdkConfiguration];
```

RadarSDK/RadarLocationManager.m
```diff
@@ -961,7 +960,6 @@ - (void)sendLocation:(CLLocation *)location stopped:(BOOL)stopped source:(RadarL
                                              }
 
                                              [self updateTrackingFromMeta:config.meta];
-                                             [RadarSettings setFeatureSettings:config.meta.featureSettings];
```
lmeier, line 964: same as above

lmeier (commented):
once you make the changes, we'll want to cut a waypoint for testing. almost there!

RadarSDK/RadarLocationManager.m
ShiCheng-Lu: this is not the config endpoint, sdk config is not returned.

lmeier (approved):
🚀

PR #358, opened 2024-06-28
 pre commit hook to prevent commiting key

ShiCheng-Lu:
verified hook working, committing any key that is not zeros will abort the commit and output a message with the detected key


KennyHuRadar (approved):
nice

PR #360, opened 2024-07-01
add buttons to example

ShiCheng-Lu:
Screenshot 2024-07-01 at 4 15 04 PM

### DIFF
--- a/Example/Example/AppDelegate.swift
+++ b/Example/Example/AppDelegate.swift
@@ -10,22 +10,68 @@ import UserNotifications
 import RadarSDK
 
 @UIApplicationMain
-class AppDelegate: UIResponder, UIApplicationDelegate, UNUserNotificationCenterDelegate, CLLocationManagerDelegate, RadarDelegate, RadarVerifiedDelegate {
+class AppDelegate: UIResponder, UIApplicationDelegate, UIWindowSceneDelegate, UNUserNotificationCenterDelegate, CLLocationManagerDelegate, RadarDelegate, RadarVerifiedDelegate {
 
     let locationManager = CLLocationManager()
+    var window: UIWindow? // required for UIWindowSceneDelegate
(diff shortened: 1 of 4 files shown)


Example/Example/AppDelegate.swift
```diff
@@ -10,22 +10,68 @@ import UserNotifications
...
+    }
+    
+    func createButton(text: String, function: @escaping () -> Void) {
+        guard let scrollView = self.scrollView else { return }
+        
+        let buttonHeight = 30
+        scrollView.contentSize.height += CGFloat(buttonHeight)
+        
+        let button = UIButton(frame: CGRect(x: 0, y: demoFunctions.count * buttonHeight, width: Int(scrollView.frame.width), height: buttonHeight),
+                              primaryAction:UIAction(handler:{ _ in
```
KennyHuRadar, line 43: is this formatting correct or just showing up funky on github?

Example/Example/AppDelegate.swift
```diff
@@ -10,22 +10,68 @@ import UserNotifications
...
+        
         // Replace with a valid test publishable key
         Radar.initialize(publishableKey: "prj_test_pk_0000000000000000000000000000000000000000")
         Radar.setDelegate(self)
         Radar.setVerifiedDelegate(self)
+        
+        return true
+    }
+    
+    func createButton(text: String, function: @escaping () -> Void) {
```
KennyHuRadar, line 36: granted this is for parity with android, but it kinda seems out of place within the context of iOS native development. We should just name it `demoButton` or something imo.

Example/Example/AppDelegate.swift
ShiCheng-Lu: its a long line, I'll take out frame to its own variable

lmeier (approved):

PR #362, opened 2024-07-08
start tracking when trip starts if not already tracking

ShiCheng-Lu:


lmeier (approved):
@ShiCheng-Lu lgtm, let's merge and ship in the next release

PR #367, opened 2024-07-15
Remove location permissions for now

KennyHuRadar:


lmeier (commented):
missing version change in the podspec

lmeier (approved):

PR #371, opened 2024-07-31
validate address should allow addresslabel in place of street + number, fix timezone json conversion issues. 

ShiCheng-Lu:
https://linear.app/radarlabs/issue/FENCE-2045/validate-address-should-allow-addresslabel-in-addition-to-street

also:

https://linear.app/radarlabs/issue/FENCE-2044/some-dictionaryvalues-does-not-convert-nsdate-to-string-properly

all tested with example app with each producing the correct outputs when converted to JSON.


RadarSDK/RadarTimeZone.m
```diff
@@ -56,7 +67,7 @@ - (NSDictionary *)dictionaryValue {
     dict[@"_id"] = self._id;
     dict[@"name"] = self.name;
     dict[@"code"] = self.code;
-    dict[@"currentTime"] = self.currentTime;
+    dict[@"currentTime"] = [timezoneDateFormatter() stringFromDate:self.currentTime];
```
ShiCheng-Lu, line 70: the current time string from server is a different format (no millisecond) so we need a different parser than RadarUtils.isoDateFormatter

RadarSDK/RadarTripOptions.m
```diff
@@ -89,7 +89,7 @@ - (NSDictionary *)dictionaryValue {
     dict[kDestinationGeofenceTag] = self.destinationGeofenceTag;
     dict[kDestinationGeofenceExternalId] = self.destinationGeofenceExternalId;
     dict[kMode] = [Radar stringForMode:self.mode];
-    dict[kScheduledArrivalAt] = self.scheduledArrivalAt;
```
ShiCheng-Lu, line 92: NSDate object does not get converted, this field was null, format it to string since all our other date formats are converted to string with JSON

RadarSDK/RadarTimeZone.m
```diff
@@ -8,6 +8,16 @@
 #import "RadarTimeZone.h"
 #import "RadarUtils.h"
 
+NSDateFormatter *_timezoneDateFormatter = nil;
```
KennyHuRadar, line 11: Probably there no action items for with respect to this PR, but we really should consider having a more "central" place where we define our time parsers. There are quite a few definitions littered around the codebase. Just a call out to keep in mind for the future.

KennyHuRadar:
Looks mostly good but lets resolve that merge conflict.

KennyHuRadar (approved):
looks good, wait for @liam to take a pass too?

PR #375, opened 2024-08-05
patch to track on permissions granted

KennyHuRadar:


RadarSDK/RadarLocationManager.m
```diff
@@ -50,6 +50,8 @@ @interface RadarLocationManager ()
  */
 @property (nonnull, strong, nonatomic) NSMutableArray<RadarLocationCompletionHandler> *completionHandlers;
 
+@property (nonatomic) BOOL firstPermissionCheck;
```
lmeier, line 53: We need in init` _firstPermissionCheck = YES;`.

RadarSDK/RadarLocationManager.m
```diff
@@ -1127,8 +1129,17 @@ - (void)locationManager:(CLLocationManager *)manager didFailWithError:(NSError *
 }
 
 - (void)locationManager:(CLLocationManager *)manager didChangeAuthorizationStatus:(CLAuthorizationStatus)status {
+    if (_firstPermissionCheck) {
+        _firstPermissionCheck = NO;
```
lmeier, line 1133:
```suggestion
    if (firstPermissionCheck) {
        firstPermissionCheck = NO;
```

KennyHuRadar: We need the `_` as this is an internal flag not defined in the header.

lmeier (approved):

PR #376, opened 2024-08-06
use NSURLComponent to build query strings

ShiCheng-Lu:
properly escape each query param instead of escaping the entire url string at the end. uses NSURLComponent and NSQueryItem.

Tested with example app and made sure all the requests are returning the same as before.


RadarSDK/RadarAPIHelper.h
```diff
@@ -22,6 +22,7 @@ typedef void (^_Nullable RadarAPICompletionHandler)(RadarStatus status, NSDictio
                     sleep:(BOOL)sleep
                logPayload:(BOOL)logPayload
           extendedTimeout:(BOOL)extendedTimeout
+                 verified:(BOOL)verified
```
ShiCheng-Lu, line 25: moved verified into API helper to match android.

RadarSDK/RadarAPIClient.m
```diff
@@ -86,27 +86,24 @@ - (void)getConfigForUsage:(NSString *_Nullable)usage verified:(BOOL)verified com
...
-    [queryString appendFormat:@"&id=%@", [RadarSettings _id]];
-    NSString *locationAuthorization = [RadarUtils locationAuthorization];
-    if (locationAuthorization) {
-        [queryString appendFormat:@"&locationAuthorization=%@", locationAuthorization];
-    }
-    NSString *locationAccuracyAuthorization = [RadarUtils locationAccuracyAuthorization];
-    if (locationAccuracyAuthorization) {
-        [queryString appendFormat:@"&locationAccuracyAuthorization=%@", locationAccuracyAuthorization];
-    }
+    NSMutableArray *queryString = [[NSMutableArray alloc] init];
```
ShiCheng-Lu, line 89: NSArray<NSQueryItem>

PR #380, opened 2024-08-09
return public key error if no key detected in get config

ShiCheng-Lu:
Verified the trackVerified call now returns public key error if Radar is not initialized (public key = nil)


ShiCheng-Lu:
closed by https://github.com/radarlabs/radar-sdk-ios/pull/393

PR #381, opened 2024-08-12
Expose token.failureReasons and token._id, add setExpectedJurisdiction(), fix hanging getConfig()

nickpatrick:


RadarSDK/RadarAPIClient.m
```diff
@@ -83,7 +83,7 @@ + (NSDictionary *)headersWithPublishableKey:(NSString *)publishableKey {
 - (void)getConfigForUsage:(NSString *_Nullable)usage verified:(BOOL)verified completionHandler:(RadarConfigAPICompletionHandler _Nonnull)completionHandler {
     NSString *publishableKey = [RadarSettings publishableKey];
     if (!publishableKey) {
-        return;
+        return completionHandler(RadarStatusErrorPublishableKey, nil);
```
nickpatrick, line 86: Fixes hanging callback if `trackVerified()` is called with missing pk

RadarSDK/RadarVerificationManager.m
```diff
@@ -194,6 +198,11 @@ - (void)startTrackingVerifiedWithInterval:(NSTimeInterval)interval beacons:(BOOL
                 minInterval = minInterval - 10;
             }
             
+            // min interval is 10 seconds
```
nickpatrick, line 201: Don't allow `interval` faster than 10 seconds

RadarSDK/Include/Radar.h
```diff
@@ -492,6 +492,14 @@ typedef void (^_Nonnull RadarLogConversionCompletionHandler)(RadarStatus status,
  */
 + (void)getVerifiedLocationToken:(RadarTrackVerifiedCompletionHandler _Nullable)completionHandler NS_SWIFT_NAME(getVerifiedLocationToken(completionHandler:));
 
+/**
+ Optionally sets the user's expected country and state for jurisdiction checks.
+ 
+ @param countryCode The user's expected two-letter country code.
+ @param stateCode The user's expected two-letter state code.
+ */
++ (void)setExpectedJurisdictionWithCountryCode:(NSString *_Nullable)countryCode stateCode:(NSString *_Nullable)stateCode NS_SWIFT_NAME(setExpectedJurisdiction(countryCode:stateCode:));
```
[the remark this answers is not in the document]
nickpatrick: If you scroll all the way to the right, it will be `setExpectedJurisdiction()` in Swift. `setExpectedJurisdictionWithCountryCode:stateCode:` is just an Obj-C convenion

RadarSDK/RadarVerificationManager.m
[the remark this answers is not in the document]
nickpatrick: I think we can finesse this later if needed (will be rare to have interval < 60 seconds, mostly just making sure someone can't shoot themselves in the foot with a 0 interval)

RadarSDK/RadarVerificationManager.m
```diff
@@ -194,6 +198,11 @@ - (void)startTrackingVerifiedWithInterval:(NSTimeInterval)interval beacons:(BOOL
...
+            if (minInterval < 10) {
+                minInterval = 10;
+            }
+
```
lmeier, line 205: Not really an issue with the new code here, but `startTrackingVerifiedWithInterval` initiates a `trackVerified` immediately (unlike the other `startTracking`) — we don't have a check on `self.started` or `sending`. We maybe want one to prevent simultaneous `trackVerified`s.

At the beginning of `startTrackingVerifiedWithInterval`:
```
if (self.started) {
    [[RadarLogger sharedInstance] logWithLevel:RadarLogLevelDebug message:@"Already tracking verified"];
    return
}
```

RadarSDK/RadarVerificationManager.m
```diff
@@ -288,6 +297,11 @@ - (void)getVerifiedLocationTokenWithCompletionHandler:(RadarTrackVerifiedComplet
     [self trackVerifiedWithBeacons:self.lastTokenBeacons completionHandler:completionHandler];
 }
 
+- (void)setExpectedJurisdictionWithCountryCode:(NSString *)countryCode stateCode:(NSString *)stateCode {
+    self.expectedCountryCode = countryCode;
+    self.expectedStateCode = stateCode;
+}
```
lmeier, line 303: We're not persisting the expected jurisdiction, which I kinda expected we would. Could be an issue if `trackVerified` or `startTrackingVerified` is ever called before `setExpectedJurisdiction`.

Also would we want `stateCode` (and maybe even `countryCode` to be a nullable array? I imagine you might want to set an expected `countryCode` without a specific expected `stateCode`, and I also imagine you might want to expect multiple states.

RadarSDK/Include/RadarVerifiedLocationToken.h
```diff
@@ -46,6 +46,16 @@
...
 
+/**
+ An array of failure reasons for jurisdiction and fraud detection checks.
+ */
+@property (nullable, copy, nonatomic, readonly) NSArray<NSString *> *failureReasons;
+
+/**
+ The Radar ID of the location check.
+ */
+@property (nullable, copy, nonatomic, readonly) NSString *_id;
```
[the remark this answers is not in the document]
nickpatrick: I can switch that if we really care. FWIW customers are never going to be initializing these, just accessing the properties, so the ordering doesn't really matter from the customer's perspective

RadarSDK/RadarVerificationManager.m
nickpatrick: Independent of `startTracking()`, the desired behavior here is definitely to do a track right away. I've tested this extensively in simulator / on real device and haven't observed duplicate tracks, so I don't think we need to make this change right now

RadarSDK/RadarVerificationManager.m
nickpatrick: The use case here is that some operators do a "geolocation pre-check" on app launch and then select a "site experience" for a single jurisdiction. So (1) it's fine to make them call it on every app launch session and (2) we don't need to support multiple jurisdictions. Happy to share more color on the use case and who needs this offline if you want

RadarSDK/RadarVerificationManager.m
lmeier: Cool, sounds good. Just a way to make things more robust to imperfect implementations (multiple `startTrackingVerified` calls). Agree that it's not necessary to change atm

lmeier (approved):

PR #385, opened 2024-08-23
Optional stopped default value

lmeier:
Presently `user.stopped` defaults to `NO` on the first `/track`. This allows us to set a default value via `RadarSdkConfiguration`. Some customers want the first `/track` to have `stopped: YES`.

### DIFF
--- a/RadarSDK/RadarSdkConfiguration.h
+++ b/RadarSDK/RadarSdkConfiguration.h
@@ -33,6 +33,8 @@ NS_ASSUME_NONNULL_BEGIN
 
 @property (nonatomic, assign) BOOL useLocationMetadata;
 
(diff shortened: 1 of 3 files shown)


KennyHuRadar:
I think we also need to clear the `[RadarState lastMovedLocation]` which we use to calculate the moved logic.
`handleLocation` will also check our last moved location here and set `stopped` based on that: https://github.com/radarlabs/radar-sdk-ios/blob/4367407b5a36bd0b557dcd73372b52500ae49b1b/RadarSDK/RadarLocationManager.m#L780
I am envisioning something like clearing that field out when the app terminates.

lmeier:
Don't think this will be necessary, closing.

PR #387, opened 2024-08-28
make lifecycle logging debug

KennyHuRadar:


lmeier (approved):
Let's ship https://github.com/radarlabs/radar-sdk-ios/pull/361/files as well

PR #388, opened 2024-08-30
Migrate RadarLogBuffe off main thread

KennyHuRadar:
Resource intensive actions should be performed off the main thread per Apple's suggestions.
This PR may seeks to migrate such tasks off the main thread where it may potentially cause freezing of the UI.
The target of migration is the RadarLogBuffer.


KennyHuRadar:
To facilitate a quicker ship, we may want to just ship a version of this PR without the changes to the track path.
We suspect that the logging workload is the prime suspect. We can then observe if changes to the track path are still needed.

PR #389, opened 2024-09-05
Log app open from notifications

KennyHuRadar:
### Changes to the interface and developer experience:

- This change is transparent to developers not intending to use this feature. No change is required
```
Radar.initialize(publishableKey: "prj_test_pk_000")
```
- For native iOS users who wishes to use the "automatic setup", they should initialize as such 
```
let radarInitializeOptions = RadarInitializeOptions()
radarInitializeOptions.autoSetupNotificationConversion = true
Radar.initialize(publishableKey: "prj_test_pk_000", radarInitializeOptions: radarInitializeOptions )
```
- For native iOS users who wish to perform the manual setup (easier to reason about). they should add this into their `UNUserNotificationCenterDelegate`
```
 func userNotificationCenter(_ center: UNUserNotificationCenter, didReceive response: UNNotificationResponse, 
    withCompletionHandler completionHandler: @escaping () -> Void) {        
            Radar.logConversion(response: response)
            completionHandler()
        }
```
- For react native users who are using SFMC, the setup is similar to the iOS manual setup.
- For react native users who are initializing in native code, they can opt to either implement the automatic or manual setup
- For react native users initializing in react native, they can call `Radar.nativeSetup()` in the app delegate to use the automatic setup. React native users who are using a library to implement `UNUserNotificationCenterDelegate` (e.g. expoNotifications) are only able to use the automatic setup (unless they wish to create and use a modified fork of that notification library) 

image
Dogfooded this current build, refer to the values that I generated on the last day (sept 6)

https://expo.dev/accounts/radarlabs/projects/waypoint/builds/079839d6-e7ca-4e2c-8351-04d18e15b721
(this is a "bricked" waypoint build which only can do trackonces using this build)

### Additional QA:
working with one signal 
image
Braze:
image
SFMC
image

### **associated server PRs:**
feature flags: https://github.com/radarlabs/server/pull/6106
proper parsing of the metadata string into a json: https://github.com/radarlabs/server/pull/6136


RadarSDK/RadarNotificationHelper.m
```diff
@@ -68,4 +75,112 @@ + (void)showNotificationsForEvents:(NSArray<RadarEvent *> *)events {
...
+            }
+        } else {
+            [[RadarLogger sharedInstance] logWithLevel:RadarLogLevelDebug message:@"Notification permissions not granted. Skipping adding notifications."];
+            return;
+        }
+    }];
+}
+
+
++ (void)checkNotificationPermissionsWithCompletion:(NotificationPermissionCheckCompletion)completion {
```
lmeier, line 172:
```suggestion
+ (void)checkNotificationPermissionsWithCompletionHandler:(NotificationPermissionCheckCompletion)completionHandler {
```

lmeier: and corresponding change below

RadarSDK/Radar.m
```diff
@@ -446,15 +456,19 @@ + (void)sendLogConversionRequestWithName:(NSString * _Nonnull) name
...
-    if (lastAppOpenTimeInterval > 1) {
-        [RadarSettings updateLastAppOpenTime];
-        [self sendLogConversionRequestWithName:@"opened_app" metadata:nil completionHandler:^(RadarStatus status, RadarEvent * _Nullable event) {
-            NSString *message = [NSString stringWithFormat:@"Conversion name = %@: status = %@; event = %@", event.conversionName, [Radar stringForStatus:status], event];
-            [[RadarLogger sharedInstance] logWithLevel:RadarLogLevelInfo message:message];
-        }];
-    }
+    // Perform a non-blocking sleep for 1 second before starting
+    dispatch_after(dispatch_time(DISPATCH_TIME_NOW, (int64_t)(1 * NSEC_PER_SEC)), dispatch_get_global_queue(DISPATCH_QUEUE_PRIORITY_DEFAULT, 0), ^{
+        // if opened_app has been logged within the last second, don't log it again
```
lmeier, line 461:
```suggestion
        // if opened_app has been logged within the last  two seconds, don't log it again
```

^that's what it needs to be per the logic on L464, right.

RadarSDK/Radar.m
```diff
@@ -446,15 +456,19 @@ + (void)sendLogConversionRequestWithName:(NSString * _Nonnull) name
...
+        NSTimeInterval lastAppOpenTimeInterval = [[NSDate date] timeIntervalSinceDate:[RadarSettings lastAppOpenTime]];
+        
+        if (lastAppOpenTimeInterval > 2) {
+            [RadarSettings updateLastAppOpenTime];
+            [self sendLogConversionRequestWithName:@"opened_app" metadata:nil completionHandler:^(RadarStatus status, RadarEvent * _Nullable event) {
```
lmeier, line 466: maybe add an explanatory comment that b/c this isn't from a push notification, metadata is nil

RadarSDK/Radar.m
```diff
@@ -489,10 +503,32 @@ + (void)logConversionWithName:(NSString *)name
 
 
 + (void)logConversionWithNotification:(UNNotificationRequest *)request {
+    [self logConversionWithNotification:request eventName: @"opened_app" conversionSource:@"notification" deliveredAfter: nil];
```
lmeier, line 506: Looks like we're always having `deliveredAfter` nil. I think we need to remove this param and use the registered at below in lieu of that, right?

RadarSDK/Radar.m
```diff
@@ -58,6 +66,11 @@ + (void)initializeWithPublishableKey:(NSString *)publishableKey {
     [RadarSettings setPublishableKey:publishableKey];
 
     RadarSdkConfiguration *sdkConfiguration = [RadarSettings sdkConfiguration];
+
+    if (NSClassFromString(@"XCTestCase") == nil) {
+        [Radar nativeSetup];
+    }
```
lmeier, line 72: This is before our GET config call, but it'd be nice if there was a way of turning this off remotely. Maybe we should check a feature flag regardless.

RadarSDK/RadarUtils.m
```diff
@@ -164,6 +164,18 @@ + (NSString *)dictionaryToJson:(NSDictionary *)dict {
...
+    NSArray<NSString *> *components = [identifier componentsSeparatedByString:@"_"];
+    if (components.count != 4) {
+        return nil; // Invalid format
+    }
+    
+    NSString *geofenceId = components[2];
+    NSString *registeredAt = components[3];
+    
+    return @{@"geofenceId": geofenceId, @"registeredAt": registeredAt};
+}
```
lmeier, line 177: I think there's a chance we graduate from using the geofence id here to using something like a "notificationId" at some point. Perhaps we'd want to future proof this method (e.g. make sure that we register geofenceId and registeredAt if the second component is "geofence"

lmeier:
I also think we might want to start sending up notification authorization in config and track just like we do for location authorization. The server work to handle can follow.

RadarSDK/Radar.m
KennyHuRadar: Yes. Honestly not the most "academic" solution to this issue, but favoring it right now for its simplicity.

RadarSDK/Radar.m
KennyHuRadar: The `deliveredAfter` is currently unused as this PR excludes the on premise notification delivered piece. We still need that field.
We do not need a `registeredAt` field as that information is encoded into our identifier and can be inferred.

RadarSDK/Radar.m
KennyHuRadar: Having this capability will be ideal, but this can result in unsafe behavior.
As of now, our API calls are not blocking and relies on callbacks. Calling `nativeSetup` within a callback is unsafe as there is no guarantee when this code will be called. 
Having the swizzle called later in the life cycle may result in missed logging of app opens via notifications.
If we are still interested in using background processing, registering background processes later in the lifecycle will crash the app.
We may need to consider configurations that we can pass in as variables to `initialize`.

RadarSDK/Radar.m
lmeier: Oh yes, duh. Makes sense.

RadarSDK/RadarUtils.m
KennyHuRadar: that's a really good point, but since this internal helper method doesn't need to be backward/forward compatible, maybe we can make that change when we graduate to a new form of identifiers to take the guesswork out of the equation?

RadarSDK/Include/Radar.h
```diff
@@ -649,13 +662,22 @@ typedef void (^_Nonnull RadarLogConversionCompletionHandler)(RadarStatus status,
             completionHandler:(RadarLogConversionCompletionHandler)completionHandler NS_SWIFT_NAME(logConversion(name:revenue:metadata:completionHandler:));
 
 /**
-logConversionWithNotification
+Logs a conversion with a notification
```
nickpatrick, line 665: @KennyHuRadar @lmeier Let's try to be more consistent with comments/docs. Other functions are doc'd like `Manually updates the user's location.`, with a leading space and a period at the end. So maybe change `Logs a conversion with a notification` to ` Logs a conversion with a notification.`

RadarSDK/Radar.m
```diff
@@ -41,7 +42,14 @@ + (id)sharedInstance {
...
 
-+ (void)initializeWithPublishableKey:(NSString *)publishableKey {
++ (void) nativeSetup {
+    static dispatch_once_t onceToken;
+    dispatch_once(&onceToken, ^{
+        [RadarNotificationHelper swizzleNotificationCenterDelegate];
+    });
+}
+
++ (void)initializeWithPublishableKey:(NSString *)publishableKey radarInitializeOptions:(RadarInitializeOptions *)radarInitializeOptions {
```
nickpatrick, line 52: Let's just call the param `options:` instead of `radarInitializeOptions:`

RadarSDK/Include/Radar.h
```diff
@@ -1203,6 +1225,13 @@ logConversionWithNotification
  */
 + (NSDictionary *)dictionaryForLocation:(CLLocation *)location NS_SWIFT_NAME(dictionaryForLocation(_:));
 
+
+/**
+ Perform optional setup for Radar SDK within the app delegate. This method only needs to be called if Radar is initalized in cross platform code.
+
+ */
++ (void) nativeSetup NS_SWIFT_NAME(nativeSetup());
```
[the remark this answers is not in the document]
nickpatrick: Is this the best possible name for this function?

RadarSDK/Include/RadarInitializeOptions.h
```diff
@@ -0,0 +1,17 @@
...
+//  Created by Kenny Hu on 9/10/24.
+//  Copyright © 2024 Radar Labs, Inc. All rights reserved.
+//
+
+
+#import <Foundation/Foundation.h>
+
+@interface RadarInitializeOptions : NSObject
+
+@property (assign, nonatomic) BOOL autoSetupNotificationConversion;
```
nickpatrick, line 14: Is `setup` the right verb here? Should `conversion` be plural? Would `autoLogNotificationConversions` be clearer?

lmeier (commented):
1 change necessary so that we don't require a migration, then this is g2g

RadarSDK/Radar.m
```diff
@@ -41,7 +42,14 @@ + (id)sharedInstance {
...
 
-+ (void)initializeWithPublishableKey:(NSString *)publishableKey {
++ (void) nativeSetup {
+    static dispatch_once_t onceToken;
+    dispatch_once(&onceToken, ^{
+        [RadarNotificationHelper swizzleNotificationCenterDelegate];
+    });
+}
+
++ (void)initializeWithPublishableKey:(NSString *)publishableKey options:(RadarInitializeOptions *)options {
```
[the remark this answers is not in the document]
lmeier: nevermind, it's there

lmeier (approved):

RadarSDK/RadarSettings.m
```diff
@@ -333,4 +333,8 @@ + (NSString *)xPlatformSDKVersion {
     return [[NSUserDefaults standardUserDefaults] stringForKey:kXPlatformSDKVersion];
 }
 
++ (BOOL)useOpenedAppConversion {
+    return [[self sdkConfiguration] useOpenedAppConversion];
+}
+
```
lmeier, line 339: This should default true I think (first one should happen prior to getting anything synced down from initialize, right?)

lmeier (changes requested):

lmeier (approved):

PR #390, opened 2024-09-12
change log level reference to that stored in sdkconfiguration

KennyHuRadar:


RadarSDK/Radar.m
```diff
@@ -1295,9 +1295,9 @@ - (void)dealloc {
     [[NSNotificationCenter defaultCenter] removeObserver:self];
 }
 
-+ (BOOL)isTestKey {
++ (BOOL)canFlushLogs {
     NSString *publishableKey = [RadarSettings publishableKey];
-    if ([publishableKey hasPrefix:@"prj_test_pk"] || [publishableKey hasPrefix:@"org_test_pk"] || [RadarSettings userDebug]) {
+    if ([publishableKey hasPrefix:@"prj_test_pk"] || [publishableKey hasPrefix:@"org_test_pk"] || [RadarSettings sdkConfiguration].logLevel == RadarLogLevelDebug) {
```
lmeier, line 1300:
```suggestion
    if ([publishableKey hasPrefix:@"prj_test_pk"] || [publishableKey hasPrefix:@"org_test_pk"] || [RadarSettings userDebug] || ([RadarSettings sdkConfiguration].logLevel && [RadarSettings sdkConfiguration].logLevel != RadarLogLevelNone)) {
```

lmeier: ^if we're remotely setting a log level, we should be flushing, so long as it's not log level none. Also need to flush if the user has been manually overwritten

lmeier (approved):
worth 1 quick sanity check perhaps, but this is good to go

PR #393, opened 2024-09-24
short circuit track verified if getConfig fails 

KennyHuRadar:
- fix bug of track verified not returning if `getConfig` api call fails
- short circuit the track verified process if `getConfig` fails, we should not bother performing the rest of the `trackVerified`


nickpatrick (approved):
Good catch. Let's test thoroughly before releasing.

Do we also need to make changes on Android?

Suggested release notes:

```
- Improves `Radar.trackVerified()` error handling.
```

PR #394, opened 2024-09-24
use none-deprecated archive functions when possible

ShiCheng-Lu:
The archive function that we were using was deprecated, and it was giving security warnings to customers integrating the SDK. The new, secure archive functions were introduces with iOS 11.0. So for iOS 11.0 +, we are switching to using the secure archive functions in order to avoid security issues.

I the archiving and unarchiving of RadarLog is already tested. I tested it with original archiving with new unarchiving just to make sure that previously saved data do not become invalid after this function change.

For RadarReplay, I added the test for archiving and unarchiving by creating the replayParam the same way it was created in /track, then using the `writeNewReplayToBuffer` and `loadReplaysFromPersistentStore` methods to verify the archiving/unarchiving is working. Same test with old archiving to verify updating sdk doesn't invalidate existing stored data.

(+267 lines might look scary, but most of it is in `RadarTestUtil` to construct the track param for replay test)


RadarSDK/RadarReplayBuffer.m
```diff
@@ -150,14 +158,25 @@ - (void)removeReplaysFromBuffer:(NSArray<RadarReplay *> *)replays {
...
+    }
     [[NSUserDefaults standardUserDefaults] setObject:replaysData forKey:@"radar-replays"];
 }
 
 - (void)loadReplaysFromPersistentStore {
     NSData *replaysData = [[NSUserDefaults standardUserDefaults] objectForKey:@"radar-replays"];
     if (replaysData) {
-        NSArray *replays = [NSKeyedUnarchiver unarchiveObjectWithData:replaysData];
+        NSArray *replays;
+        if (@available(iOS 11.0, *)) {
```
KennyHuRadar, line 174: I'm a little confused about iOS 11 version gate since NSSecureCoding is available from iOS 6

RadarSDKTests/RadarTestutils.m
```diff
@@ -18,4 +18,181 @@ + (NSDictionary *)jsonDictionaryFromResource:(NSString *)resource {
     return jsonDict;
 }
 
++ (NSMutableDictionary *)createTrackParamWithLocation:(CLLocation *_Nonnull)location
```
KennyHuRadar, line 21: I guess using the default values returned by `radarState` and `radarSettings` is perfectly fine in this case, but can we just put a comment noting that?

RadarSDKTests/RadarSDKTests.m
```diff
@@ -1489,6 +1490,22 @@ - (void)test_RadarLogBuffer_purge {
     [[RadarLogBuffer sharedInstance]clearBuffer];
 }
 
+- (void)test_RadarReplayBuffer_writeAndRead {
```
KennyHuRadar, line 1493: thank you for adding this test

RadarSDK/RadarReplayBuffer.m
ShiCheng-Lu: the `archivedDataWithRootObject: requireingSecureCoding:` function is only introduced in iOS 11.0. the function that we were using (without the second param) is the one raising security alerts.

RadarSDKTests/RadarTestutils.m
ShiCheng-Lu: added function docstring in the header file for this behaviour

PR #395, opened 2024-09-25
migrate RadarUtils

ShiCheng-Lu:


ShiCheng-Lu:
merge conflict is the sdk version update, will update it after review since there may be more version updates while this is still in review.

PR #398, opened 2024-10-02
Kennyhu/fence 1912 better abstraction for permissions handling

KennyHuRadar:


.github/workflows/build-and-test.yml
```diff
@@ -12,10 +12,13 @@ jobs:
     runs-on: macos-latest
 
     steps:
+      - uses: maxim-lobanov/setup-xcode@v1
```
KennyHuRadar, line 15: we need this because `macos-latest` tells lies, this is normally fine but in this case we need a specific version of xcode to build, we should be able to remove this in the future.

PR #399, opened 2024-10-02
default log level to NONE for live project

ShiCheng-Lu:
https://linear.app/radarlabs/issue/FENCE-2088/default-log-level-to-none-for-production-projects

clean up the code path for flushing log buffer, we want to flush whenever we do have logs. the default log level for live is NONE and for test is INFO, which are set from the server via remote SDK config.

verified in example app.


lmeier (approved):
wait for @KennyHuRadar to also 👍

PR #404, opened 2024-10-16
Offline ramp up

KennyHuRadar:
This PR implements offline ramp ups.
* Config calls will sync down all 3 configurations of RTOs for the user, default, on-trip and in-geofence
* As we sync nearby geofences, we will also store nearby geofences into RadarState
* We deviate from the tech spec by not currently storing the entire RadarUser, instead we simply use the existing `geofenceIds` methods from `RadarState` to denote the geofences we are in. It should also be noted that this would affect the results reported by users using anonymous tracks, but the behavior "makes sense".  
* if the `sendLocation` network call fails, we instead use the `RadarOfflineManager`, written in swift, to contextualize the location lat longs.
    * The `RadarOfflineManager` is written as a stand in for the server that has access to all the information the client has stored. 
    * It takes the lat long and nearby geofences to perform point in circle calculations. From those calculation, it determines if the user is in a geofence or not. 
    * It then takes the stored RTO configurations of the user to ramp up or ramp down as needed.
* The main change in the "critical path" of the track lifecycle is that the SDK will update its RTO from config even if `sendLocation` returns a network error. This allows the `RadarOfflineManager` to ramp up or ramp down the SDK.
* Since RTO needs to be turned on anyways to this feature to be relevant, we simply use `RadarOfflineManager` to override the RTO, we do not need an new fields to manage.


RadarSDK/RadarGeofence.m
```diff
@@ -228,7 +228,7 @@ - (NSDictionary *)dictionaryValue {
         RadarCircleGeometry *circleGeometry = (RadarCircleGeometry *)self.geometry;
         [dict setValue:@(circleGeometry.radius) forKey:@"geometryRadius"];
         [dict setValue:[circleGeometry.center dictionaryValue] forKey:@"geometryCenter"];
-        [dict setValue:@"Circle" forKey:@"type"];
+        [dict setValue:@"circle" forKey:@"type"];
```
KennyHuRadar, line 231: we pass down the type as `circle` and `polygon` from the server and parse it as such, however, we were setting the dict with the capitalized version, this causes issues. We need to make sure this does not impact our cross-platform change.

RadarSDK/RadarLocationManager.m
```diff
@@ -585,10 +587,11 @@ - (void)replaceSyncedGeofences:(NSArray<RadarGeofence *> *)geofences {
             }
         }
     }
-
-    [RadarNotificationHelper removePendingNotificationsWithCompletionHandler: ^{
-        [RadarNotificationHelper addOnPremiseNotificationRequests:requests];
-    }];
+    if (NSClassFromString(@"XCTestCase") == nil) {
```
KennyHuRadar, line 590: this method is now ran in our test suite, so we need to suppress this chunk in tests as test env does not have a notification manager.

RadarSDK/RadarAPIClient.m
lmeier, line 1: I think if there's a network error and `replay: all` is on, we should change the log level of a network error from `error` to `info`.

KennyHuRadar: how deeply do we want to couple these offline server fallback behaviors with the idea of `replay`? Asking this as I'm already creating new flags.

RadarSDK/RadarGeofence.m
lmeier: I vaguely remember fixing this (maybe in RN, changing it to expect capitalization)

RadarSDK/RadarSdkConfiguration.m
```diff
@@ -75,6 +75,45 @@ - (instancetype)initWithDict:(NSDictionary *)dict {
...
+
+    NSObject *onTripTrackingOptionsObj = dict[@"onTripTrackingOptions"];
+    _onTripTrackingOptions = nil;
+    if (onTripTrackingOptionsObj && [onTripTrackingOptionsObj isKindOfClass:[NSDictionary class]]) {
+        RadarTrackingOptions *radarTrackingOptions = [RadarTrackingOptions trackingOptionsFromObject:onTripTrackingOptionsObj];
+        if (radarTrackingOptions) {
+            _onTripTrackingOptions = radarTrackingOptions;
+        }
+    } 
+
```
lmeier, line 110: See server PR for an alternative struct that I think we should prefer

RadarSDK/RadarGeofence.m
KennyHuRadar: hmm it might be easier to just change the fromJson within the native SDK to accept both capitalized and non capitalized names of the shape to avoid introducing knock on effects downstream, this can help avoid introducing a breaking change.

RadarSDK/Radar.m
```diff
@@ -198,20 +198,19 @@ + (void)trackOnceWithDesiredAccuracy:(RadarTrackingOptionsDesiredAccuracy)desire
...
-                                         
-                                     }
-
-                                     if (completionHandler) {
-                                         [RadarUtils runOnMainThread:^{
-                                             completionHandler(status, location, events, user);
-                                         }];
-                                     }
-                                 }];
+                                    if (config != nil) {
```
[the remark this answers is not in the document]
KennyHuRadar: This is one of the changes I highlighted regarding the track lifecycle. When the server track request fails, we fallback to the offline manager to act as a stand-in. In these cases, although we return a non-success status, we still receive a valid config, which the rest of the track lifecycle should process.

Therefore, instead of relying on the status code, we should check if the config is non-nil. A non-nil config only occurs either from a successful server-side track or when the offline manager handles the network error.

RadarSDK/RadarLocationManager.m
```diff
@@ -970,13 +980,16 @@ - (void)sendLocation:(CLLocation *)location stopped:(BOOL)stopped source:(RadarL
...
                                          completionHandler:^(RadarStatus status, NSDictionary *_Nullable res, NSArray<RadarEvent *> *_Nullable events, RadarUser *_Nullable user,
                                                              NSArray<RadarGeofence *> *_Nullable nearbyGeofences, RadarConfig *_Nullable config, RadarVerifiedLocationToken *_Nullable token) {
-                                             self.sending = NO;
-                                             if (status != RadarStatusSuccess || !config) {
-                                                 return;
-                                             }
+                                            self.sending = NO;
+
+                                            if (config != nil) {
+                                                [self updateTrackingFromMeta:config.meta];
```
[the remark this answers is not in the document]
KennyHuRadar: agreed

RadarSDK/RadarOfflineManager.swift
```diff
@@ -0,0 +1,86 @@
...
+//
+//  Created by Kenny Hu on 10/16/24.
+//  Copyright © 2024 Radar Labs, Inc. All rights reserved.
+//
+
+import Foundation
+import CoreLocation
+
+@objc(RadarOfflineManager) class RadarOfflineManager: NSObject {
+    @objc public static func contextualizeLocation(_ location: CLLocation, completionHandler: @escaping (RadarConfig?) -> Void) {
```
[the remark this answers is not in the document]
KennyHuRadar: I intentionally kept it more vague, it was my understanding that this might become a place where stuff like client side event generation and trip updates may occur?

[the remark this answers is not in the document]
KennyHuRadar: `updateTrackingOptionsFromOfflineLocation`

RadarSDK/RadarState.m
```diff
@@ -190,4 +193,31 @@ + (BOOL)notificationPermissionGranted {
     return [[NSUserDefaults standardUserDefaults] boolForKey:kNotificationPermissionGranted];
 }
 
++ (void)setNearbyGeofences:(NSArray<RadarGeofence *> *_Nullable)nearbyGeofences {
```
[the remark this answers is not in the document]
KennyHuRadar: per offline discussion, no change required as of now.

KennyHuRadar:
We should extend this PR to more robustly handle high number of desired synced geofences (more than 20), I will circle back to address this asap.

PR #405, opened 2024-10-21
Replay time ordering

lmeier:
We do 2 things here:
- for all tracks we include the `updatedAtMsDiff`, which will be used to determine the `location.updatedAt`
- we modify `updatedAtMsDiff` in `RadarAPIHelper` to update `updatedAtMsDiff` at request time, as `updatedAtMsDiff` is meant to represent the difference between the client request time and the timestamp of the location processed by the SDK. This is an important fix to locations being replayed as previously their `updatedAtMsDiff` was determined by their original (failed) request time — not actually the time at successful request.


RadarSDK/RadarAPIClient.m
```diff
@@ -251,10 +253,12 @@ - (void)trackWithLocation:(CLLocation *_Nonnull)location
         params[@"floorLevel"] = @(location.floor.level);
     }
     long nowMs = (long)([NSDate date].timeIntervalSince1970 * 1000);
-    if (!foreground) {
-        long timeInMs = (long)(location.timestamp.timeIntervalSince1970 * 1000);
+    long timeInMs = (long)(location.timestamp.timeIntervalSince1970 * 1000);
```
KennyHuRadar, line 256: I think this var can be named more descriptively, like `locationTimeInMS`?

RadarSDK/RadarAPIHelper.m
```diff
@@ -67,9 +67,24 @@ - (void)requestWithMethod:(NSString *)method
             }
 
             if (params) {
-                [req setHTTPBody:[NSJSONSerialization dataWithJSONObject:params options:0 error:NULL]];
+                if (!params[@"updatedAtMsDiff"]) {
+                    [req setHTTPBody:[NSJSONSerialization dataWithJSONObject:params options:0 error:NULL]];
+                } else {
```
tjulien, line 72: do we need to change the code around `RadarAPIClient line 423`? Seems relevant

ShiCheng-Lu:
I think on [here](https://github.com/radarlabs/radar-sdk-ios/blob/109283d1bbafd15f8c26c9286e277a206d8160cc/RadarSDK/RadarAPIClient.m#L421C36-L421C77) we want to save the location's time (timeInMs) instead of the request time for the replay

ShiCheng-Lu:
> I think on [here](https://github.com/radarlabs/radar-sdk-ios/blob/109283d1bbafd15f8c26c9286e277a206d8160cc/RadarSDK/RadarAPIClient.m#L421C36-L421C77) we want to save the location's time (timeInMs) instead of the request time for the replay

nvm, this is not going to be used. Maybe we should get rid of the timestamps that will no longer be used to avoid future confusion.

RadarSDK/RadarAPIHelper.m
```diff
@@ -67,9 +67,38 @@ - (void)requestWithMethod:(NSString *)method
...
+                            NSMutableDictionary *updatedReplay = [replay mutableCopy];
+                            NSNumber *replayLocationMs = replay[@"locationMs"];
+                            if (replayLocationMs) {
+                                long replayUpdatedAtMsDiff = nowMs - [replayLocationMs longValue];
+                                updatedReplay[@"updatedAtMsDiff"] = @(replayUpdatedAtMsDiff);
+                            }
+                            [updatedReplays addObject:updatedReplay];
+                        }
+                        requestParams[@"replays"] = updatedReplays;
+                    }
```
ShiCheng-Lu, line 95: we should put this logic in `RadarAPIClient flushReplays:completionHandler:`

RadarSDK/RadarAPIClient.m
```diff
@@ -251,10 +251,12 @@ - (void)trackWithLocation:(CLLocation *_Nonnull)location
...
+    RadarTrackingOptions *options = [Radar getTrackingOptions];
+    if (!foreground || options.replay != RadarTrackingOptionsReplayNone) {
         params[@"updatedAtMsDiff"] = @(nowMs - timeInMs);
```
ShiCheng-Lu, line 257: all track requests should have this updatedAtMsDiff, likely = 0 for normal tracks, but it'll keep our server code path consistent. We also don't need to set locationMs in the request as its not used, a bit of wasted bandwidth, we can send it at line:417 if the request fails.

RadarSDK/RadarAPIHelper.m
lmeier: Yea, think that's a more natural spot for it, but also, I think we want to do the top level `updatedAtMsDiff` replacement in `APIHelper` because if `/track` is called twice, there will be a sleep interval, and we want to replace it after that time.

So the option would be to have the above in `RadarAPIHelper` and the replay replacement in `RadarAPIClient flushReplays:completionHandler` or have it all in the same place in `RadarAPIHelper`.

I'm not sure, but I lean towards the latter I think.

RadarSDK/RadarAPIClient.m
ShiCheng-Lu: we can basically think that all tracks are replays, but some are replayed immediately.

lmeier: We toss in the `locationMs` for updating the `updatedAtMsDiff` in `RadarAPIHelper`.

lmeier: I had initial hesitancy towards making all `/track`s have `updatedAtMsDiff`, but yea, your model that they'd all have it matches mine. Might as well change that.

RadarSDK/RadarAPIClient.m
```diff
@@ -251,10 +251,10 @@ - (void)trackWithLocation:(CLLocation *_Nonnull)location
         params[@"floorLevel"] = @(location.floor.level);
     }
     long nowMs = (long)([NSDate date].timeIntervalSince1970 * 1000);
-    if (!foreground) {
-        long timeInMs = (long)(location.timestamp.timeIntervalSince1970 * 1000);
-        params[@"updatedAtMsDiff"] = @(nowMs - timeInMs);
-    }
+    long timeInMs = (long)(location.timestamp.timeIntervalSince1970 * 1000);
```
nickpatrick, line 254: Nit: Call this var `locationMs`?

RadarSDK.podspec
```diff
@@ -1,6 +1,6 @@
 Pod::Spec.new do |s|
   s.name                  = 'RadarSDK'
-  s.version               = '3.18.4'
+  s.version               = '3.18.4-beta.2'
```
KennyHuRadar, line 3: bump version

RadarSDK/RadarAPIClient.m
```diff
@@ -251,10 +251,13 @@ - (void)trackWithLocation:(CLLocation *_Nonnull)location
         params[@"floorLevel"] = @(location.floor.level);
     }
     long nowMs = (long)([NSDate date].timeIntervalSince1970 * 1000);
-    if (!foreground) {
-        long timeInMs = (long)(location.timestamp.timeIntervalSince1970 * 1000);
-        params[@"updatedAtMsDiff"] = @(nowMs - timeInMs);
+    long locationMs = (long)(location.timestamp.timeIntervalSince1970 * 1000);
+    RadarTrackingOptions *options = [Radar getTrackingOptions];
+    // if not verified
```
KennyHuRadar, line 256: nit: this comment not very useful considering next line lol

RadarSDK/RadarAPIClient.m
```diff
@@ -251,10 +251,12 @@ - (void)trackWithLocation:(CLLocation *_Nonnull)location
         params[@"floorLevel"] = @(location.floor.level);
     }
     long nowMs = (long)([NSDate date].timeIntervalSince1970 * 1000);
-    if (!foreground) {
-        long timeInMs = (long)(location.timestamp.timeIntervalSince1970 * 1000);
-        params[@"updatedAtMsDiff"] = @(nowMs - timeInMs);
+    long locationMs = (long)(location.timestamp.timeIntervalSince1970 * 1000);
+    RadarTrackingOptions *options = [Radar getTrackingOptions];
+    if (!verified || !foreground) {
```
tjulien, line 256: so verified background reqs will use `updatedAtMsDiff`, is that intended?
Are we intending to change how we treat foreground timestamps or holding off on that bc it's a bigger change not needed for the fix?  

Would recommend updating the PR description with a high level description of the changes to make it easier to understand.

tjulien:
@lmeier what's the linear for this?

RadarSDK/RadarAPIClient.m
lmeier: Yes that's correct. Per slack conversation with Nick, the intention is to not change any timestamp logic for `verified` requests.

All non verified `/tracks` will now use `updatedAtMsDiff`, both foreground and background. There's no reason to hold off on making (non-verified) foreground requests use this logic -- in fact it's required for making sure that foreground requests slot in correctly in between background requests.

Ooo yea, good call out re: pr description -- just updated it.

tjulien: should the code be:
```
if (!verified) {
```

as is it doesn't match your comment.  Unless I'm missing something

RadarSDK/RadarNotificationHelper.m
```diff
@@ -136,7 +136,6 @@ + (void)removePendingNotificationsWithCompletionHandler:(void (^)(void))completi
         NSMutableArray *identifiers = [NSMutableArray new];
         for (UNNotificationRequest *request in requests) {
             if ([request.identifier hasPrefix:kSyncGeofenceIdentifierPrefix]) {
-                [[RadarLogger sharedInstance] logWithLevel:RadarLogLevelDebug message:[NSString stringWithFormat:@"Found pending notification to remove | identifier = %@", request.identifier]];
```
tjulien, line 139: intentional?

RadarSDK/RadarAPIClient.m
lmeier: We do want background verified tracks to use the existing logic, i.e. `if (!foreground) ...`

So either `if (!verified || !foreground) {`

Does that make sense?

RadarSDK/RadarNotificationHelper.m
lmeier: This isn't related to this (but intentional). I can take it to keep it atomic.

RadarSDK/RadarAPIHelper.m
```diff
@@ -67,9 +67,38 @@ - (void)requestWithMethod:(NSString *)method
             }
 
             if (params) {
-                [req setHTTPBody:[NSJSONSerialization dataWithJSONObject:params options:0 error:NULL]];
+                if (!params[@"updatedAtMsDiff"] && !params[@"replays"]) {
```
tjulien, line 70: nit: it would be nice if the Android logic matched this `if` check

RadarSDK/RadarAPIClient.m
```diff
@@ -251,10 +251,9 @@ - (void)trackWithLocation:(CLLocation *_Nonnull)location
         params[@"floorLevel"] = @(location.floor.level);
     }
     long nowMs = (long)([NSDate date].timeIntervalSince1970 * 1000);
-    if (!foreground) {
-        long timeInMs = (long)(location.timestamp.timeIntervalSince1970 * 1000);
-        params[@"updatedAtMsDiff"] = @(nowMs - timeInMs);
-    }
+    long locationMs = (long)(location.timestamp.timeIntervalSince1970 * 1000);
```
tjulien, line 254: ok so I guess the approach now is:
• change `foreground = true` to also use `updatedAtMsDiff` (for consistency?  I might have been the one to bring this up?)
• Don't special case `verified = true` on the client side, handle it differently on the server side

If we decided not to change `foreground = true` to `updatedAtMsDiff`, would the other changes to `updatedAtMsDiff` in these PRs be enough to fix the bug?  Or is the inconsistent behavior between `foreground = true` and `foreground = false` causing a problem?

RadarSDK/RadarAPIHelper.m
lmeier: Good callout, fixed: https://github.com/radarlabs/radar-sdk-ios/pull/405/commits/1cd42d92bdca78fd5fa211d06fdce75fc3d2ff55

RadarSDK/RadarAPIClient.m
lmeier: >  change foreground = true to also use updatedAtMsDiff (for consistency? I might have been the one to bring this up?)

Yes, for three reasons:
- not using `updatedAtMsDiff` causes a problem when a foreground location gets forced to server time while background locations are adjusted per `updatedAtMsDiff`, leading to inconsistent ordering (a subsequent background location can be brought erroneously before the foreground location by the `updatedAtMsDiff` adjustment).
- for consistency
- because there's no empirical reason to specially trust foreground locations from the OS as time-correct. It's true that foreground locations aren't subject to delays related to backgrounding, but the OS can and does deliver locations when foregrounded that have a non `now` timestamp

lmeier: > • Don't special case `verified = true` on the client side, handle it differently on the server side

Correct, per slack conversation. I'll link in #mobile

RadarSDK/RadarAPIHelper.m
```diff
@@ -67,9 +67,39 @@ - (void)requestWithMethod:(NSString *)method
             }
 
             if (params) {
-                [req setHTTPBody:[NSJSONSerialization dataWithJSONObject:params options:0 error:NULL]];
+                NSNumber *prevUpdatedAtMsDiff = params[@"updatedAtMsDiff"];
```
tjulien, line 70: Now that we're collecting `locationMS` we could also pass up `updatedAt` (we probably already are) we could do this bookkeeping server side but I think let's do that in a v2 in the interest of getting this out.

PR #406, opened 2024-10-23
startTrackingVerified() and stopTrackingVerified() improvements, add passed check in getVerifiedLocationToken()

nickpatrick:
Ports some changes from https://github.com/radarlabs/radar-sdk-android/pull/409 to iOS


RadarSDK/RadarVerificationManager.m
```diff
@@ -265,40 +255,40 @@ - (void)startTrackingVerifiedWithInterval:(NSTimeInterval)interval beacons:(BOOL
...
+                    [self callTrackVerified];
                 }
             });
 
             nw_path_monitor_start(_monitor);
         }
     }
 
-    trackVerified();
+    [self callTrackVerified];
```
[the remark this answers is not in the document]
ShiCheng-Lu: nvm, the 258 is inside a callback. github hid just the right part

RadarSDK/RadarVerificationManager.m
```diff
@@ -182,57 +182,47 @@ - (void)trackVerifiedWithBeacons:(BOOL)beacons completionHandler:(RadarTrackVeri
...
+        // min interval is 10 seconds
+        if (minInterval < 10) {
+            minInterval = 10;
+        }
+        
+        [NSObject cancelPreviousPerformRequestsWithTarget:self selector:@selector(intervalFired) object:nil];
+        
+        [[RadarLogger sharedInstance] logWithLevel:RadarLogLevelDebug message:[NSString stringWithFormat:@"Requesting token again in %f seconds | minInterval = %f; expiresIn = %f; startedInterval = %f", minInterval, minInterval, expiresIn, self.startedInterval]];
+        
+        [self performSelector:@selector(intervalFired) withObject:nil afterDelay:minInterval];
```
ShiCheng-Lu, line 217: I think there might be an edge case here where if `stopTrackingVerified` is called right after the `intervalFired` is starting, the `intervalFired` function will still run, and call this `callTrackVerified` and continue the loop. 
(depending on how iOS does time slicing and threading of the calls)

ShiCheng-Lu: I can reproduce this with a low tracking interval with the example app. if I click(call) stopTrackingVerified at certain times, it doesn't stop the tracking loop

RadarSDK/RadarVerificationManager.m
```diff
@@ -182,57 +182,47 @@ - (void)trackVerifiedWithBeacons:(BOOL)beacons completionHandler:(RadarTrackVeri
     }];
 }
 
-- (void)startTrackingVerifiedWithInterval:(NSTimeInterval)interval beacons:(BOOL)beacons {
```
lmeier, line 185: If I were starting fresh approaching this bugfix, I think I'd actually just opt for some state checks on `self.started` and `self.scheduled` (ie only start the loop if you weren't previously started).

I think the approach taken here works as well, but we should be cognizant of the bug ShiCheng pointed out re: stopping failing when a request is in flight: https://github.com/radarlabs/radar-sdk-ios/pull/406/files#r1812908221

RadarSDK/RadarVerificationManager.m
```diff
@@ -182,57 +182,47 @@ - (void)trackVerifiedWithBeacons:(BOOL)beacons completionHandler:(RadarTrackVeri
...
-        }];
-    };
+            // if expiresIn is shorter than interval, override interval
+            minInterval = MIN(expiresIn, self.startedInterval);
+        }
+        
+        // re-request early to maximize the likelihood that a cached token is available
+        if (minInterval > 20) {
+            minInterval = minInterval - 10;
+        }
```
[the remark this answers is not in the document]
nickpatrick: Good suggestion, made this change

RadarSDK/RadarVerificationManager.m
nickpatrick: Good catch, I added back a `started` flag to cover this

RadarSDK/RadarVerificationManager.m
nickpatrick: That was a good catch, fixed that

Intention here was to fix the double-start bug but also mirror the code on Android for consistency, so felt that it was worth making a slightly bigger change

PR #407, opened 2024-10-25
add set version script

ShiCheng-Lu:
There are two components of this PR, improved CI checks, and version updating script.

1. Improved CI checks.
- adding a check to make sure both versions do not exist in pod yet
- lint both SDKs before uploading either
This should make sure that if CI passes, both are successfully uploaded, and if CI fails, neither are uploaded.

2. version updating script.
- this script does everthing in the iOS version bump/release document in notion, I believe running the script will overall result in less version bumping mistakes than doing it by hand every time. Of course it's still recommended to follow through the document and check the changes in git.

(There was no documentation on bumping MotionSDK in notion, which should have been there, I've updated the notion document)


.github/workflows/release-sdk.yml
```diff
@@ -74,12 +74,21 @@ jobs:
...
+          echo "RadarSDKMotion version is OK"
+
       - name: Deploy to Cocoapods
         run: |
           set -eo pipefail
           pod lib lint --allow-warnings RadarSDK.podspec
-          pod trunk push --allow-warnings RadarSDK.podspec
           pod lib lint --allow-warnings RadarSDKMotion.podspec
+          pod trunk push --allow-warnings RadarSDK.podspec
           pod trunk push --allow-warnings RadarSDKMotion.podspec
```
ShiCheng-Lu, line 92: we definitely want to lint both before publishing any of the two.

RadarSDK.xcodeproj/project.pbxproj
```diff
@@ -949,7 +949,6 @@
 					"@executable_path/Frameworks",
 					"@loader_path/Frameworks",
 				);
-				MARKETING_VERSION = 3.9.6;
```
ShiCheng-Lu, line 952: this is the package settings, which overwrites the project settings, we should just have the version set for project settings which will be resolved everywhere to the single correct marketing version

RadarSDK.xcodeproj/project.pbxproj
```diff
@@ -979,7 +978,6 @@
...
```
ShiCheng-Lu, line 982: same as above

RadarSDKMotion/RadarSDKMotion.xcodeproj/project.pbxproj
```diff
@@ -347,6 +348,7 @@
 				GCC_WARN_UNUSED_VARIABLE = YES;
 				IPHONEOS_DEPLOYMENT_TARGET = 12.0;
 				LOCALIZATION_PREFERS_STRING_CATALOGS = YES;
+				MARKETING_VERSION = 3.18.4;
```
ShiCheng-Lu, line 351: adding the marketing version to motion sdks. (project setting)

set_version.sh
lmeier, line 1: I think there's also a `CFBundleShortVersionString` in the `Info.plist` that has lagged for a while and we don't generally update. We should get that too.

lmeier (commented):
I'm gonna test this with a bump today to verify it's good before 👍, but this looks good. Thank you — nice proactive change.

set_version.sh
ShiCheng-Lu: CFBundleShortVersionString is set to parameter substitution of #(MARKETING_VERSION) so we are good there.

KennyHuRadar (approved):
script works well from my testing and the changes to release script makes sense imo

PR #409, opened 2024-10-28
Perform nil check before clearing geofences

lmeier:
Relevant if what's being sent up is a `/track/replay`, where we early 200.


KennyHuRadar:
Took a look at android, it too faces the same issue that should be addressed.

lmeier:
> Took a look at android, it too faces the same issue that should be addressed.

https://github.com/radarlabs/radar-sdk-android/pull/413 @KennyHuRadar

KennyHuRadar:
Its worth noting that this does not alter tracking behavior in the case where there are no nearby geofences as the server would return an empty array instead of a null value. I don't think there is anything actionable right away but it does feel like a fragile part of our logic.

KennyHuRadar (approved):

PR #411, opened 2024-11-04
Allow for createTrip without starting tracking immediately

lmeier:
Allows you to pass `tripOptions.startTracking = NO` to not automatically start tracking. The default remains that tracking starts automatically.

Starting tracking also doesn't happen automatically if `trackingOptions.startTrackingAfter` is non-nil.

We explicitly check the local tracking options for `startTrackingAfter` and `stopTrackingAfter` so that RTO can be used in conjunction with these local tracking options.

Finally, we make sure that if you call `Radar.stopTracking()` after having started tracking with a `trackingOption.startTrackingAfter`, that tracking remains off until `Radar.startTracking()` is called again.


ahwreck:
@lmeier what if we change `scheduled: YES/NO` to `startTracking: true/false`?

RadarSDK/Radar.m
```diff
@@ -584,13 +586,14 @@ + (void)startTripWithOptions:(RadarTripOptions *)tripOptions
                                                      [RadarSettings removePreviousTrackingOptions];
                                                  }
 
-                                                 if (trackingOptions) {
+                                                 if (trackingOptions && trackingOptions.startTrackingAfter == nil) {
                                                      [self startTrackingWithOptions:trackingOptions];
-                                                 } else if (!Radar.isTracking) {
+                                                 } else if (trackingOptions) {
+                                                     [RadarSettings setTrackingOptions:trackingOptions];
+                                                 } else if (!Radar.isTracking && tripOptions && tripOptions.startTracking) {
```
KennyHuRadar, line 593: we are only calling `[RadarSettings setPreviousTrackingOptions:[RadarSettings trackingOptions]];` if we are tracking, I think we need to update `previousTrackingOptions` regardless under this implementation to avoid some edge casey situations where we dont' restore our old tracking options.

RadarSDK/RadarLocationManager.m
```diff
@@ -304,17 +304,18 @@ - (void)updateTracking:(CLLocation *)location fromInitialize:(BOOL)fromInitializ
     dispatch_async(dispatch_get_main_queue(), ^{
         BOOL tracking = [RadarSettings tracking];
         RadarTrackingOptions *options = [Radar getTrackingOptions];
+        RadarTrackingOptions *localOptions = [RadarSettings trackingOptions];
 
         [[RadarLogger sharedInstance] logWithLevel:RadarLogLevelDebug
                                            message:[NSString stringWithFormat:@"Updating tracking | options = %@; location = %@", [options dictionaryValue], location]];
 
-        if (!tracking && [options.startTrackingAfter timeIntervalSinceNow] < 0) {
+        if (!tracking && [localOptions.startTrackingAfter timeIntervalSinceNow] < 0) {
```
KennyHuRadar, line 312: I understand that we are implementing this to support scheduled trips in the future. But say we schedule the trip a few days from now, what would trigger our SDK attempting to `updateTracking` and actually start tracking at the scheduled time?

RadarSDK/Radar.m
lmeier: talked in person, we're good on this b/c `previousTrackingOptions` is exclusively about the tracking options prior to the trip lifecycle.

RadarSDK/RadarLocationManager.m
lmeier: also discussed irl. `startTrackingAfter` requires a `trackOnce` to indeed actually start tracking.

Example/Example/AppDelegate.swift
```diff
@@ -29,7 +29,7 @@ class AppDelegate: UIResponder, UIApplicationDelegate, UIWindowSceneDelegate, UN
         let radarInitializeOptions = RadarInitializeOptions()
         // Uncomment to enable automatic setup for notification conversions
         // radarInitializeOptions.autoSetupNotificationConversion = true
-        Radar.initialize(publishableKey: "prj_test_pk_0000000000000000000000000000000000000000", options: radarInitializeOptions )
+        Radar.initialize(publishableKey: "prj_test_pk_0000000000000", options: radarInitializeOptions )
```
KennyHuRadar, line 32: nit: do we want to change this back?

Example/Example/AppDelegate.swift
```diff
@@ -271,12 +271,42 @@ class AppDelegate: UIResponder, UIApplicationDelegate, UIWindowSceneDelegate, UN
         }
 
         demoButton(text: "startTrip") {
-            let tripOptions = RadarTripOptions(externalId: "299", destinationGeofenceTag: "store", destinationGeofenceExternalId: "123")
+            let tripOptions = RadarTripOptions(externalId: "300", destinationGeofenceTag: "store", destinationGeofenceExternalId: "123")
```
KennyHuRadar, line 274: nit: this may cause an annoyance for manual testing, I personally use UUIDS for these external IDs to avoid the server complaining about external ID collision.

KennyHuRadar (approved):
lgtm baring nits for example app changes

PR #412, opened 2024-11-13
SDK profiler

ShiCheng-Lu:
Switching over to using OpenTelemetry trace structure: 
https://github.com/radarlabs/radar-sdk-ios/pull/470

Introduce a RadarProfiler. keeps track of what time each operation started/ended, and produce a formatted string of the performance.

RadarProfiler has two timing methods `start(operation)` and `end(operation)` which tracks the time elapsed for a specific operation. In this PR it's being used for `trackVerified`. 

the `formatted` method returns a formatted string of all the operations that were tracked. e.g. `getLocation: 0.003, trackAPI: 1.518, total: 3.448, getAttestToken: 1.477, getConfig: 0.450`

then it's sent to logs. (currently debug level, but maybe this should be info) which then will be parsed and processed by the server (in the future)

---------------------
The test in swift does run in CI, but the log doesn't show up. It does show up it if fails, it'll say`RadarSDKTestsSwift.test_profiler()` failed. But no output for success for some reason (for now).

This test is written with the Swift Testing framework, which requires xcode 16. (updated github workflow to public pre-release for this).

### DIFF
--- a/RadarSDK/RadarTelemetry.swift
+++ b/RadarSDK/RadarTelemetry.swift
@@ -7,7 +7,7 @@
 
 import Foundation
 
-@objc(RadarProfiler) class RadarProfiler: NSObject {
+@objc(RadarTelemetry) class RadarTelemetry: NSObject {
     
     var startTimes = [String:Double]()
     var endTimes = [String:Double]()
--- a/RadarSDKTests/RadarSDKTestsSwift.swift
+++ b/RadarSDKTests/RadarSDKTestsSwift.swift
@@ -12,28 +12,28 @@ import Testing
 struct RadarSDKTestsSwift {
     
     @Test func test_profiler() async throws {
-        let profiler = RadarProfiler()
-        profiler.start("total")
-        profiler.start("test")
+        let telemetry = RadarTelemetry()
+        telemetry.start("total")
+        telemetry.start("test")
         
         usleep(1500000)
         
-        profiler.end("test")
+        telemetry.end("test")
         
-        #expect(profiler.formatted() == String(format: "test: %.3f", profiler.get("test")))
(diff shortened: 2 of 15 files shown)


RadarSDK/RadarTelemetry.swift
```diff
@@ -0,0 +1,39 @@
...
+        if let start = startTimes[tag],
+           let end = endTimes[tag] {
+            return (end - start)
+        } else {
+            return 0;
+        }
+    }
+    
+    @objc func formatted() -> String {
+        return endTimes.map { (tag, end) in
```
cameron-morrow-toptal, line 35: you might want to include the tags that have starts and no ends as like a "unfinished" or a -1 or something, just so they don't fail silently

cameron-morrow-toptal, line 35: will iterating over the map keep the original order of the tags? might be helpful for log readability

RadarSDKTests/RadarSDKTestsSwift.swift
```diff
@@ -0,0 +1,39 @@
...
+//  RadarSDKTests
+//
+//  Copyright © 2024 Radar Labs, Inc. All rights reserved.
+//
+
+import Foundation
+import Testing
+
+@Suite("SDK Tests")
+struct RadarSDKTestsSwift {
```
cameron-morrow-toptal, line 12: could add a test for starting a tag and not ending it, and a test for trying to end a tag that wasn't started

RadarSDK/RadarTelemetry.swift
ShiCheng-Lu: going to switch this over to using open telemetry specs.

PR #418, opened 2024-12-03
Suppress first run debug logs, add verified error callbacks

nickpatrick:


lmeier:
Heads up, looks like you need to update the example app delegate:
```
/Users/distiller/project/Example/Example/AppDelegate.swift:13:7: type 'AppDelegate' does not conform to protocol 'RadarVerifiedDelegate'
```

PR #419, opened 2024-12-04
Campaign opened_app analytics 

KennyHuRadar:
image


RadarSDK/Radar.m
```diff
@@ -538,8 +538,7 @@ + (void)logConversionWithNotification:(UNNotificationRequest *)request
                        deliveredAfter:(NSDate *)deliveredAfter {
     
     NSMutableDictionary *metadata = [[NSMutableDictionary alloc] initWithDictionary:request.content.userInfo];
-    NSDictionary<NSString *, NSString *> *result = [RadarUtils
-                                                    extractGeofenceIdAndTimestampFromIdentifier:request.identifier];
+    NSDictionary<NSString *, NSString *> *result = [RadarNotificationHelper extractMetadataFromNotificationIdentifier: request.identifier];
```
lmeier, line 541: Is it correct to be using the `request.identifier` for this? Are we liable to run into length issues? Wondering how much we can stuff in here.

Would be nice if we had some E2E validation of this, I'm worried about the campaign id getting cut off.

RadarSDK/RadarNotificationHelper.m
```diff
@@ -214,8 +214,38 @@ + (void)checkNotificationPermissionsWithCompletionHandler:(NotificationPermissio
...
+}
+
++ (NSString *)getNotificationIdentifier:(NSString *)geofenceId campaignName:(NSString *)campaignName {
+    NSString *identifier = [NSString stringWithFormat:@"%@%@", kSyncGeofenceIdentifierPrefix, geofenceId];
+    NSDate *now = [NSDate new];
+    NSTimeInterval lastSyncInterval = [now timeIntervalSince1970];
+    identifier = [identifier stringByAppendingString:[NSString stringWithFormat:@"_%f", lastSyncInterval]];
+    if (campaignName) {
+        identifier = [identifier stringByAppendingString:@"_"];
+        identifier = [identifier stringByAppendingString:campaignName];
```
lmeier, line 227: Why do we want (or need) the campaign name? Isn't the `id` enough?

lmeier: Ah, I see, we're not doing any `campaignId` — can we do that instead?

lmeier (dismissed):

lmeier (commented):
do we need a `extractMetadataFromMetadata` function?

lmeier (approved):

PR #423, opened 2024-12-11
update set_version.sh to work with beta versions

ShiCheng-Lu:


gregsadetsky:
this is my first time seeing this file - don't mean to reshuffle everything, but the `sed` seems somewhat risky, since it regexps _any_ `\d.\d.\d` string in RadarUtils..? i.e. if any other string like that appeared in that file, it would also be replaced with the sdk version?

could the actual version string be moved to a `#define` first (so that the version string appears on a single line with a clear name like `SDK_VERSION`), and then the sed line can be modified to only affect that SDK_VERSION? finally, `sdkVersion` would simply return that constant?

again, this feels safer than regexp'ing `return ...`

ShiCheng-Lu:
> this is my first time seeing this file - don't mean to reshuffle everything, but the `sed` seems somewhat risky, since it regexps _any_ `return \d.\d.\d.+` string in RadarUtils..? i.e. if any other string like that appeared in that file, it would also be replaced with the sdk version?
> 
> could the actual version string be moved to a `#define` first (so that the version string appears on a single line with a clear name like `SDK_VERSION`), and then the sed line can be modified to only affect that SDK_VERSION? finally, `sdkVersion` would simply return that constant?
> 
> again, this feels safer than regexp'ing `return ...`



> this is my first time seeing this file - don't mean to reshuffle everything, but the `sed` seems somewhat risky, since it regexps _any_ `return \d.\d.\d.+` string in RadarUtils..? i.e. if any other string like that appeared in that file, it would also be replaced with the sdk version?
> 
> could the actual version string be moved to a `#define` first (so that the version string appears on a single line with a clear name like `SDK_VERSION`), and then the sed line can be modified to only affect that SDK_VERSION? finally, `sdkVersion` would simply return that constant?
> 
> again, this feels safer than regexp'ing `return ...`

Thats a great idea, It'll be safer

gregsadetsky:
Great work! 


I don't want to be annoying but if you could also add a comment above the DEFINE line just to note that it's important that this line stays as it is (in case someone is tempted to refactor it), that'd be great

Thanks again!

gregsadetsky (approved):

PR #425, opened 2024-12-19
Notification diff

lmeier:


RadarSDK/RadarAPIClient.m
```diff
@@ -376,6 +377,48 @@ - (void)trackWithLocation:(CLLocation *_Nonnull)location
                                          }];
     }
 
+    if (YES) {
```
KennyHuRadar, line 380: why is this `YES`?

lmeier: placeholder while I was implementing sdk config

RadarSDK/RadarSdkConfiguration.m
```diff
@@ -81,6 +81,13 @@ - (instancetype)initWithDict:(NSDictionary *)dict {
         _useForegroundLocationUpdatedAtMsDiff = [(NSNumber *)useForegroundLocationUpdatedAtMsDiffObj boolValue];
     }
 
+    NSObject *useNotificationDiffObj = dict[@"useNotificationDiff"];
+    // TODO: make this default NO
+    _useNotificationDiff = YES;
```
lmeier, line 86: Need to default to off before shipping

PR #433, opened 2025-02-05
Event metadata for campaigns

KennyHuRadar:


RadarSDK/RadarNotificationHelper.m
```diff
@@ -72,6 +93,53 @@ + (void)showNotificationsForEvents:(NSArray<RadarEvent *> *)events {
...
+        return nil;
+    }
+
+    NSString *notificationText = [metadata objectForKey:@"radar:notificationText"];
+    NSString *notificationTitle = [metadata objectForKey:@"radar:notificationTitle"];
+    NSString *notificationSubtitle = [metadata objectForKey:@"radar:notificationSubtitle"];
+    NSString *notificationURL = [metadata objectForKey:@"radar:notificationURL"];
+    NSString *campaignId = [metadata objectForKey:@"radar:campaignId"];
+
+    if (notificationText) {
```
lmeier, line 113:
```suggestion
    if (notificationText & campaignType = 'notification')
```

KennyHuRadar: Am I right to say that this check is just for consistency?
Or do we see a future where we are explicitly using nearby geofences synced down for in app messaging as well?

lmeier: We discussed this in-person — idea is to be able to use the other metadata keys for different campaign types (notificaitons, in-app messages, etc)

lmeier (approved):

PR #437, opened 2025-03-03
Release build log level

lmeier:


RadarSDK/RadarSettings.m
```diff
@@ -261,12 +261,20 @@ + (RadarSdkConfiguration *_Nullable)sdkConfiguration {
...
+#endif
+}
+
 + (RadarLogLevel)logLevel {
     RadarLogLevel logLevel;
     if ([RadarSettings userDebug]) {
         logLevel = RadarLogLevelDebug;
     } else if ([[NSUserDefaults standardUserDefaults] objectForKey:kLogLevel] == nil) {
-        logLevel = RadarLogLevelInfo;
+        logLevel = [self isDebugBuild] ? RadarLogLevelInfo : RadarLogLevelNone;
```
nickpatrick, line 277: Similar feedback to Android. Should we move this into the conditional, and have the conditional be the same across iOS and Android? Behavior will be easier to follow then

nickpatrick (approved):

PR #439, opened 2025-03-10
Kenny/add accelerometer and magnetometer add air Pressure and motion flags

KennyHuRadar:
QA: https://github.com/radarlabs/schemas/pull/236


RadarSDK/RadarSettings.h
```diff
@@ -65,7 +65,6 @@ NS_ASSUME_NONNULL_BEGIN
 + (void)updateLastAppOpenTime;
 + (NSDate *)lastAppOpenTime;
 + (BOOL)useRadarModifiedBeacon;
-+ (BOOL)useLocationMetadata;
```
lmeier, line 68: If we totally rip this out in this PR, we'll need to proactively let people using `useLocationMetadata` know.

RadarSDK/Include/RadarMotionProtocol.h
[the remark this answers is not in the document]
KennyHuRadar: But I can also see a case where we want this from a UIUX perspective. 
I think the tradeoff is if want to take on this surface area and complexity (with shiftsmart being a recent data point) vs providing this functionality.

RadarSDK/RadarLocationManager.m
```diff
@@ -1141,14 +1172,6 @@ - (void)locationManager:(CLLocationManager *)manager didFailWithError:(NSError *
     [self callCompletionHandlersWithStatus:RadarStatusErrorLocation location:nil];
 }
 
-- (void)stopActivityAndMotionUpdates {
```
KennyHuRadar, line 1144: dead code

lmeier (approved):

PR #440, opened 2025-03-12
Kenny/merge in indoors inference

KennyHuRadar:
Hey @gregsadetsky , putting together a PR that we can potentially merge with the barometric sensor code to potentially cut a beta for United.

Here are the main changes I did:
- removes the public interface for surveying since we don't need it now.
- removes the diff for the example app.
- when the `start` method of `RadarIndoorSurvey` has a valid known location, we just use that instead of attempting to request a new location.
- removes the magnetometer portion of the code. 

Let me know if this merge makes sense and I missed anything :)


gregsadetsky:
hey this looks really great, huge congrats!

just to check, `[[RadarIndoorSurvey sharedInstance] start:` will always be called with a location, right? just want to double check what you mean by "we just use that instead of attempting to request a new location."

otherwise... I think it looks very good

KennyHuRadar:
I believe so, we are only calling it when we are about to send a location to the server in the tracking lifecycle or after we got a location as part of a track once.

gregsadetsky:
> I believe so, we are only calling it when we are about to send a location to the server in the tracking lifecycle or after we got a location as part of a track once.

super!!

PR #450, opened 2025-06-17
fix: timezone id key from server response

diegogarciar:
fixes https://github.com/radarlabs/flutter-radar/issues/86

The timezone id is defined as `id` not as `_id`
https://docs.radar.com/api#geocoding


diegogarciar:
@lmeier could you review

KennyHuRadar (approved):
lgtm

PR #453, opened 2025-06-25
fence 2365 Add indoor scan

gvivan:


KennyHuRadar:
you should drag the workspace of the new sub module into the workspace of the example app, and then link them together so that we can build the example with the submodules.
image

RadarSDK/Radar.m
```diff
@@ -1151,6 +1152,41 @@ + (void)logResigningActive {
 }
 
 
+#pragma mark - Indoors
+
++ (void)doIndoorSurvey:(NSString *)placeLabel
+             forLength:(int)surveyLengthSeconds
+        isWhereAmIScan:(BOOL)isWhereAmIScan
+     completionHandler:(RadarIndoorsSurveyCompletionHandler)completionHandler {
```
lmeier, line 1160:
```suggestion
+ (void)startIndoorSurvey:(NSString *)placeLabel
             duration:(int)surveyLengthSeconds
     completionHandler:(RadarIndoorsSurveyCompletionHandler)completionHandler {
```
I'd suggest the above renaming (a bit more clean, idiomatic).


We don't need `isWhereAmIScan:(BOOL)isWhereAmIScan` because that's always true for a survey. And I think we should rename that downstream as `survey` or something of the like.

RadarSDK/Radar.m
```diff
@@ -1151,6 +1152,41 @@ + (void)logResigningActive {
...
+    [[RadarLogger sharedInstance] logWithLevel:RadarLogLevelInfo type:RadarLogTypeSDKCall message:@"doIndoorsSurvey()"];
+    
+    Class RadarSDKIndoors = NSClassFromString(@"RadarSDKIndoors");
+    if (RadarSDKIndoors) {
+        // get location from the SDK using trackOnce
+        [Radar trackOnceWithDesiredAccuracy:RadarTrackingOptionsDesiredAccuracyHigh
+                                    beacons:NO
```
lmeier, line 1167: Per all our earlier discussion, I think we'll want to use the beacon RSSI values in our surveys as well. Probably should turn this to YES (or at least make it configurable) and pass the results through.

lmeier (commented):
a few thing

RadarSDK/Radar.m
```diff
@@ -236,6 +238,31 @@ + (void)trackOnceWithDesiredAccuracy:(RadarTrackingOptionsDesiredAccuracy)desire
                                  }];
                          };
 
+                         void (^maybeIndoorSurveyThenCallTrackAPI)(NSArray<RadarBeacon *> *_Nullable) = ^(NSArray<RadarBeacon *> *_Nullable beacons) {
```
KennyHuRadar, line 241: there's got to be a better name than `maybeIndoorSurveyThenCallTrackAPI`

RadarSDK/Radar.m
```diff
@@ -210,14 +211,15 @@ + (void)trackOnceWithDesiredAccuracy:(RadarTrackingOptionsDesiredAccuracy)desire
                              return;
                          }
 
-                         void (^callTrackAPI)(NSArray<RadarBeacon *> *_Nullable) = ^(NSArray<RadarBeacon *> *_Nullable beacons) {
+                         void (^callTrackAPI)(NSArray<RadarBeacon *> *_Nullable, NSString *_Nullable) = ^(NSArray<RadarBeacon *> *_Nullable beacons, NSString *_Nullable indoorSurvey) {
```
KennyHuRadar, line 214: are we sure we want to use NSString instead of NSData in the SDK. is there a reason why we are preferring strings here?

RadarSDK.xcodeproj/project.pbxproj
```diff
@@ -956,6 +959,8 @@
 				PROVISIONING_PROFILE_SPECIFIER = "";
 				"PROVISIONING_PROFILE_SPECIFIER[sdk=macosx*]" = "";
 				SKIP_INSTALL = YES;
+				SWIFT_OPTIMIZATION_LEVEL = "-Onone";
+				SWIFT_VERSION = 6.0;
```
ShiCheng-Lu, line 963: I swear swift 6.0 didn't work for me with CI, it was complaining swift 6 is not available.

RadarSDK/Radar.m
```diff
@@ -1151,6 +1180,48 @@ + (void)logResigningActive {
...
+                                [RadarSettings setInSurveyMode:NO];
+                                if (completionHandler) {
+                                    completionHandler(indoorSurveyResult, locationAtStartOfSurvey);
+                                }
+                            }];
+        }];
+    } else {
+        [RadarSettings setInSurveyMode:NO];
+        if (completionHandler) {
+            completionHandler(@"ERROR: RadarSDKIndoors submodule not available", [[CLLocation alloc] init]);
```
KennyHuRadar, line 1220: allocing a location here feels strange, does returning a null feel more natural?

RadarSDK/RadarBeacon.m
```diff
@@ -232,4 +232,9 @@ - (NSUInteger)hash {
     return [self.uuid hash] ^ [self.major hash] ^ [self.minor hash];
 }
 
+// set rssi
+- (void)setRssi:(NSInteger)rssi {
```
KennyHuRadar, line 236: when is this RSSI value ever used?
from my impression we want to somehow pass these RSSI value back to the server. If my understanding is correct, that information should be used to supply extra data for the ML model to work with?

Do we need to modify the methods that parses the obj back into the json so we can send the RSSI value back to the server?

RadarSDK/RadarBeaconManager.m
```diff
@@ -358,7 +358,23 @@ - (void)locationManager:(CLLocationManager *)manager rangingBeaconsDidFailForReg
...
+                    [[RadarLogger sharedInstance] logWithLevel:RadarLogLevelDebug message:[NSString stringWithFormat:@"Overwriting stale RSSI: %ld", (long)radarBeacon.rssi]];
+                    [radarBeacon setRssi:beacon.rssi];
+                }
+                foundExistingBeacon = YES;
+                *stop = YES;
+            }
+        }];
+
+        if (!foundExistingBeacon) {
+            [self.nearbyBeacons addObject:[RadarBeacon fromCLBeacon:beacon]];
```
ShiCheng-Lu, line 376: you can probably create the RadarBeacon and leverage the [RadarBeacon hash] function to find the beacon in nearby beacons from the set instead of iterating over it. then you can update the existing beacon as needed.

RadarSDK/RadarBeacon.m
gvivan: lines 200-202 in RadarBeacon.m: 
`if (self.rssi) {
    [dict setValue:@(self.rssi) forKey:@"rssi"];
}`

line 322 in RadarAPIClient.m):
`if (beacons) {
    params[@"beacons"] = [RadarBeacon arrayForBeacons:beacons];  // ← Includes RSSI
}
if (indoorSurvey) {
    params[@"indoorSurvey"] = indoorSurvey;  // ← Separate BLE scan data
}`

RadarSDK/Radar.m
```diff
@@ -236,6 +238,27 @@ + (void)trackOnceWithDesiredAccuracy:(RadarTrackingOptionsDesiredAccuracy)desire
                                  }];
                          };
 
+                         void (^performIndoorScanThenTrack)(NSArray<RadarBeacon *> *_Nullable) = ^(NSArray<RadarBeacon *> *_Nullable beacons) {
```
KennyHuRadar, line 241: right now we are only doing this via track once, is this intended? How about when the device is foregrounded and timer based track is initiated?
Is this ok to omit for now?
@lmeier

gvivan: good idea, added

RadarSDK/Radar.m
```diff
@@ -236,6 +238,31 @@ + (void)trackOnceWithDesiredAccuracy:(RadarTrackingOptionsDesiredAccuracy)desire
...
+                                     }];
+                                 } else {
+                                     [[RadarLogger sharedInstance] logWithLevel:RadarLogLevelDebug
+                                                        message:@"RadarSDKIndoors not available, skipping indoor survey"];
+                                     callTrackAPI(beacons, nil);
+                                 }
+                             } else {
+                                 callTrackAPI(beacons, nil);
+                             }
+                         };
```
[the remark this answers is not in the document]
gvivan: could be a good idea but we might be integrating beacon data in indoor survey soon, so will decide at that point

RadarSDK/RadarLocationManager.m
```diff
@@ -955,22 +988,27 @@ - (void)sendLocation:(CLLocation *)location stopped:(BOOL)stopped source:(RadarL
...
-                                                             NSArray<RadarGeofence *> *_Nullable nearbyGeofences, RadarConfig *_Nullable config, RadarVerifiedLocationToken *_Nullable token) {
-                                             self.sending = NO;
-                                             if (status != RadarStatusSuccess || !config) {
-                                                 return;
-                                             }
-
-                                             [self updateTrackingFromMeta:config.meta];
-                                             [self replaceSyncedGeofences:nearbyGeofences];
-                                         }];
+        [self performIndoorSurveyIfPossible:location
```
KennyHuRadar, line 991: this does not seem wrong.
but I also hope to clarify that we are explicitly intending to support this with the deprecated beacon behavior?
And would doing so have down stream impact as we intend to use the known beacon RSSI values in future iterations of the indoors scan?
this is more of a @lmeier question.

RadarSDK/RadarLocationManager.m
```diff
@@ -862,6 +863,33 @@ - (void)handleLocation:(CLLocation *)location source:(RadarLocationSource)source
     [self sendLocation:sendLocation stopped:stopped source:source replayed:replayed beacons:beacons];
 }
 
+- (void)performIndoorSurveyIfPossible:(CLLocation *)location 
+                               beacons:(NSArray<RadarBeacon *> *_Nullable)beacons
+                     completionHandler:(void (^)(NSArray<RadarBeacon *> *_Nullable, NSString *_Nullable))completionHandler {
+    RadarTrackingOptions *options = [Radar getTrackingOptions];
+    Class RadarSDKIndoors = NSClassFromString(@"RadarSDKIndoors");
+    
+    if (options.useIndoorScan && ![RadarSettings isInSurveyMode] && RadarSDKIndoors && [RadarUtils foreground]) {
```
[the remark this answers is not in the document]
gvivan: I verified that it doesn't work in the background when blue bar is on

KennyHuRadar:
make sure to resolve the merge conflict and the CI passes before we merge

PR #456, opened 2025-07-08
add campaignMetadata to userInfo on notification

ShiCheng-Lu:


ShiCheng-Lu:
> Member



> Kk -- did you run this by their team @ShiCheng-Lu ? Would love to have them give a 👍 on this approach before we ship.
> 
> Also will need a version bump.

yep, they just need it at notification handler. Also bumped version

PR #458, opened 2025-07-14
User tags

KennyHuRadar:


RadarSDK/Radar.m
```diff
@@ -156,6 +156,24 @@ + (NSDictionary *_Nullable)getMetadata {
...
+        [Radar trackOnceWithCompletionHandler:nil];
+    }
+}
+
++ (void)removeUserTags:(NSArray<NSString *> *_Nonnull)userTags {
+    [RadarSettings removeUserTags:userTags];
+    if ([RadarSettings sdkConfiguration].syncAfterSetUser) {
+        [Radar trackOnceWithCompletionHandler:nil];
+    }
+}
```
lmeier, line 175: It might be nice to only call `trackOnce` if there's a diff.

KennyHuRadar: to be frank I think we don't really need it, removing for now

RadarSDK/RadarSettings.m
```diff
@@ -374,4 +375,48 @@ + (RadarInitializeOptions *)initializeOptions {
...
+    }
+    
+    [existingTags removeObjectsInArray:tags];
+    
+    if (existingTags.count > 0) {
+        [[NSUserDefaults standardUserDefaults] setObject:existingTags forKey:kUserTags];
+    } else {
+        [[NSUserDefaults standardUserDefaults] removeObjectForKey:kUserTags];
+    }
+}
```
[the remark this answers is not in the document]
KennyHuRadar: Yeah but after thinking about it a little it does not seem necessary. Any updates to the tags are only applicable when we are performing a genuine track. And since we are already pushing these updates at track time, we do not need to perform a track just to sync the track.

The only world where we would want to perform a sync then is when:
- tags are set on server via api
- tags impact client behavior 

At this point neither of the above statements are true.

PR #462, opened 2025-07-30
Improve getTrackingOptions

KennyHuRadar:
There was a bug report about a crash resulting from calling a method on a null pointer. 

Initial investigation suggests a potential race condition with other threads resulting in the `getTrackingOptions` retuning a null reference. 

This patch attempts to mitigate that mode of failure by guarding against the underlying value of the remote tracking option being "changed" underneath the method. Specifically, this code change only makes a single call to `[RadarSettings remoteTrackingOptions]` which ensure the validity of the null check.


RadarSDK/Radar.m
```diff
@@ -453,7 +453,8 @@ + (BOOL)isTracking {
 }
 
 + (RadarTrackingOptions *)getTrackingOptions {
-    return [RadarSettings remoteTrackingOptions] ? [RadarSettings remoteTrackingOptions] : [RadarSettings trackingOptions];
+    RadarTrackingOptions *trackingOptions = [RadarSettings remoteTrackingOptions];
```
lmeier, line 456: call this `remoteTrackingOptions` instead of `trackingOptions`.

ShiCheng-Lu:
3.21.12 is already used, should be 3.21.13

PR #464, opened 2025-08-01
Shicheng/iam

ShiCheng-Lu:


RadarSDK/RadarIAMDelegate.swift
```diff
@@ -0,0 +1,83 @@
...
+//  RadarSDK
+//
+//  Created by ShiCheng Lu on 7/22/25.
+//  Copyright © 2025 Radar Labs, Inc. All rights reserved.
+//
+
+import Foundation
+import SwiftUI
+
+@available(iOS 13.0, *)
```
KennyHuRadar, line 12: given that this implementation can be "overridden" by the user if they so choose to just use the protocol, I think we should not include logic that is almost surely useful if re-used.

Hence, I believe that the logic to download the image from a URL should live in the manager, where it can be re-used instead of the delegate.

RadarSDK/RadarIAMDelegate.swift
```diff
@@ -0,0 +1,83 @@
...
+        request.httpMethod = "GET"
+        
+        if (!url.starts(with: "http")) {
+            let publishableKey = UserDefaults.standard.string(forKey: "radar-publishableKey")!
+            let radarHost = UserDefaults.standard.string(forKey: "radar-host")!
+            request.url = URL(string: "\(radarHost)/v1/assets/\(url)")
+            request.addValue(publishableKey, forHTTPHeaderField: "Authorization")
+        }
+        
+        URLSession.shared.dataTask(with: request) {
```
KennyHuRadar, line 29: totally get that we are writing this in swift and objective C is ugly.
However, it still feels quite messy to have network code directly living here.

Maybe we can consider either
- extending the existing objC api client and helper to interface more nicely with our swift needs
 - create a parallel swift based api client/helper that we can migrate to going forward.

RadarSDK/RadarRouteMode.m
```diff
@@ -1,23 +0,0 @@
-// RadarRouteMode.m
```
KennyHuRadar, line 1: I'm a little confused by these diffs, are they generated by the IDE or a bad merge?
Or is there a reason why we need it in the PR?

RadarSDK/RadarLocationManager.m
```diff
@@ -871,23 +872,29 @@ - (void)sendLocation:(CLLocation *)location stopped:(BOOL)stopped source:(RadarL
 
     RadarTrackingOptions *options = [Radar getTrackingOptions];
     
+    void (^callTrackAPI)(NSArray<RadarBeacon *> *_Nullable) = ^(NSArray<RadarBeacon *> *_Nullable beacons) {
```
KennyHuRadar, line 875: if this an unrelated refactor?

if so can it live in another PR?

RadarSDK/RadarIAMDelegate.m
```diff
@@ -0,0 +1,43 @@
...
+RadarInAppMessageDelegate_Swift* radarIAMDelegate = nil;
+
+- (instancetype) init {
+    if (radarIAMDelegate == nil) {
+        radarIAMDelegate = [[RadarInAppMessageDelegate_Swift alloc] init];
+    }
+    return self;
+}
+
+- (void)createInAppMessageView:(RadarInAppMessage * _Nonnull)message completionHandler:(nonnull void (^)(UIViewController * _Nonnull __strong))completionHandler {
```
KennyHuRadar, line 27: in Android, I just have it such that the manager extracts the image if it exists and passes it to the view creation method

RadarSDK/Radar.m
```diff
@@ -1287,34 +1306,39 @@ + (NSString *)stringForLocationSource:(RadarLocationSource)source {
 }
 
 + (NSString *)stringForMode:(RadarRouteMode)mode {
-    return [RadarRouteModeUtils stringForMode:mode];
+    switch (mode) {
```
KennyHuRadar, line 1309: unrelated refactor?

KennyHuRadar (commented):
we should also add some unit tests.
on android, my unit tests essentially mocked the response from track to include the IAM payload and I asserted that the delegate methods were called.

RadarSDK/RadarIAMDelegate.swift
ShiCheng-Lu: We can always have the image load as a function that they can call to return the image from Radar servers. But I think it's still best to have to option to not load the image even if it exist.

KennyHuRadar: that is fair

RadarSDK/RadarIAMManager.swift
```diff
@@ -0,0 +1,101 @@
...
+              let message = currentMessage else {
+            return
+        }
+        
+        let messageClickTime = Date()
+        let duration = messageClickTime.timeIntervalSince(messageShownTime)
+        
+        var metadata: [String: Any] = [:]
+        metadata["display_duration"] = duration
+        metadata["campaign_id"] = message.metadata["campaign_id"] as? String
```
KennyHuRadar, line 38: the server is expecting `campaignId`

PR #476, opened 2025-09-10
Add more logging for CSGN lifecycle 

KennyHuRadar:


KennyHuRadar:
not sure if we want to merge this, (prob not)
just looking for eyes before we cut a beta.

RadarSDK/RadarLocationManager.m
```diff
@@ -620,6 +626,7 @@ - (void)replaceSyncedGeofences:(NSArray<RadarGeofence *> *)geofences {
 
 - (void)removeSyncedGeofences {
     for (CLRegion *region in self.locationManager.monitoredRegions) {
+        [[RadarLogger sharedInstance] logWithLevel:RadarLogLevelDebug message:[NSString stringWithFormat:@"Encountering monitored geofence | identifier = %@", region.identifier]];
```
lmeier, line 629: this and line 255 of notification helper would be too noisy in logs. but the monitoringDidFail logging should be brought in to the main sdk certainly. is there a registeringNotificationDidFail as well? would be good

PR #489, opened 2025-10-13
FENCE-2308: Add batching options to tracking options

alanjcharles14:
- Adds `batchInterval` and `batchSize` to `trackingOptions` 
- Adds batch handling logic to `RadarReplayBuffer` 
- Adds batch check to `RadarAPIClient`


RadarSDK/RadarLocationManager.m
```diff
@@ -1053,7 +1053,7 @@ - (void)locationManager:(CLLocationManager *)manager didUpdateLocations:(NSArray
     }
 
     CLLocation *location = [locations lastObject];
-    if (self.completionHandlers.count) {
+    if (self.completionHandlers.count && [RadarUtils foreground]) {
```
lmeier: This is the most significant change. I  _do_ think we want to do this — it's weird for us to be calling, e.g. locations when the app is bluebar-ed as foregound — but we should be cautious of downstream impacts.

@nickpatrick expressed some hesitancy about this — wonder Nick if you have further thoughts?

Per our discussion in today's meeting, we could make this

```
    if (self.completionHandlers.count && (sdkConfiguration.skipForegroundCheck || [RadarUtils foreground])) {
```

where `skipForegroundCheck` defaults to true and we roll it off via a FeatureSetting (see server repo).

lmeier: @alanjcharles14 I think to ship this we need the `sdkConfiguration.skipForegroundCheck` that's default on.

We can a cut a beta in the meantime, or just wait until we do that (and get in in tomorrow's server release).

lmeier (commented):
Just need an `sdkConfiguration` and then we're g2g.

lmeier (approved):

PR #490, opened 2025-10-14
FENCE-2531: Pressure logging

lmeier:
- Adds collection of `motionAuthorization` status
- Adds more logging for pressure and altitude measuring


lmeier (commented):
maybe we change the request key (other things up to you) to `motionAuthorization` instead of `motionAuthorizationStatus` but otherwise lgtm

PR #504, opened 2025-11-17
use Env variables for output

ShiCheng-Lu:
https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/
https://github.blog/changelog/2023-07-24-github-actions-update-on-save-state-and-set-output-commands/

Github are deprecating these, so before it's broken, we should update it to use the environment files

Warnings during publish: https://github.com/radarlabs/radar-sdk-ios/actions/runs/19439889806

! monitor the next release to make sure it works.

### DIFF
--- a/.github/workflows/release-sdk.yml
+++ b/.github/workflows/release-sdk.yml
@@ -85,15 +85,15 @@ jobs:
       # get sha256 checksum of the XCFramework
       - name: Get SHA256 checksum (RadarSDK)
         id: checksum_radarsdk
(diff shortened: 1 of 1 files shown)


stevepopovich:
Did you checkout Copilot's comment? Looks like it could be correct 🤔

ShiCheng-Lu:
> Did you checkout Copilot's comment? Looks like it could be correct 🤔

Yeah I think the PR is wrong. but I remember following some migration guide, but I'm probably blind or the guide was wrong

stevepopovich:
👍🏼 Should we close then?

PR #527, opened 2026-01-22
FENCE-2607: Reduce Track Calls

alanjcharles14:
## Summary
Adds a server-driven sync region system that reduces `/v1/track` API calls by performing local event detection on-device. When enabled via server config (`useSyncRegion: true`), the SDK fetches nearby entities from a new `POST /v1/sync/region` endpoint and only makes `/v1/track` calls when it detects a geofence entry/exit/dwell, place change, or beacon change locally.

## New SDK Components

**`RadarTrackingOptionsSyncEvents`** — New sync mode (`sync: "events"`) that enables local state-change suppression of track calls.

**`RadarTrackingOptionsType`** — New enum (`default`, `onTrip`, `inGeofence`, `isUser`) parsed from the server's tracking options response, indicating which set of tracking options is currently active.

**`RadarSyncManager`** — Handles the sync region lifecycle and local event detection:
- Fetches sync region on startup and on a repeating timer (6 hours)
- Triggers an initial fetch on the first location update if no synced region exists yet
- Refreshes when the user approaches the region boundary (within 20%)
- Performs local containment checks for circle geofences (point-in-circle) and polygon geofences (ray-casting with geodesic distance-to-edge)
- Supports buffered entries/exits using `horizontalAccuracy`
- Supports stop detection gating (project-level and per-geofence)
- Supports dwell threshold detection (project-level and per-geofence)
- Only suppresses `/v1/track` calls when `syncLocations == .events` and `useSyncRegion` is enabled

## How It Works

1. Server config sets `useSyncRegion: true` when any of a project's tracking option sets (`defaultTrackingOptions`, `onTripTrackingOptions`, or `inGeofenceTrackingOptions`) use `sync: "events"`
2. SDK calls `POST /v1/sync/region` to fetch nearby geofences, places, beacons, and a bounding region
3. On location updates, `RadarSyncManager.shouldTrack` determines if a `/v1/track` call is needed:
   - Always track if outside the synced region or no region exists
   - Always track if on an active trip *unless* the server sent explicit `onTripTrackingOptions` with `sync: "events"` (determined by `options.type == .onTrip`)
   - Otherwise, compare current local state (geofence containment, place proximity, beacon range) against last known state
   - Only track if an entry, exit, or dwell event is detected
   - If the user nears the region boundary, the sync region is proactively refreshed
4. `trackOnce` calls bypass sync region entirely — they go directly to the API without passing through the `shouldTrack` gate


RadarSDKTests/RadarSettingsTest.swift
```diff
@@ -81,7 +81,7 @@ actor RadarSettingsTest {
         #expect(UserDefaults.standard.string(forKey: "radar-appGroup") == "test.app.group")
         
         #expect(RadarSettings.userId == "test")
-        #expect(RadarSettings.getAppGroup() == "test.app.group")
+        #expect(RadarSDK.RadarSettings.getAppGroup() == "test.app.group")
```
ShiCheng-Lu: ?

RadarSDKTests/RadarSyncManagerTests.swift
```diff
@@ -0,0 +1,608 @@
...
+//  RadarSDK
+//
+//  Created by Alan Charles on 3/19/26.
+//  Copyright © 2026 Radar Labs, Inc. All rights reserved.
+//
+
+import XCTest
+@testable import RadarSDK
+
+class RadarSyncManagerTests: XCTestCase {
```
ShiCheng-Lu: did claude do this? use SwiftTesting instead of XCTest

RadarSDK/RadarSdkConfiguration.h
```diff
@@ -39,6 +39,16 @@ NS_ASSUME_NONNULL_BEGIN
...
 
+@property (nonatomic, assign) BOOL useSyncRegion;
+
+@property (nonatomic, assign) NSInteger defaultGeofenceDwellThreshold;
+
+@property (nonatomic, assign) BOOL bufferGeofenceEntries;
+
+@property (nonatomic, assign) BOOL bufferGeofenceExits;
+
+@property (nonatomic, assign) BOOL stopDetection;
```
ShiCheng-Lu: I think it might be worth spliting this out into it's own tracklite TM configuration, which can be returned with the sync endpoint.

RadarSDK/RadarBeacon.swift
```diff
@@ -0,0 +1,82 @@
...
+//  RadarSDK
+//
+//  Created by Alan Charles on 3/19/26.
+//  Copyright © 2026 Radar Labs, Inc. All rights reserved.
+//
+
+import Foundation
+import CoreLocation
+
+public struct RadarBeaconSwift: Codable, Sendable {
```
ShiCheng-Lu: all of these should be internal (or just remove the public as internal is default).

RadarSDKTests/RadarSettingsTest.swift
alanjcharles14: I had to do this because I added `#import "RadarSettings.h"` to the test bridging header so that `RadarSyncManagerTests.swift` could access the ObjC `RadarSettings` class. That created an ambiguity as the Swift test target now had two things called RadarSettings — the ObjC class (via the bridging header) and the Swift class (via @testable import RadarSDK). Prefixing with `RadarSDK.RadarSettings` explicitly tells Swift to use the Swift version from the RadarSDK module.

RadarSDKTests/RadarSyncManagerTests.swift
alanjcharles14: ah yeah good catch- i ported these over when i re-did the SyncManager in swift from obj c and just didn't even think to change it.

RadarSDK/RadarBeacon.swift
alanjcharles14: another good catch i had made them public to use them in the `MapView` when testing.

RadarSDKTests/RadarSettingsTest.swift
ShiCheng-Lu: does RadarSyncManagerTests still need RadarSettings.h? or is it just an artifact of the objc test? couldn't the test just access the swift version?

ShiCheng-Lu: you can use `RadarSettings.sdkConfiguration = config` in the tests instead

ShiCheng-Lu: for swift testing, you can mark the tests are serialized, like in RadarSettings test, or if you're up for some work, you can make RadarSettings mockable by adding an interface and make it injectable. Maybe easier would be to mock RadarUserDefaults instead since it only has 4 methods, and make the RadarSettings's RadarUserDefaults injectable.

alanjcharles14: yeah all fair- i've removed it from the bridging header and used this pattern in my tests `RadarSettings.sdkConfiguration = config`. The tests are marked as serialized already (in my last commit) so i think we can make things mockable in the future but it's perhaps a bit out of scope here. Thoughts?

RadarSDK/RadarSdkConfiguration.h
alanjcharles14: yea probably not a bad idea. I think it's probably fine for now but something we could follow up with. Do you have a particularly strong opinion for doing this now? Happy to consider it but would like to get v1 of this out without relying on another server deploy, if possible.

ShiCheng-Lu: yeah, it's fine

lmeier (approved):
🚢 🚢 🚢

PR #552, opened 2026-03-18
Fix notification diff

ShiCheng-Lu:
this new notification helper gets the list of pending notifications from the list instead of assuming any notifications registered will exist.

Also sdkConfig in swift.


RadarSDK/RadarNotificationHelper.swift
```diff
@@ -0,0 +1,263 @@
...
+        let notificationCenter = UNUserNotificationCenter.current()
+        
+        // if notifications is not null, we update the pending notifications, otherwise we only update the registered notifications list
+        if let notifications {
+            // remove all geofence notifications
+            let requests = await notificationCenter.pendingNotificationRequests()
+            if Task.isCancelled {
+                return
+            }
+            let notificationIdentifiersToRemove = requests.compactMap {
```
lmeier: Related to registration silently failing, might we wanna also create a list of notifications that we're not touching (aka quietly occupying slots)? At least to log?

lmeier (commented):
Would love to see:
- some tests (esp for concurrent `trackOnce` — might need to invest in good mocking of the track response)
- an articulation of our rollout plan
- a manual QA checklist (esp w/ edge cases)

lmeier (approved):
Be careful w/ the merge, given we change `RadarSDKConfiguration` in your bug fix PR. But LGTM, this is safe to roll out

PR #563, opened 2026-04-06
Fix notification diff on replay

ShiCheng-Lu:
removes the "sent" notification diffs from registered, as they'll be saved in replay

tested on simulator.

* project, turn replay for tracking options to all.

* sync down notitifcations
* list registered + pending
* simulate a send by removing one of the notifications (or walk around)
* turn off wifi
* track once
* track should fail due to network error
* registered notifications should be updated to be only the two remaining.

### DIFF
--- a/RadarSDK/RadarAPIClient.m
+++ b/RadarSDK/RadarAPIClient.m
@@ -443,7 +443,7 @@
                                                                 source:source
                                                             verified:verified
                                                         publishableKey:publishableKey
-                                                notificationsRemaining:notificationsDelivered
+                                                notificationsRemaining:notificationsRemaining
                                                 locationMetadata:locationMetadata
                                                     completionHandler:completionHandler];
         }];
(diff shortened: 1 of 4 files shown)


RadarSDK/RadarAPIClient.m
```diff
@@ -512,8 +512,21 @@ - (void)makeTrackRequestWithParams:(NSDictionary *)params
...
+                                             completionHandler:^() {}
+                                            ];
+                                        } else {
+                                            [[RadarLogger sharedInstance] logWithLevel:RadarLogLevelInfo message:[NSString stringWithFormat:@"Setting %lu notifications remaining", (unsigned long)notificationsRemaining.count]];
+                                            [RadarState setRegisteredNotifications:notificationsRemaining];
+                                        }
+                                    } else {
+                                        [[RadarLogger sharedInstance] logWithLevel:RadarLogLevelInfo message:[NSString stringWithFormat:@"Setting %lu notifications remaining", (unsigned long)notificationsRemaining.count]];
+                                        [RadarState setRegisteredNotifications:notificationsRemaining];
+                                    }
```
ShiCheng-Lu, line 528: this just uses the new update for registered notification in case of replay if < iOS 13 and feature setting not on.
 
else use old `setRegisteredNotifications:notificationsRemaining`

RadarSDK/RadarNotificationHelper.swift
```diff
@@ -171,6 +171,24 @@ actor RadarNotificationHelper: NSObject {
...
+        }
+        if isRegistering {
+            return
+        }
+        let idsToRemove = notifications.compactMap { $0["identifier"] as? String }
+        
+        registered.removeAll { notification in
+            idsToRemove.contains { $0 == notification.identifier }
+        }
+        radarState.registeredNotifications = registered
```
ShiCheng-Lu, line 196: remove registered notifications that were "sent" (saved into replay)

RadarSDK/RadarAPIClient.m
```diff
@@ -443,7 +443,7 @@ - (void)trackWithLocation:(CLLocation *_Nonnull)location
                                                                 source:source
                                                             verified:verified
                                                         publishableKey:publishableKey
-                                                notificationsRemaining:notificationsDelivered
+                                                notificationsRemaining:notificationsRemaining
```
lmeier, line 446: Yea this is the long-lingering oddity that has probably messed stuff up.

Just wanna confirm that you've thought this through, as this should be non-trivial

ShiCheng-Lu: yes this is right, I don't think the original worked correctly with replays. Regardless, the new version won't be using this path.

RadarSDK/RadarAPIClient.m
```diff
@@ -512,8 +512,21 @@ - (void)makeTrackRequestWithParams:(NSDictionary *)params
...
 
-                                    [[RadarLogger sharedInstance] logWithLevel:RadarLogLevelInfo message:[NSString stringWithFormat:@"Setting %lu notifications remaining", (unsigned long)notificationsRemaining.count]];
-                                    [RadarState setRegisteredNotifications:notificationsRemaining];
+                                    if (@available(iOS 13.0, *)) {
+                                        if ([RadarSettings sdkConfiguration].useNotificationDiffV2) {
+                                            [[RadarNotificationHelper_Swift shared]
+                                             removeRegisteredNotificationsWithNotifications:params[@"notificationDiff"]
+                                             completionHandler:^() {}
+                                            ];
+                                        } else {
```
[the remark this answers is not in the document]
ShiCheng-Lu: no, remove registered notifications will run in the actor, which is isolated. and since theres no suspense points, it'll run to completion.

lmeier (approved):

PR #564, opened 2026-04-06
Use ssh for submodules

lmeier:


lmeier:
Ehhh, maybe this is not the best idea per: https://github.com/radarlabs/radar-sdk-ios/pull/564#discussion_r3040534289

Closing.

PR #568, opened 2026-04-15
FENCE-2717: Add offline event detection and RTO ramping 

alanjcharles14:
## Summary

Introduces `RadarOfflineEventManager` — a client-side system for generating synthetic geofence events and switching tracking options when the device is offline. This is gated by two independent server-controlled flags: `offlineEventGenerationEnabled` (for synthetic events) and `useOfflineRTOUpdates` (for tracking option switching). Also adds `RadarRemoteTrackingOptions` to parse and look up the array of tagged tracking presets sent by the server.

## New SDK Components

**`RadarOfflineEventManager`** — Generates synthetic geofence entry/exit events by evaluating the device's location against locally cached geofence data from the sync region. On track failure, fires events via the delegate callback and optionally switches tracking options based on geofence tag matching. Builds a synthetic `RadarUser` using the cached user identity from `RadarState`.

**`RadarRemoteTrackingOptions`** — Parses the `remoteTrackingOptions` array from `sdkConfiguration`. Each entry has a `type` (`default`, `onTrip`, `inGeofence`), a `trackingOptions` preset, and optional `geofenceTags`. Provides lookup helpers to retrieve tracking options or geofence tags by type.

**`RadarOfflineEventManager.h`** — Obj-C header exposing `RadarOfflineEventManager` methods to `RadarAPIClient.m`, replacing the previous `NSClassFromString`/`performSelector` pattern with direct method calls.

**`RadarLocationSourceOffline`** — New location source enum value (`OFFLINE_DETECTION`) to identify events and users generated by the offline event manager.

## How It Works

1. **Server sends configuration**: `offlineEventGenerationEnabled` and `useOfflineRTOUpdates` arrive in `sdkConfiguration`. When `useOfflineRTOUpdates` is true, the server also sends a `remoteTrackingOptions` array with tagged presets. These are parsed by `RadarSdkConfiguration` and `RadarRemoteTrackingOptions`.

2. **Track failure triggers offline logic** (`RadarAPIClient.m`): When a track call fails, the API client calls `handleTrackFailure:` (gated by `offlineEventGenerationEnabled`) and `updateTrackingOptionsFor:` (gated by `useOfflineRTOUpdates`). These are now direct method calls via the new Obj-C header.

3. **Event generation** (`generateEvents`): Compares the current location against synced geofences from `RadarSyncManager`. Detects entries and exits relative to the previously known geofence set. Builds synthetic `RadarEvent` objects with `metadata: {offline: true}` and delivers them via `didReceiveEvents` delegate callback.

4. **Synthetic user** (`buildSyntheticUser`): Retrieves the cached `RadarUser` from `RadarState` (persisted on every successful track) and constructs a new user with updated location, geofences, and source (`OFFLINE_DETECTION`), while preserving identity fields (`_id`, `userId`, `deviceId`) and geographic context (`country`, `state`, `dma`, `postalCode`, `trip`).

5. **Tracking option switching** (`updateTrackingOptions`): Checks if the user is in a geofence whose tag matches the `inGeofence` remote tracking option's `geofenceTags`. If matched, ramps up to in-geofence tracking options. Otherwise falls back to on-trip (if active) or default tracking options.

6. **Sync region re-fetch** (`RadarLocationManager.m`): When `offlineEventGenerationEnabled` is true, proactively re-fetches the sync region if the device moves outside or near the boundary of the current synced region, ensuring fresh geofence data for offline event evaluation.

7. **User caching** (`RadarState`): Added `setRadarUser:`/`radarUser` to persist the full `RadarUser` to `NSUserDefaults` on every successful track response. Exposed to Swift via the `RadarSwiftBridge` protocol.

## Test Coverage

Unit tests in `RadarOfflineEventManagerTests.swift` covering:
- `handleTrackFailure` gating by `offlineEventGenerationEnabled`
- `updateTrackingOptions` geofence tag matching and fallback logic
- `generateEvents` entry/exit detection and empty-result cases
- `reset` clearing offline geofence state


RadarSDKTests/RadarSyncManagerTests.swift
```diff
@@ -10,858 +10,860 @@ import Foundation
...
-                "coordinates": [testLng, testLat]
-            ],
-            "geofences": [] as [[String: Any]],
-            "beacons": [] as [[String: Any]]
-        ]
-        if let placeId = placeId {
-            dict["place"] = [
-                "_id": placeId,
-                "name": "Test Place",
+extension RadarSerializedTests {
```
stevepopovich, line 13: Hmm this diff is a bit tricky to review, do you mind marking which bits of the file are formatting/unchanged and which are the new tests to review? Sorry 😅

alanjcharles14: ah yeah- nothing in this file changed besides adding the `extension` at the top (and reformatting the rest to nest inside of it). I had to do this because this file and `RadarOfflineEventManagerTests` both alter the the `syncStore` so i had to force them to run one at a time. Not sure if there's a better way, it is a bit weird.

offline-events.md
```diff
@@ -0,0 +1,37 @@
+## Summary
+
+Introduces `RadarOfflineEventManager` — a client-side system for generating synthetic geofence events and switching tracking options when the device is offline. This is gated by two independent server-controlled flags: `offlineEventGenerationEnabled` (for synthetic events) and `useOfflineRTOUpdates` (for tracking option switching). Also adds `RadarRemoteTrackingOptions` to parse and look up the array of tagged tracking presets sent by the server.
```
lmeier: First thought: is there reason to simplify by removing `useOfflineRTOUpdates` and always sending down the full set of RTO? I don't imagine people not wanting to have RTO work when offline.

offline-events.md
```diff
@@ -0,0 +1,37 @@
...
+
+## New SDK Components
+
+**`RadarOfflineEventManager`** — Generates synthetic geofence entry/exit events by evaluating the device's location against locally cached geofence data from the sync region. On track failure, fires events via the delegate callback and optionally switches tracking options based on geofence tag matching. Builds a synthetic `RadarUser` using the cached user identity from `RadarState`.
+
+**`RadarRemoteTrackingOptions`** — Parses the `remoteTrackingOptions` array from `sdkConfiguration`. Each entry has a `type` (`default`, `onTrip`, `inGeofence`), a `trackingOptions` preset, and optional `geofenceTags`. Provides lookup helpers to retrieve tracking options or geofence tags by type.
+
+**`RadarOfflineEventManager.h`** — Obj-C header exposing `RadarOfflineEventManager` methods to `RadarAPIClient.m`, replacing the previous `NSClassFromString`/`performSelector` pattern with direct method calls.
+
+**`RadarLocationSourceOffline`** — New location source enum value (`OFFLINE_DETECTION`) to identify events and users generated by the offline event manager.
```
lmeier: I don't think this is really a `source`: `source` moreso describes what part of the location manager we're getting the location from (e.g. `GEOFENCE_EXIT`)

offline-events.md
```diff
@@ -0,0 +1,37 @@
...
+
+## How It Works
+
+1. **Server sends configuration**: `offlineEventGenerationEnabled` and `useOfflineRTOUpdates` arrive in `sdkConfiguration`. When `useOfflineRTOUpdates` is true, the server also sends a `remoteTrackingOptions` array with tagged presets. These are parsed by `RadarSdkConfiguration` and `RadarRemoteTrackingOptions`.
+
+2. **Track failure triggers offline logic** (`RadarAPIClient.m`): When a track call fails, the API client calls `handleTrackFailure:` (gated by `offlineEventGenerationEnabled`) and `updateTrackingOptionsFor:` (gated by `useOfflineRTOUpdates`). These are now direct method calls via the new Obj-C header.
```
lmeier: cc @stevepopovich — just to be aware of this new codepath coming in given you're also working on failures

offline-events.md
```diff
@@ -0,0 +1,37 @@
...
+
+3. **Event generation** (`generateEvents`): Compares the current location against synced geofences from `RadarSyncManager`. Detects entries and exits relative to the previously known geofence set. Builds synthetic `RadarEvent` objects with `metadata: {offline: true}` and delivers them via `didReceiveEvents` delegate callback.
```
lmeier: One could imagine introducing a new offline event delegate `didReceiveOfflineEvents` — advantage would be that we could turn this on for people w/o worrying about messing w/ their delegate. But using the same delegate also has upside... just pointing out.

offline-events.md
```diff
@@ -0,0 +1,37 @@
...
+
+4. **Synthetic user** (`buildSyntheticUser`): Retrieves the cached `RadarUser` from `RadarState` (persisted on every successful track) and constructs a new user with updated location, geofences, and source (`OFFLINE_DETECTION`), while preserving identity fields (`_id`, `userId`, `deviceId`) and geographic context (`country`, `state`, `dma`, `postalCode`, `trip`).
```
lmeier: The geographic context is likely but not certainly still correct. Probably fine.

offline-events.md
alanjcharles14: hmmm not a bad idea. Racking my brain trying to think of a reason not to do this and not really coming up with much other than maybe cost but even that's not a great reason since they're coming down in the config anyway. Do you remember the intention behind this flag in the first place? Like- was there a reason we did it besides the offline event stuff that would be generated automatically in the old implementation?

offline-events.md
alanjcharles14: fair- i just pulled this from the old PR. Should we scrap it? Hard to say how people are currently using it if at all since we also include `offline` in the event metadata, but was kind of trying to not have customers do any re-implementation if we can help it.

offline-events.md
alanjcharles14: yeahhhh fair point...was just another thing i pulled from the old PR for parity's sake but can give it another think if it's going to cause any problems

offline-events.md
alanjcharles14: ohh i find this quite appealing from an architecture perspective but also it's pretty straightforward for a customer to filter for `offline` events in the existing event delegate since we're slapping the `offline` key in the metadata for every generated event.

alanjcharles14: i could go either way, I think

offline-events.md
lmeier: Is there a compelling reason to maintaining this geographic context? I think it makes sense to remove until we can do it correctly.

offline-events.md
lmeier: Really it's just used by us in analyzing locations.

But yeah, I'd scrap it — it's not the right way of indicating that the event is an offline event (metadata is better).

offline-events.md
lmeier: yea, I don't feel super strongly here. let's go with the existing approach

offline-events.md
alanjcharles14: okay cool- removed for now

offline-events.md
alanjcharles14: sounds like a plan- removed it

stevepopovich:
This is def out of scope for this PR, just looking for some more clarification:

We are creating synthetic events? Can we not just save a timestamp and the location events we got from the OS, batch them up and send them when we get back online? Then we don't have to worry if they are "real events" or not.

BUT I am probably missing some context 🤔

alanjcharles14:
@stevepopovich Yeah, it's a good question. We are already doing that, actually (or at least have the functionality in the SDK to do that if a user has `replay: all` in their remote tracking options). There are other reasons a user might want the synthetic events on the client when the device is offline. For instance, you might want to send a notification if a user is in a particular geofence or change the state of your app in some way.

stevepopovich:
Gotcha, thanks for the explanation!

RadarSDKTests/RadarSyncManagerTests.swift
stevepopovich: no, I think this is the way, just yields an unwieldy diff, makes sense. Thanks!

stevepopovich (commented):
I personally don't see any red flags, though I may not be the most well equipped reviewer!

RadarSDK/RadarState.m
```diff
@@ -305,4 +306,18 @@ + (void)setAltitudeAdjustments:(NSArray<NSDictionary *> *_Nullable)altitudeAdjus
...
+        [[NSUserDefaults standardUserDefaults] setObject:radarUserDict forKey:kRadarUser];
+    } else {
+        [[NSUserDefaults standardUserDefaults] removeObjectForKey:kRadarUser];
+    }
+}
+
++ (RadarUser *_Nullable)radarUser {
+    NSDictionary *radarUserDict = [[NSUserDefaults standardUserDefaults] objectForKey:kRadarUser];
+    return [[RadarUser alloc] initWithObject:radarUserDict];
+}
```
lmeier, line 324: I think we might wanna nil guard after line 319

RadarSDKTests/RadarOfflineEventManagerTests.swift
```diff
@@ -0,0 +1,314 @@
...
+        func handleTrackFailure_nilConfig() {
+            RadarSettings.sdkConfiguration = nil
+            
+            let location = CLLocation(latitude: testLat, longitude: testLng)
+            RadarOfflineEventManager.handleTrackFailure(location)
+        }
+        
+        // MARK: - updateTrackingOptions
+        
+        @Test("updateTrackingOptions returns inGeofence options when tags match")
```
lmeier, line 119: I think we need a test for on-trip tracking options

offline-events.md
lmeier, line 1: reminder to remove before merging

lmeier (commented):
a couple of minor things. once these are resolved, it lgtm

stevepopovich (approved):
Second pass makes a lot of sense to me!

PR #570, opened 2026-04-16
Propagate NSError through RadarAPICompletionHandler

stevepopovich:
https://linear.app/radarlabs/issue/FENCE-2756/sdk-support-for-alternative-hosts

[ADR for failover hosts in the SDKS](https://www.notion.so/SDK-Failover-ADR-344e42b18f96800fa6bdf8ecb9828afe?source=copy_link)

## Summary
- Adds `NSError` parameter to `RadarAPICompletionHandler` so callers can inspect specific `NSURLError` codes, this will be needed for the `RadarHostFailover` class so it knows when to try the failover host.
- Updates all callsites in `RadarAPIClient` and `RadarAPIHelper` to pass the error through
- Foundation for host failover logic in follow-up PRs
- Add a basic test verifying the error gets passed through
- It is worth noting that we are only applying a failover to the verified endpoint for now, so that is why this is only set up for that right now.

## Stack
1. **This PR**
2. #571 Add RadarHostFailover class with exponential backoff
3. #572 Wire up host failover + feature flag

## Test plan
No functional changes yet, just regression tests
- [ ] Verify existing API calls still work end-to-end

### DIFF
--- a/RadarSDK/RadarAPIClient.m
+++ b/RadarSDK/RadarAPIClient.m
@@ -139,7 +139,7 @@
                                 sleep:NO
                            logPayload:YES
                       extendedTimeout:NO
-                    completionHandler:^(RadarStatus status, NSDictionary *_Nullable res) {
+                    completionHandler:^(RadarStatus status, NSDictionary *_Nullable res, NSError *_Nullable error) {
                         if (!res) {
                             completionHandler(status, nil);
                             return;
@@ -176,7 +176,7 @@
                                 sleep:NO
                            logPayload:NO
                       extendedTimeout:YES
-                    completionHandler:^(RadarStatus status, NSDictionary *_Nullable res) {
+                    completionHandler:^(RadarStatus status, NSDictionary *_Nullable res, NSError *_Nullable error) {
                             id eventsObj = res[@"events"];
                             id userObj = res[@"user"];
 
@@ -505,7 +505,7 @@
                                     sleep:YES
                             logPayload:YES
                         extendedTimeout:NO
-                        completionHandler:^(RadarStatus status, NSDictionary *_Nullable res) {
+                        completionHandler:^(RadarStatus status, NSDictionary *_Nullable res, NSError *_Nullable error) {
                             if (status != RadarStatusSuccess || !res) {
                                 if (options.replay == RadarTrackingOptionsReplayAll) {
                                     // create a copy of params that we can use to write to the buffer in case of request failure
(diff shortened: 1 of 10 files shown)


RadarSDK/RadarAPIHelper.m
```diff
@@ -138,7 +138,7 @@ - (void)requestWithMethod:(NSString *)method
                 id resObj = [NSJSONSerialization JSONObjectWithData:data options:0 error:&deserializationError];
                 if (deserializationError || ![resObj isKindOfClass:[NSDictionary class]]) {
                     dispatch_async(dispatch_get_main_queue(), ^{
-                        completionHandler(RadarStatusErrorServer, nil);
+                        completionHandler(RadarStatusErrorServer, nil, nil);
```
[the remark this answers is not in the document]
stevepopovich: not needed!

RadarSDK/RadarAPIClient.m
```diff
@@ -1083,7 +1083,7 @@ - (void)searchGeofencesNear:(CLLocation *_Nonnull)near
                                 sleep:NO
                            logPayload:YES
                       extendedTimeout:NO
-                    completionHandler:^(RadarStatus status, NSDictionary *_Nullable res) {
+                    completionHandler:^(RadarStatus status, NSDictionary *_Nullable res, NSError *_Nullable error) {
```
nickdonnelly, line 1086: For all these callbacks that have the shape `(RadarStatus, Result dict, Error)` maybe we can alias them to something more readable/globally modifiable.

```objc
typedef void (^RadarCallback)(RadarStatus status, NSDictionary * _Nullable res, NSError * _Nullable error);
```

RadarSDK/RadarSettings.swift
```diff
@@ -12,6 +12,7 @@ internal class RadarSettings: NSObject {
     
     static let DefaultHost = "https://api.radar.io"
     static let DefaultVerifiedHost = "https://api-verified.radar.io"
+    static let DefaultVerifiedHostFallback = "https://api-verified2.radar.io"
```
nickdonnelly, line 15: ~Why do we have an `api-verified2`?~
Nevermind.

stevepopovich: it is the fallback/failover URL

RadarSDK/RadarAPIClient.m
stevepopovich: not a bad idea!

stevepopovich:
I am going to close this since the implementation is changing significantly, see https://radarlabs.slack.com/archives/C03LAGZSUR0/p1776867988460529

PR #573, opened 2026-04-16
Add CLAUDE.md with Swift language policy, properly ignore Claude local settings

stevepopovich:
## Summary

- Adds `CLAUDE.md` — the Claude Code agent configuration file for this repo
- Establishes that all new code should be written in Swift
- Requires migrating Objective-C files to Swift when making non-trivial changes to them
- Documents build/test commands and project structure for AI-assisted development
- The claude settings file should be gitignored too, so just doing that too. Those are settings for each user on their machine


alanjcharles14:
I'm a bit hesitant about this line: 
`Existing Objective-C files → migrate to Swift when touching them for non-trivial changes`

I'm not sure what claude will consider non-trivial and I don't think we want to re-write things like RadarLocationManager or RadarApiClient on the fly- we'll need to be super intentional when we do re-write those.

stevepopovich:
That's totally fair! I changed the language to get the agent to ask before migrating files, wdyt?

alanjcharles14:
lgtm!

alanjcharles14 (approved):

PR #576, opened 2026-04-23
FENCE-2756: Verified-host failover on exceptions and non-radar errors

stevepopovich:
## Summary

- Adds transparent failover for the verified host, gated behind the remotely-togglable `useVerifiedHostFailover` SDK-config flag (default `false`). When enabled and requests to `https://api-verified.radar.io` return a response that does **not** look like it came from a Radar server (**transport error, or an HTTP response whose body is missing a top-level `meta` key**), the SDK retries once against the secondary host and remembers to prefer secondary for the next 60 seconds.
- After 60 seconds the next verified request probes primary again. On a primary success, the SDK returns to primary. On a primary probe failure, failover resumes and the timer resets. If the secondary also fails, the 60s timer is cleared so the very next request immediately probes primary again (avoids hammering a dead secondary).
- Scope is verified requests only: `getConfigForUsage:verified:YES` and `makeTrackRequestWithParams:...verified:YES` in [RadarSDK/RadarAPIClient.m](RadarSDK/RadarAPIClient.m). Non-verified requests are untouched, and when the flag is off these two call sites fall through to the pre-branch direct request against `[RadarSettings verifiedHost]`.
- Primary host: `https://api-verified.radar.io`. Secondary host: `https://api-verified.radar.com` (`DefaultVerifiedHostSecondary` in [RadarSDK/RadarSettings.swift](RadarSDK/RadarSettings.swift)).

## Implementation

- New [`RadarFailoverHostSelector`](RadarSDK/RadarFailoverHostSelector.swift) — state machine tracking the current host index + next-probe time over an ordered N-host list. Injectable clock for tests.
- New [`RadarFailoverAPICoordinator`](RadarSDK/RadarFailoverAPICoordinator.swift) — fronts `RadarAPIHelper` via a closure-injected request function so it doesn't need to see the ObjC type from Swift. Takes a `hostsProvider` closure and exposes a `verifiedSharedInstance` pre-configured with the two verified hosts, so the same failover machinery can be reused for other hosts later. Classifies each response using presence of top-level `meta` and reports back to the selector.
- Minimal ObjC edits in [RadarSDK/RadarAPIClient.m](RadarSDK/RadarAPIClient.m): the two verified branches are gated on `verified && [RadarSettings sdkConfiguration].useVerifiedHostFailover` — when true they hand a path + request-execution block to the coordinator; when false they issue a direct request against `[RadarSettings verifiedHost]`.
- `useVerifiedHostFailover` added to `RadarSdkConfiguration` alongside the other remote flags so it can be flipped from the server-side config.
- `DefaultVerifiedHostSecondary` constant added next to `DefaultVerifiedHost`.

## Example app

- [Example/Example/Info.plist](Example/Example/Info.plist) — added an `NSPinnedDomains` entry for the secondary host (`api-verified.radar.com`) so ATS pinning doesn't block the failover path in the example app. Uses `NSPinnedCAIdentities` with the SPKI-SHA256 hashes of the 5 Amazon Trust Services root CAs (Amazon Root CA 1–4 and Starfield Services Root CA G2), since the secondary terminates on AWS-issued certs rather than the Cloudflare leaf pinned for `api-verified.radar.io`. Also flattened a pre-existing malformed `<dict>` wrapper under `NSPinnedDomains` that was failing `plutil -lint`.

These are what we will add to https://docs.radar.com/geofencing/fraud#ssl-pinning-2 when we roll out

## Test plan

See https://www.notion.so/radarlabs/SDK-Failover-Testing-Guide-DNS-Hijacking-349e42b18f9680abb34be1b259598553?source=copy_link


Example/Example/Info.plist
```diff
@@ -62,6 +62,34 @@
 				</array>
 			</dict>
 		</dict>
+		<key>api-verified.radar.com</key>
+		<dict>
+			<key>NSIncludesSubdomains</key>
+			<true/>
+			<key>NSPinnedCAIdentities</key>
+			<array>
```
stevepopovich: These were found at https://www.amazontrust.com/repository/

Example/Example/Info.plist
```diff
@@ -62,6 +62,34 @@
...
+					<key>SPKI-SHA256-BASE64</key>
+					<string>9+ze1cZgR9KO1kZrVDxA4HQ6voHRCSVNz4RdTCx4U8U=</string>
+				</dict>
+				<dict>
+					<key>SPKI-SHA256-BASE64</key>
+					<string>KwccWaCgrnaw6tsrrSO61FgLacNgG2MMLq8GE6+oP5I=</string>
+				</dict>
+			</array>
+		</dict>
 	</dict>
```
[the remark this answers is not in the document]
stevepopovich: https://github.com/radarlabs/radar-sdk-ios/pull/576/commits/9aa9d40d597ceeaca2a0d0074c29bcb4dba71e3a

RadarSDK/RadarFailoverAPICoordinator.swift
```diff
@@ -0,0 +1,129 @@
...
+            completionHandler: completionHandler
+        )
+    }
+
+    private func perform(
+        index: Int,
+        path: String,
+        allowFailover: Bool,
+        performRequest: @escaping (String, @escaping (RadarStatus, [AnyHashable: Any]?) -> Void) -> Void,
+        completionHandler: @escaping (RadarStatus, [AnyHashable: Any]?) -> Void
```
ShiCheng-Lu, line 76: maybe use async for this?  callback to async bridge is:

```
Task {
   await perform(...)
   completionHandler()
}
```
you can use for `request`

this will need to be iOS 13.0. We should be able to safely increase our target to iOS 15.0

https://developer.apple.com/news/upcoming-requirements/

If we trust this dude on reddit: https://www.reddit.com/r/iOSProgramming/comments/1qqidyg/enforcing_the_use_of_xcode_26_starting_april/
Image

stevepopovich: Im ok with that, but I would rather do that in a separate PR or followup, wdyt?

RadarSDKTests/RadarAPIClientWiringTests.swift
```diff
@@ -0,0 +1,83 @@
...
+            // stay on the primary verified host. If the coordinator is in the path,
+            // the URL would flip to the secondary.
+            setFailoverFlag(false)
+            apiHelperMock.mockStatus = .errorServer
+
+            let lastUrl = await capturedLastUrl(verified: true)
+
+            #expect(lastUrl?.contains("api-verified.radar.io") == true,
+                    "expected URL to stay on the primary verified host, got: \(lastUrl ?? "nil")")
+            #expect(lastUrl?.contains("api-verified-cf.use1.radar-staging.com") == false,
```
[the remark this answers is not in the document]
stevepopovich: this is fine to leave

PR #581, opened 2026-04-27
Add SwiftLint and swift-format with baseline-grandfathered lint

stevepopovich:
## Motivation

I had some time so I thought it could be nice to improve our tooling in this repo. I like strong lint and formatting, what do you all think? I think it will help as we scale the team and possibly if open source contributions happen. What do you think? Open to not using this at all too.

## Summary

Closes the Swift tooling gap. Before this PR, ObjC had clang-format (formatting) but Swift had nothing. This adds both a linter (SwiftLint) and a formatter (swift-format), wired into \`make lint-swift\`, \`make format\`, and \`make format-check\`, enforced in both CI systems.

| | Format | Lint |
|---|---|---|
| Objective-C | clang-format ✓ (existing) | clang-tidy ✗ (gap) |
| Swift | **swift-format ✓ (this PR)** | **SwiftLint ✓ (this PR)** |

## SwiftLint — baseline mode

Running default rules across the existing Swift files produced ~1,300 violations. Cleaning those up would either bloat this PR or force disabling so many rules the lint would be useless. **Baseline mode** grandfathers existing violations in \`.swiftlint-baseline.json\`; new code is held to the full ruleset.

Config notes (`.swiftlint.yml`):
- \`line_length\` loosened to 200 (matches clang-format + buffer)
- \`identifier_name\` allows \`r/g/b/a\`, \`x/y/z\`, \`id\` (color/coordinate code)
- \`todo\` disabled — \`// TODO:\` comments should not fail CI
- Submodules, Carthage, Pods excluded

Baseline: **166 grandfathered violations** (trailing whitespace and todo entries cleaned up in this PR rather than grandfathered; remaining are \`identifier_name\`, \`trailing_comma\`, \`control_statement\`, etc.).

## swift-format — applied to all existing files

All 28 existing Swift files are formatted so \`make format\` is a no-op from day one. Config notes (`.swift-format`):
- 4-space indent (matches existing files)
- 200-col line length (matches SwiftLint and clang-format)
- \`OrderedImports\` enabled
- \`AlwaysUseLowerCamelCase\` and \`FileScopedDeclarationPrivacy\` disabled — codebase has SCREAMING_SNAKE constants

## CI enforcement

Both checks run in GitHub Actions and CircleCI before the build:

- **\`make lint-swift\`** — SwiftLint with baseline; fails on new violations only
- **\`make format-check\`** — runs swift-format on Swift files changed in the PR (three-dot diff vs \`origin/master\`); fails if any would be reformatted

## Commands

```
make format         # clang_format.sh (ObjC) + swift-format (Swift) — auto-fixes
make lint-swift     # SwiftLint with baseline — reports new violations
make format-check   # swift-format dry-run on changed files — CI gate
```

## Out of scope

- ObjC lint (clang-tidy) — still a gap; separate PR


Makefile
```diff
@@ -31,10 +31,35 @@ lint:
...
+	done
+
+FORMAT_BASE ?= origin/master
+
+lint-swift:
+	@if ! command -v swiftlint >/dev/null; then \
+		echo "swiftlint not installed; run 'make bootstrap' or 'brew install swiftlint'"; \
+		exit 1; \
+	fi
+	swiftlint lint --strict --baseline .swiftlint-baseline.json
```
ShiCheng-Lu: put here because the baseline.json file is huge:

The TODO rule should be ignored if possible, or only warn instead of error. (if that's the current behaviour, that's good, but we should remove those from baseline)

trailing white space I think we can just remove them in the PR.

ShiCheng-Lu: maybe the formatter takes care of the trailing white spaces?

stevepopovich: >The TODO rule should be ignored if possible, or only warn instead of error. (if that's the current behaviour, that's good, but we should remove those from baseline)

Good call.

> trailing white space I think we can just remove them in the PR.

Doing trailing whitespace in this PR was a lot of changes I think, but lemme check the diff

stevepopovich: Cool, disabled TODO and added the white space changes, wdyt? It runs swift format on all those files so the diff is pretty large...

stevepopovich: So yeah, if we fix white space in a file, we have to `swift-format` the whole file, so this diff gets quite large

bootstrap.sh
```diff
@@ -42,4 +42,18 @@ else
   echo "clang-format already installed"
 fi
 
+if ! command -v swiftlint >/dev/null; then
+  echo "installing swiftlint..."
+  brew install swiftlint
```
nickdonnelly: On one hand, this is a convenient way to do things. On the other hand, relying on `brew` in setup scripts is extremely brittle.

At the very least we should lock these to a specific version so everyone is running the same instance of the linter and formatter.

My recommendation would be to have this download the binary directly from [Github releases](https://github.com/realm/SwiftLint/releases/tag/0.63.2) and place it in a gitignored or excluded directory.

stevepopovich: oooo not a bad idea! Yeah changes in linters/formatters can be super annoying. I would be open to this, wonder what others think

alanjcharles14: yeah- i think I agree with Nick on this one. I'd rather not rely on brew unless we absolutely have to and we should be locked on a specific version across the team

stevepopovich: I can adjust 👍🏼

stevepopovich: @alanjcharles14 @nickdonnelly wdyt now? Binary for `SwiftLint` and explained `swift-format` dependency

nickdonnelly (dismissed):
LGTM

ShiCheng-Lu (dismissed):

nickdonnelly (dismissed):

RadarSDK/RadarUtils.swift
```diff
@@ -48,7 +46,7 @@ class RadarUtils: NSObject {
 
     static let country = Locale.current.regionCode
     static let timeZoneOffset = NSNumber(value: TimeZone.current.secondsFromGMT())
-    static let sdkVersion = SDK_VERSION
+    static let sdkVersion = "3.32.0"
```
ShiCheng-Lu, line 49: is this lint error , this was a global scoped variable so the ./set-version.sh can look for a line that looks like "let SDK_VERSION = ...", we'll want to update the script

RadarSDK/RadarSettings.swift
```diff
@@ -316,14 +306,34 @@ internal class RadarSettings: NSObject {
...
+extension RadarSettings {
+
+    public static func updateLastTrackedTime() {
+        RadarUserDefaults.set(Date(), forKey: .LastTrackedTime)
+    }
+
+    public static func updateLastAppOpenTime() {
+        RadarUserDefaults.set(Date(), forKey: .LastAppOpenTime)
+    }
+
```
ShiCheng-Lu, line 326: lint forces you to use a RadarSettings extension?

ShiCheng-Lu (changes requested):
lets update the ./set-version.sh function to reflect the SDK_VERSION change

RadarSDK/RadarSettings.swift
stevepopovich: I think this was leftover from a thing I was trying to do previously, removed!

ShiCheng-Lu (dismissed):
more conflicts lol.

ShiCheng-Lu (approved):

PR #582, opened 2026-04-28
FENCE-2756: add trackVerifiedAutoFailover init option

stevepopovich:
## Summary

Third approach to verified-host failover (supersedes [#576](https://github.com/radarlabs/radar-sdk-ios/pull/576) and [#578](https://github.com/radarlabs/radar-sdk-ios/pull/578)). Local opt-in via a new `RadarInitializeOption`, scoped strictly to the `trackVerified()` codepath, no timers / no retained state / no health probes.

When `trackVerifiedAutoFailover = true`:
1. `trackVerified()` calls `getConfig` against the primary verified host (`api-verified.radar.io`).
2. If the response is non-Radar (transport error, or parsed body missing a top-level `meta` key — i.e. a Cloudflare-edge HTML page from a DNS hijack), it retries `getConfig` once against the secondary (`api-verified.radar.com`).
3. Whichever host returned a Radar response is reused for the subsequent `track` call within the same invocation.
4. The next `trackVerified()` invocation starts fresh at primary — no state carried across calls.

When the flag is off (default), behavior is unchanged.

## Implementation

- **`RadarInitializeOptions`** ([Include/RadarInitializeOptions.h](RadarSDK/Include/RadarInitializeOptions.h), [RadarInitializeOptions.m](RadarSDK/RadarInitializeOptions.m)) — new `trackVerifiedAutoFailover` BOOL, defaults `NO`, round-trips through the existing `dictionaryValue` / `initWithDict:` persistence path.
- **`RadarSettings`** ([RadarSettings.swift](RadarSDK/RadarSettings.swift), [RadarSettings.h](RadarSDK/RadarSettings.h)) — `DefaultVerifiedHostSecondary = "https://api-verified.radar.com"` constant exposed to ObjC.
- **`RadarAPIClient`** ([RadarAPIClient.h](RadarSDK/RadarAPIClient.h), [RadarAPIClient.m](RadarSDK/RadarAPIClient.m)) — new `verifiedHostOverride:` parameter added as overloads on `getConfigForUsage:` and `trackWithLocation:`. Existing 3-arg / 12-arg signatures are preserved by delegation, so all existing callers (`Radar.m`, `RadarLocationManager.m`, `RadarSdkConfiguration.m`, etc.) are untouched. Override is threaded through `makeTrackRequestWithParams:` to the URL construction site. Override is ignored when `verified == NO`.
- **`RadarVerificationManager`** ([RadarVerificationManager.m](RadarSDK/RadarVerificationManager.m)) — `trackVerifiedWithBeacons:...` reads the flag once, extracts the post-getConfig flow into a `continueWithConfig` block parameterized by `chosenVerifiedHost`, and dispatches to either the legacy single-call path (flag off) or the primary-then-secondary path (flag on). The non-Radar discriminator is `status != Success || !config || !config.meta`.
- **Example app** ([Example/Example/Info.plist](Example/Example/Info.plist)) — `NSPinnedDomains` entry for `api-verified.radar.com` with the 5 Amazon Trust Services SPKI hashes (the secondary terminates on AWS-issued certs rather than the Cloudflare leaf pinned for the primary). Validated with `plutil -lint`.

Explicitly **not** added:,no `useVerifiedHostFailover` SDK dconfig flag, no health probes. 

## Test plan

https://www.notion.so/radarlabs/Manual-test-guide-verified-host-auto-failover-FENCE-2756-349e42b18f9680abb34be1b259598553?source=copy_link


RadarSDK/RadarVerificationManager.m
```diff
@@ -169,12 +170,11 @@ - (void)trackVerifiedWithBeacons:(BOOL)beacons desiredAccuracy:(RadarTrackingOpt
                  indoorScan:nil
                  verified:YES
                  fraudPayload:fraudPayload
-                 // -- payload encryption --
```
nickdonnelly, line 172: Why was this here, and why was it axed?

stevepopovich: I am not sure...but a comment without explanation is worth being removed IMO 🙃

nickpatrick: @cameron-radar can share context (let's discuss offline since this is a public repo)

stevepopovich: Confirmed offline this is safe to remove 👍🏼

RadarSDK/RadarVerificationManager.m
```diff
@@ -244,7 +244,41 @@ - (void)trackVerifiedWithBeacons:(BOOL)beacons desiredAccuracy:(RadarTrackingOpt
...
+            }];
+        }];
+    } else {
+        [[RadarAPIClient sharedInstance]
+         getConfigForUsage:@"verify"
+         verified:YES
+         completionHandler:^(RadarStatus status, RadarConfig * _Nullable config) {
+            continueWithConfig(status, config, nil);
+        }];
+    }
```
cameron-radar: is it necessary to duplicate the initial getConfig call code here? why not make the call and then if it fails conditionally perform the second. it looks like the first getconfig call is specifying the default nil value of the optional verifiedHostOverride param

RadarSDK/RadarAPIClient.h
```diff
@@ -75,6 +75,11 @@ typedef void (^_Nonnull RadarSyncLogsAPICompletionHandler)(RadarStatus status);
 
 - (void)getConfigForUsage:(NSString *_Nullable)usage verified:(BOOL)verified completionHandler:(RadarConfigAPICompletionHandler _Nonnull)completionHandler;
 
+- (void)getConfigForUsage:(NSString *_Nullable)usage
+                 verified:(BOOL)verified
+     verifiedHostOverride:(NSString *_Nullable)verifiedHostOverride
```
cameron-radar: why are passing in the host instead of just a bool indicating to use the new DefaultVerifiedHostSecondary in settings

cameron-radar: this applies to track as well

cameron-radar: for context, RadarApiClient already reads the default host and verified host from config

RadarSDK/RadarSettings.swift
```diff
@@ -12,6 +12,9 @@ internal class RadarSettings: NSObject {
     
     static let DefaultHost = "https://api.radar.io"
     static let DefaultVerifiedHost = "https://api-verified.radar.io"
+    static let DefaultVerifiedHostSecondary = "https://api-verified.radar.com"
+
+    @objc public static var defaultVerifiedHostSecondary: String { DefaultVerifiedHostSecondary }
```
cameron-radar: does this need to be marked @objc? can't we copy the pattern for host and verifiedHost?

cameron-radar (commented):
a few code quality nitpicks but it looks like the logic is what was being agreed to on the friday meeting (before I had to drop for another meeting) 👍

nickdonnelly (dismissed):
Tested all test cases from notion on an iOS simulator successfully.

stevepopovich:
@cameron-radar great review, my obj-c is...not great haha cleaned those up

RadarSDK/RadarSettings.swift
stevepopovich: 6522fcc1b2d308e6b2e4aaeee39c4d23cbaabe21

RadarSDK/RadarAPIClient.h
stevepopovich: 154d1f05b422e99aa636ebce2da13894c55c3305

RadarSDK/RadarVerificationManager.m
stevepopovich: b55a41b9e7fdab3afa033f4108f2df772bf6b8ec

alanjcharles14:
test passed for me as well- although i started testing before your latest commits. I don't see anything in there that should change the outcome though

nickdonnelly (dismissed):

tjulien (dismissed):

RadarSDK/RadarInitializeOptions.m
```diff
@@ -18,6 +18,7 @@ - (instancetype)init {
         _autoLogNotificationConversions = NO;
         _autoHandleNotificationDeepLinks = NO;
         _silentPush = NO;
+        _trackVerifiedAutoFailover = NO;
```
lmeier, line 21: I didn't follow the conversation in real time but I imagine we went this direction to avoid having a successful `/config` in the codepath of having this on or off.

You could imagine still wanting to have a killswitch for this behavior, which could come from `/config`. Idk if necessary / worth the complexity.

RadarSDK/RadarVerificationManager.m
```diff
@@ -169,12 +170,11 @@ - (void)trackVerifiedWithBeacons:(BOOL)beacons desiredAccuracy:(RadarTrackingOpt
                  indoorScan:nil
                  verified:YES
                  fraudPayload:fraudPayload
-                 // -- payload encryption --
-                 // fraudKeyVersion:fraudKeyVersion
                  expectedCountryCode:self.expectedCountryCode
                  expectedStateCode:self.expectedStateCode
                  reason:reason
                  transactionId:transactionId
+              useSecondaryVerifiedHost:useSecondaryVerifiedHost
```
lmeier: nit: indentation is off here

RadarSDK/RadarVerificationManager.m
```diff
@@ -244,6 +244,29 @@ - (void)trackVerifiedWithBeacons:(BOOL)beacons desiredAccuracy:(RadarTrackingOpt
                 }
             }];
         }];
+    };
+
```
lmeier, line 248: Hmmm, there's a `searchBeaconsNear` call that we might also want to route through secondary at line 204

lmeier: In both cases we continue through regardless of result so fine, I think.

RadarSDK/RadarInitializeOptions.m
stevepopovich: I the idea is that we would leave that up to the integrators. This is opt-in now so folks can turn on with their own systems, should they want to

RadarSDK/RadarVerificationManager.m
stevepopovich: https://github.com/radarlabs/radar-sdk-ios/pull/582/commits/7557dc7dcbc469c27026d6219b86925588cb2b60

lmeier (approved):

stevepopovich:
Went through the steps one more time with a physical device and Charles, everything worked as expected, merging!

PR #585, opened 2026-05-04
FENCE-2795: add ability to configure default network timeout

nickdonnelly:
- Adds a new initialization option, `networkTimeoutInterval`, that allows consumers to override the default 10 second timeout.
- Sets the extended timeout to 2.5x the network timeout instead of hardcoding 25 seconds.


alanjcharles14 (dismissed):

stevepopovich (dismissed):
I'd be interested to document the 2.5x multiplier on the extended timeout, cause it seems so random (magic number), but I also think: I would have picked something similar lol. I understand it is pre-existing though

Looks good, thanks!

alanjcharles14 (approved):

PR #586, opened 2026-05-04
FENCE-2797: Move synchronize() to coalesced background queue

alanjcharles14:
## Summary

Moves `userDefaults.synchronize()` to a coalesced background queue. Per [Apple's docs](https://developer.apple.com/documentation/foundation/userdefaults/synchronize()) the call is "unnecessary and shouldn't be used"  but to be conservative I kept it alive on a background queue rather than dropping it entirely — preserving cross-process visibility timing for app-group consumers without paying for it on the calling thread.

## New SDK Components

**`flushQueue`** — A labeled serial `DispatchQueue` (`io.radar.userdefaults.flush`, QoS `.utility`) dedicated to background `synchronize()` calls. Labeled rather than using the global utility queue so flushes show up clearly in stack traces and Instruments.

**`flushLock` / `pendingFlushTargets`** — A lock-protected `[ObjectIdentifier: UserDefaults]` dictionary used to coalesce a burst of writes against the same target into a single background flush. Keying by `ObjectIdentifier(target)` ensures writes against different `UserDefaults` instances each get their own flush, while writes against the same instance dedup correctly.

**`scheduleFlush(for:)`** — Internal helper called from `set(_:forKey:)` with the snapshotted target instance. Performs the locked dedup check on `pendingFlushTargets[id]`, enqueues a flush if needed, and the closure looks the target back up from the dictionary at flush time so the same `UserDefaults` instance that received the write is the one that gets `synchronize()` called on it.

## How It Works

1. `RadarUserDefaults.set(_:forKey:)` snapshots the current `userDefaults` once into a local `target`, writes the value to `target` synchronously (in-memory, immediate), then calls `scheduleFlush(for: target)`. Capturing the target at write time means a swap of `RadarUserDefaults.userDefaults` (e.g. via `RadarSettings.setAppGroup`) between the write and the eventual flush does not redirect the synchronize to the wrong instance.
2. `scheduleFlush(for:)` derives `id = ObjectIdentifier(target)`, takes `flushLock`, checks `pendingFlushTargets[id] != nil`, sets the entry, and releases. If a flush against this target was already scheduled, the function returns and the in-flight flush will pick up the new write.
3. Otherwise it dispatches one block onto `flushQueue`. Inside, the block re-acquires `flushLock`, calls `removeValue(forKey: id)` to clear the entry *before* synchronizing (so any write that races with the flush schedules a follow-up flush against the same target), releases the lock, and calls `synchronize()` on the captured `UserDefaults`.
4. The synchronize runs on the background queue — never on the calling thread.
5. `clone(from:to:)` is intentionally left writing directly to the target `UserDefaults` rather than going through `RadarUserDefaults.set`, so an `setAppGroup` swap doesn't enqueue redundant flushes inside `Key.allCases`. The post-clone `set(nil, forKey: .AppGroup)` and `set(appGroup, forKey: .AppGroup)` writes (one against the previous defaults, one against the new) each correctly schedule their own per-target flush.


stevepopovich:
Is it possible to write some tests for this? These changes are kinda scary IMO. Could be weird to test though...

alanjcharles14:
@stevepopovich yea if you take a look at the last few commits i did have a regression test in there that seemed to work pretty well locally but not at all on the CI (which is slower) and it kind of got to the point that getting it to work with the CI wasn't really testing anything worthwhile... But I agree, it's a scary fix just hard to test reliably.

stevepopovich (approved):
I am not super well versed on locks, but seems reasonable. The fact that it is "unnecessary" gives me the mental OK to ship without tests or some sort of observability that it is working.

Are you able to manually test that things are working? Not sure what success really looks like here

lmeier (approved):

PR #594, opened 2026-05-12
FENCE-2791: Add guard for NSException in dictionaryToJson

alanjcharles14:
- Adds a guard for `NSJSONSerialization's` `NSException` path (NaN, non-string key, unsupported types), which Swift's do/catch doesn't catch.
- Resolves crash in react native as it still uses objective-c under the hood.


RadarSDK/RadarUtils.swift
```diff
@@ -190,6 +190,13 @@ class RadarUtils: NSObject {
     static func dictionaryToJson(_ dict: [String: Any]?) -> String {
         guard let dict = dict else { return "{}" }
 
+        // Defends against NSJSONSerialization's NSException path (NaN, non-string keys,
+        // unsupported types), which Swift's do/catch below doesn't catch.
+        guard JSONSerialization.isValidJSONObject(dict) else {
+            RadarLogger.shared.warning("RadarUtils.dictionaryToJson: input is not a valid JSON object, returning empty")
```
stevepopovich, line 196: Missing some context, should we log as error? Not sure when we would expect an integrator to run into this.

stevepopovich (commented):
So in what cases was this reproducible?  Just curious about what was going on here

RadarSDK/RadarUtils.swift
alanjcharles14: yeah it will only present itself in our cross-platform SDKs afaik. I ran into it working on a deep-linking issue. Something the customer is sending isn't valid json and because our native modules in react native are still objective-c an NSException is thrown. 

There will likely be some follow-up work to sanitize the `logConversionWithNotification` function to handle whatever weird json they are sending so they don't lose the metadata when logging the conversion but this will unblock the crash for now.

stevepopovich (dismissed):
Sounds good!

lmeier (dismissed):

lmeier (approved):

PR #595, opened 2026-05-13
FENCE-2804: Add SDK config flag for Swift location manager rollout and simplify tests, make CI more robust (1/N)

stevepopovich:
Similar to https://github.com/radarlabs/radar-sdk-ios/pull/589 just opening a new one for better stacking

First PR in a multi-step migration of RadarLocationManager from Objective-C to Swift

This PR is plumbing only — adds a server-driven feature flag in RadarSdkConfiguration that future PRs will use to route between the existing ObjC manager and a new Swift implementation. No behavior change.

**CI reliability update**

While investigating a GitHub Actions failure where XCTest reported `157 tests, 0 failures` but `xcodebuild` still exited with `TEST FAILED`, we found that the workflow was running raw `xcodebuild` commands that had drifted from the Makefile. The visible XCTest summary did not include Swift Testing output, and the `xcpretty` step made the failure mode harder to diagnose.

This PR now routes the CI build/analyze and test steps through dedicated Makefile targets:

- `make ci-build-analyze`
- `make ci-test-pretty`
- `make ci-test-swift`

Benefits:

- Keeps local reproduction and GitHub Actions using the same commands.
- Preserves the existing CI build graph by continuing to use the Example workspace and `RadarSDK` scheme.
- Splits the XCTest-heavy `xcpretty` phase from Swift Testing suites so Swift Testing output stays visible.
- Tees raw `xcodebuild` output to `/tmp` logs before `xcpretty`, so CI failures have searchable diagnostics even when pretty output hides details.
- Aligns the GitHub Actions skip/only-testing filters with the Makefile so newly added Swift Testing suites do not accidentally run in the wrong phase.

**What comes next**

[This is the AI-driven, code focused implementation plan](https://www.notion.so/radarlabs/Multi-PR-Breakdown-for-RadarLocationManager-Migration-359e42b18f9680ccb0a5c425e4d210bb?source=copy_link)


nickdonnelly (dismissed):
🚀

Makefile
```diff
@@ -21,9 +27,6 @@ build:
 test:
 	xcodebuild $(XC_TEST_ARGS) test
 
-test-swift:
```
stevepopovich, line 24: This was already defined below with a more correct definition

nickdonnelly (approved):

PR #596, opened 2026-05-13
FENCE-2834: Update release actions for Node 24

stevepopovich:
## Summary
- update GitHub Actions workflow pins from Node 20-era majors to Node 24-compatible versions
- replace deprecated set-output usage in the iOS release workflow
- keep the change scoped to workflow maintenance for FENCE-2834

### DIFF
--- a/.github/workflows/build-and-test.yml
+++ b/.github/workflows/build-and-test.yml
@@ -13,7 +13,7 @@ jobs:
 
     steps:
       - name: Checkout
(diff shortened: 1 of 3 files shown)


ShiCheng-Lu:
https://github.com/radarlabs/radar-sdk-ios/pull/504

I had these for set-output, but no one wanted to review them.

but I'll review yours

ShiCheng-Lu (approved):

PR #598, opened 2026-05-13
FENCE-2804: Swift cutover scaffold for RadarLocationManager (2/N)

stevepopovich:
Built on top of https://github.com/radarlabs/radar-sdk-ios/pull/595

## Summary

Introduces a minimal seam for migrating `RadarLocationManager` from Objective-C to Swift one method at a time, gated on the existing `useSwiftLocationManager` SDK config flag (default off, no behavior change for existing users).

- New `RadarLocationManagerSwift` Swift class holds methods that have been ported to Swift.
- Each migrated method in `RadarLocationManager.m` gets a single guard at the top that forwards to its Swift twin when the flag is on, and otherwise runs the existing ObjC body unchanged.
- `restartPreviousTrackingOptions` is migrated as the first method, both to validate the seam and to give the test suite something real to exercise.

This is probably the most important PR in this migration, since it makes the decision of how we will migrate the rest of these functions.

## Approach and rationale

**Per-method dispatch via a flag guard, not a class-level facade.** The flag is checked at the top of each method that has been ported; everything else in `RadarLocationManager.m` is untouched. This avoids splitting the class into a facade + implementation pair (which would force every method — migrated or not — through an extra indirection) and lets us migrate methods independently without churning the rest of the class.

**A standalone `@objc` Swift class plus a thin ObjC header, rather than a Swift extension on `RadarLocationManager`.** `RadarLocationManager.h` is a project-visibility header and is not in the framework's auto-synthesized Swift module, so Swift inside this target cannot extend the class directly. The codebase already handles this case the same way for other Swift-implemented types (`RadarLogger.h` + `RadarLogger.swift`, `RadarSettings.h` + `RadarSettings.swift`): keep a small ObjC `.h` as the compile-time handle, put the implementation in Swift. We follow that pattern. For methods that need `self`-state from the manager, we'll pass it through a small `@objc` protocol — added when the first such method is migrated, not preemptively.

**Swift twins exist only for migrated methods.** We don't pre-create stubs for the whole class surface. If a method hasn't been ported, the flag-on path just runs the ObjC body — same as flag-off. This keeps the diff for each migration small and self-contained.

## Migration workflow per method

1. Port the body to `RadarLocationManager+Swift.swift` as `@objc static func <name>(...)`. If state from `self` is needed, accept a parameter typed as a small `@objc` protocol that exposes the @objc-visible properties on `RadarLocationManager`.
2. Add the method declaration to `RadarLocationManagerSwift.h`.
3. Prepend the 5-line guard to the matching method in `RadarLocationManager.m`.
4. Add a behavior test in `RadarLocationManagerSwiftSeamTests` that exercises the Swift twin directly and the public method with the flag on.
5. Once the Swift port has soaked in production behind the flag, delete the ObjC body and the dispatch guard, leaving only the Swift implementation.

When every method has been migrated and the dispatch guards are gone, `RadarLocationManager.m` becomes empty enough to delete, and `RadarLocationManagerSwift` either takes over the `RadarLocationManager` name (via `@objc(RadarLocationManager)`) or absorbs the public ObjC surface.

## Test plan

Build and run `Example/Example.xcodeproj` on a simulator with location permission granted.

**1 — Flag off (default, ObjC path):**
Start a trip via `Radar.startTrip(tripOptions:trackingOptions:.presetContinuous)`, then call `Radar.completeTrip` (or `cancelTrip`). Confirm tracking resumes with the pre-trip options (or stops if there were none). No crash.

**2 — Flag on (Swift path):**
`Radar.initialize` fetches SDK config from the server and overwrites `UserDefaults`, so a pre-launch inject is clobbered before any trip runs. To force the flag on locally, temporarily change the default in `RadarSdkConfiguration.swift`:

```swift
// Before (line 59):
useSwiftLocationManager = dict?["useSwiftLocationManager"] as? Bool ?? false
// Change to:
useSwiftLocationManager = dict?["useSwiftLocationManager"] as? Bool ?? true
```

Revert this line before merging. Repeat the same trip cycle and confirm:
- The debug log line `"🦅 Restarting previous tracking options"` appears.
- Tracking behavior is identical to step 1.
- No crash.


RadarSDK/RadarLocationManager+Swift.swift
```diff
@@ -0,0 +1,36 @@
...
+// auto-synthesized Swift module, so we cannot extend `RadarLocationManager` from Swift.
+// Static methods on this class are called from `RadarLocationManager.m` via
+// `RadarSDK-Swift.h`. When a method needs state from the manager instance, pass it in
+// via a small @objc protocol.
+@objc(RadarLocationManagerSwift)
+final class RadarLocationManagerSwift: NSObject {
+
+    @objc static func restartPreviousTrackingOptions() {
+        let previousTrackingOptions = RadarSettings.previousTrackingOptions
+        RadarLogger.shared.debug("🦅 Restarting previous tracking options")
```
ShiCheng-Lu, line 26: new icon unlocked?

ShiCheng-Lu (approved):
LGTM,

I would like this to be parallel testable though (ok if you don't want to do that), to do that, you'd want to make `RadarLocationManagerSwift` an instanced class, with a `shared` instance, and calls will call on the shared instance.

Then each test can instantiate `RadarLocationManagerSwift` without worrying about test canoodling.

Actually might just do that with a different PR that make RadarSettings mockable.

RadarSDK/RadarLocationManager+Swift.swift
stevepopovich: Built in MacOS emoji lmao

nickdonnelly (approved):
_Seams_ like a straightforward approach to me... 🥁

stevepopovich:
@ShiCheng-Lu I am not sure what you mean...we could pair on this if you feel strongly but not sure I understand the motivation of what you mean

alanjcharles14 (approved):
this project overall would be a good thing to test with the new example app

ShiCheng-Lu:
> @ShiCheng-Lu I am not sure what you mean...we could pair on this if you feel strongly but not sure I understand the motivation of what you mean

Not currently, but when we add more functions to the RadarLocationManager, we might want to add tests for it in swift that are all capable of running in parallel without stepping on each other's toes with global variables / settings / states.

So we'll want to be able to create multiple RadarLocationManager each with their own RadarSettings/RadarState environment, there'll need to be RadarState and RadarSettings changes as well to support this.

nickdonnelly:
> @ShiCheng-Lu I am not sure what you mean...we could pair on this if you feel strongly but not sure I understand the motivation of what you mean

Basically a singleton, the swift pattern is typically:
```swift
class MyClass {
    static let shared = MyClass()
    private init() {}

    func method() {} // <--- this is NOT static
}

// Elsewhere, access using
MyClass.shared.method()
```

True it does helps with testability.

stevepopovich:
Ok I thought about this, I think it is probably not worth it to complicate this cutover. I think it is more important to keep this cutover simple rather than maximize testing in parallel IMO. Tests should be able to override the cutover setting so we don't have global test state issues

PR #601, opened 2026-05-17
Add didChangeIP() delegate and setUserLanguage()

nickpatrick:
## Summary

- Adds `RadarVerifiedDelegate.didChangeIP()` and `didChangeSharing(_:)`. Both are `@optional`, so existing conformers are unaffected.
- IP monitoring auto-starts when `setVerifiedDelegate` is called with a non-nil delegate, OR while verified tracking is active. Clearing the delegate auto-stops it (unless verified tracking is still active). Verified tracking continues to trigger `trackVerified()` on IP change as before.
- `didChangeSharing(_:)` fires whenever `Radar.isSharing()` transitions. The fraud module bridge exposes `setSharingDidChangeHandler:` (see `RadarSDKFraudProtocol.h`); the main SDK registers a block that forwards to the verified delegate via `RadarDelegateHolder`. Dedup is enforced inside the fraud module so the delegate only sees real transitions.
- Adds `Radar.setUserLanguage(_:)` / `getUserLanguage()` backed by `RadarSettings`. When set, the value is sent as `lang` on `POST /track`. Pass-through, so any BCP-47 tag works (e.g. `"en"`, `"es-PR"`).
- Bumps `RadarSDKFraud` submodule to `sharing-calls` for the sharing-change handler (radarlabs/radar-sdk-ios-fraud#21).

Mirrors the Android changes in https://github.com/radarlabs/radar-sdk-android/pull/520.


RadarSDK/RadarUserDefaults.swift
```diff
@@ -16,6 +16,7 @@ class RadarUserDefaults: NSObject {
         case SessionId = "radar-sessionId"
         case Id = "radar-_id"
         case UserId = "radar-userId"
+        case userLanguage = "radar-userLanguage"
```
nickpatrick, line 19: Lowercase is inconsistent with other values, but needed to pass linting

RadarSDK/Include/Radar.h
```diff
@@ -616,6 +630,22 @@ typedef void (^_Nonnull RadarIndoorsScanCompletionHandler)(NSString *_Nullable r
  */
 + (void)stopTrackingVerified NS_SWIFT_NAME(stopTrackingVerified());
 
+/**
+ Starts monitoring for changes to the device's IP address. When the IP changes, `didChangeIP()` is delivered to the verified delegate. If verified tracking is started, an IP change will also trigger a `trackVerified()` call.
+
+ @note `startTrackingVerified()` automatically starts IP change monitoring, and `stopTrackingVerified()` stops it.
```
cameron-radar: will anyone expect to be able to call startMonitoringIPChanges, then startTrackingVerifiedWithInterval, then stopTrackingVerified, and still expect ip changes to be monitored because they haven't called stopMonitoringIPChanges?

nickpatrick: Not a use case we need to support. Generally folks are going to be using one of (a) `startTrackingVerified()` or (b) `trackVerifed()` + `startMonitoringIPChanges()`, not both

RadarSDK/Radar.m
```diff
@@ -206,6 +206,14 @@ + (NSString *_Nullable)getUserId {
     return [RadarSettings userId];
 }
 
++ (void)setUserLanguage:(NSString *)userLanguage {
+    [RadarSettings setUserLanguage:userLanguage];
+}
```
lmeier, line 216: Imagine that it's totally fine that we don't get this server side until next `/track`, but worth calling out.

lmeier (dismissed):

RadarSDK/Radar.m
nickpatrick: Yeah that's fine

RadarSDK/RadarUtils.swift
```diff
@@ -45,7 +45,7 @@ class RadarUtils: NSObject {
 
     static let country = Locale.current.regionCode
     static let timeZoneOffset = NSNumber(value: TimeZone.current.secondsFromGMT())
-    static let sdkVersion = "3.33.1"
+    static let sdkVersion = "3.34.0-beta.1"
```
alanjcharles14: looks like we missed the version here.

alanjcharles14 (dismissed):

alanjcharles14 (approved):

PR #602, opened 2026-05-18
FENCE-2769: Add offline beacons and dwell events

alanjcharles14:
## Summary

Two-part change:

1. Refactor `RadarSyncManager` to collapse the duplication between the online `checkForGeofence*` private helpers (which both detected and mutated state) and the public `getGeofence*` helpers (which only detected). State detection now lives in pure, location-driven getters; state mutation lives in clearly-named helpers that both the online and offline paths share.
2. Extend `RadarOfflineEventManager.generateEvents` to also emit `user.entered_beacon`, `user.exited_beacon`, and `user.dwelled_in_geofence` events, reusing the new shared `RadarSyncManager` APIs.

## New SDK Components

`RadarSyncManager.getGeofenceEntries/Exits/Dwells(for:against:)`Three pure, side-effect-free getters that, given a location and a baseline of last-known geofence IDs, return the geofences that just entered / exited / crossed their dwell threshold. `Dwells` is new; `Entries`/`Exits` already existed but are now the single source of truth (no more duplicated logic in `checkForGeofenceEntries/Exits`).

`RadarSyncManager.getBeaconEntries/Exits(for:against:)`Proximity-based beacon diff helpers. Mirrors the geofence pattern but uses location distance instead of BLE ranging, since offline event generation can't rely on `RadarBeaconManager`.

`RadarSyncManager.recordGeofenceEntryTimestamps(_:) / clearGeofenceEntryState(_:)`Shared state-mutation primitives that write to `RadarSyncState.geofenceEntryTimestamps` / `dwellEventsFired`. Now used by both `hasGeofenceStateChanged` (online) and `RadarOfflineEventManager.generateEvents` (offline), so the two paths can't drift.

`RadarOfflineEventManager` beacon + dwell event generation `generateEvents` now emits beacon enter/exit and geofence dwell events alongside the existing geofence enter/exit. New `offlineBeaconIds` cache mirrors the existing `offlineGeofenceIds` cache so subsequent offline ticks don't re-fire the same beacon entries.

## Test Plan
- 8 new cases in `RadarOfflineEventManagerTests` covering beacon enter/exit/no-change, dwell threshold reached/not-reached/already-fired, and offline-side geofence timestamp record/clear.


.swiftlint-baseline.json
```diff
@@ -1 +1 @@
...location) { events, user, loc in"},{"violation":{"reason":"Unused parameter in a closure should be replaced with _","ruleIdentifier":"unused_closure_parameter","ruleDescription":"Unused parameter in a closure should be replaced with _","location":{"line":22,"file":"RadarSDKTests\/RadarSettingsTest.swift","character":64},"severity":"warning","ruleName":"Unused Closure Parameter"},"text":"        userDefaults.dictionaryRepresentation().forEach { key, value in"},{"violation":{"reason":"Struct body should span 350 lines or less excluding comments and whitespace: currently spans 670 lines","ruleIdentifier":"type_body_length","ruleDescription":"Type bodies should not span too many lines","location":{"line":16,"file":"RadarSDKTests\/RadarSyncManagerTests.swift","character":5},"severity":"error","ruleName":"Type Body Length"},"text":"    struct RadarSyncManagerTests {"},{"violation":{"reason":"File should contain 400 lines or less: currently contains 874","ruleIdentifier":"file_length","ruleDescription":"Files should not span too many lines.","location":{"line":874,"file":"RadarSDKTests\/RadarSyncManagerTests.swift","character":1},"severity":"warning","ruleName":"File Length"},"text":"}"}]
```
stevepopovich: Damn this diff stinks 😂 I will ponder if we can make this more readable for diffs

RadarSDK/RadarSyncManager.swift
```diff
@@ -374,17 +374,32 @@ public final class RadarSyncManager: NSObject {
...
-        let currentGeofences = getGeofences(for: location)
-        let currentGeofenceIds = Set(currentGeofences.map { $0.id })
+        let lastKnownIds = Set(state.lastSyncedGeofenceIds)
 
-        if checkForGeofenceEntries(currentGeofences: currentGeofences, currentGeofenceIds: currentGeofenceIds, lastKnownGeofenceIds: lastKnownGeofenceIds) {
+        let entries = getGeofenceEntries(for: location, against: lastKnownIds)
+        if !entries.isEmpty {
+            for geofence in entries {
+                RadarLogger.shared.debug("SyncManager: Detected geofence entry: \(geofence.id)")
+            }
```
stevepopovich: Random optimization thing: Should we iterate over entries twice just to log? Can we combine the log loop and the map loop? Same thing below

stevepopovich (dismissed):
Seems great to me!

RadarSDK/RadarOfflineEventManager.swift
```diff
@@ -201,3 +277,46 @@ class RadarOfflineEventManager: NSObject {
...
+            "type": "ibeacon",
+        ]
+        if let desc = beacon.description { dict["description"] = desc }
+        if let tag = beacon.tag { dict["tag"] = tag }
+        if let externalId = beacon.externalId { dict["externalId"] = externalId }
+        if let geometry = beacon.geometry {
+            dict["geometry"] = ["coordinates": [geometry.longitude, geometry.latitude]]
+        }
+        return dict
+    }
```
ShiCheng-Lu, line 306: are we intentionally skipping certain fields on geofences/beacons?

maybe should use Encoder to encode to json string then parse to dict

RadarSDK/RadarSyncManager.swift
alanjcharles14: mm good call!

RadarSDK/RadarOfflineEventManager.swift
alanjcharles14: yah i have no clue why i did that. Good call

ShiCheng-Lu (dismissed):

.swiftlint-baseline.json
stevepopovich, line 1: Im so sorry but I tried to make the swift baseline more readable in [this PR](https://github.com/radarlabs/radar-sdk-ios/pull/614).  You just gotta run `make bootstrap` and regenerate the baseline and the diff should be more reasonable. Sorry about that 🙇🏼

alanjcharles14: ohh i don't think I have that on my branch, i'll update my branch and try again. No worries!

alanjcharles14: did I do that correctly?

stevepopovich: yes! Now we can actually see the diff 🙃 thank you!

stevepopovich (approved):
Seems great to me!

PR #604, opened 2026-05-21
consolidate geofence swift

ShiCheng-Lu:
we had RadarGeofenceSwift for event sync and RadarGeofence_Swift for notifications, RadarGeofenceSwift was missing metadata, added that so it can be consolidated.

see https://github.com/radarlabs/radar-sdk-ios/pull/569
The other PR got a little messy with merge conflicts, so I've re-done the consolidation on main.

https://linear.app/radarlabs/issue/FENCE-2853/swift-geofence-clean-up

### DIFF
--- a/RadarSDK/RadarCoordinate.swift
+++ b/RadarSDK/RadarCoordinate.swift
@@ -9,7 +9,7 @@
 import CoreLocation
 import Foundation
 
-struct RadarCoordinateSwift: Codable, Sendable {
+struct RadarCoordinateSwift: Codable, Sendable, Equatable {
     let latitude: Double
     let longitude: Double
 
--- a/RadarSDK/RadarGeofence.swift
+++ b/RadarSDK/RadarGeofence.swift
@@ -149,42 +149,6 @@ private struct GeoJSONPolygon: Codable, Sendable {
     let coordinates: [[[Double]]]
 }
 
-// MARK: - ObjC Bridge Types
-
-struct RadarCoordinateCodable: Codable, Sendable, Equatable {
-    static func == (lhs: RadarCoordinateCodable, rhs: RadarCoordinateCodable) -> Bool {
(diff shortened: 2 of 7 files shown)


RadarSDK/RadarCoordinate.swift
```diff
@@ -9,7 +9,7 @@
 import CoreLocation
 import Foundation
 
-struct RadarCoordinateSwift: Codable, Sendable {
```
stevepopovich, line 12: So this just implements `==` implicitly right?

RadarSDK/RadarGeofence.swift
```diff
@@ -27,16 +27,33 @@ enum RadarGeofenceGeometrySwift: Codable, Sendable {
...
+        }
+        
+        // maybe this should check for coordinates match as well, but for notification purposes, center + radius is enough
+        if case .polygon(_, let lhsCenter, let lhsRadius) = lhs,
+           case .polygon(_, let rhsCenter, let rhsRadius) = rhs {
+            return lhsCenter == rhsCenter && lhsRadius == rhsRadius
+        }
+        
+        // types don't match, not both circular or both polygonal
+        return false
```
stevepopovich: why not use `switch` so it is exhaustive by default?

stevepopovich (dismissed):
Is there an attached ticket? Looks legit though

alanjcharles14:
@ShiCheng-Lu happy to re-approve once the linter is happy. Would do it now but it'll just get dismissed I think

RadarSDK/RadarCoordinate.swift
ShiCheng-Lu: yeah, adding Equatable automatically implements == since everyfield of this class is equatable.

RadarSDK/RadarGeofence.swift
ShiCheng-Lu: I forgot that's possible

alanjcharles14 (approved):

PR #605, opened 2026-05-28
FENCE-2804: Port region/beacon sync and lifecycle methods to Swift (3/N)

stevepopovich:
## Summary

- Ports 10 small `RadarLocationManager` methods to Swift twins on `RadarLocationManagerSwift`, dispatched from `RadarLocationManager.m` when `useSwiftLocationManager` is enabled. ObjC bodies are kept in place; both implementations coexist until the Swift port is trusted.
- Methods: `matchBeaconIds:syncedBeacons:`, `replaceBubbleGeofence:radius:`, `removeBubbleGeofence`, `removeSyncedGeofences`, `replaceSyncedBeacons:`, `replaceSyncedBeaconUUIDs:`, `removeSyncedBeacons`, `removeAllRegions`, `requestLocation`, `shutDown`.
- Swift twins take the relevant `CLLocationManager` (and `lowPowerLocationManager` for `shutDown`) directly. A small `@objc` host protocol was prototyped but couldn't be made visible to Swift inside the framework target without exposing it in the umbrella header — passing managers directly keeps the seam minimal until more shared instance state (timer, completion handlers) makes a protocol worthwhile.
- Mirrors the identifier-prefix constants (`radar_`, `radar_bubble_`, etc.) as `private static let` on `RadarLocationManagerSwift` so the two implementations stay in sync until full cutover.

### Manual smoke in the Example app

Run with `useSwiftLocationManager` first **off** (default) to confirm no regression, then **on** to exercise the new Swift twins. Flip the flag by setting the SDK config in the example app's setup, e.g.:

- [ ] **Region sync (geofences):** Start tracking with `presetResponsive`. While near a synced geofence in the Radar dashboard, inspect `RadarLocationManager.sharedInstance().locationManager.monitoredRegions` in the debugger — should contain `radar_geofence_*` identifiers. Stop tracking, then re-inspect — they should be removed. Repeat with flag both off and on; behavior should be identical.
- [ ] **Bubble geofence:** With `presetResponsive` (which sets up a bubble geofence on stops), wait for stopped state and verify a `radar_bubble_*` region appears in `monitoredRegions`. Move and watch it get replaced.
- [ ] **Beacons:** If you have a synced beacon on your account with `options.beacons = YES`, verify a `radar_beacon_*` region appears in `monitoredRegions` after a track call.
- [ ] **`requestLocation`:** Call any code path that triggers `getLocationWithCompletionHandler:` (e.g. `Radar.trackOnce`) and confirm a `didUpdateLocations` callback fires via `RadarLogger` output.
- [ ] **`shutDown`:** Stop tracking and wait ~10s; logs should show \"Shutting down\" and standard/low-power `CLLocationManager` updates should stop.
- [ ] **`removeAllRegions`:** Call via any path that clears all radar-prefixed regions (e.g. SDK reinit); confirm `monitoredRegions` no longer contains anything with the `radar_` prefix.

Switch the flag back **off** between each step to confirm the ObjC path still works.

### DIFF
--- a/RadarSDK.xcodeproj/project.pbxproj
+++ b/RadarSDK.xcodeproj/project.pbxproj
@@ -158,6 +158,10 @@
 		BA3540222F90020F00E553A1 /* RadarOfflineEventManager.h in Headers */ = {isa = PBXBuildFile; fileRef = BA3540212F90020900E553A1 /* RadarOfflineEventManager.h */; };
 		BA3540242F9033A900E553A1 /* RadarSerializedTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = BA3540232F90339F00E553A1 /* RadarSerializedTests.swift */; };
 		B0A0F0032FBC000100111111 /* RadarLocationManagerSwiftSeamTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = B0A0F0072FBC000100111111 /* RadarLocationManagerSwiftSeamTests.swift */; };
+		B0A0F00C2FBC000100111111 /* TrackingCLLocationManager.swift in Sources */ = {isa = PBXBuildFile; fileRef = B0A0F00D2FBC000100111111 /* TrackingCLLocationManager.swift */; };
+		B0A0F00E2FBC000100111111 /* RadarLocationManagerSwiftRegionTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = B0A0F00F2FBC000100111111 /* RadarLocationManagerSwiftRegionTests.swift */; };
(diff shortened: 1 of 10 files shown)


stevepopovich:
Closing in favor of two smaller PRs:

- PR A: Strengthen tests for the existing Swift implementation (`restartPreviousTrackingOptions`) and introduce reusable test infrastructure.
- PR B: Port only the four beacon-related methods (`matchBeaconIds`, `replaceSyncedBeacons`, `replaceSyncedBeaconUUIDs`, `removeSyncedBeacons`) with full coverage.

This PR's geofence / region-removal / lifecycle ports will come in a later batch.

PR #607, opened 2026-05-29
FENCE-2804: Port beacon sync methods to Swift

stevepopovich:
> **Depends on #606.** This PR targets that branch so the diff stays focused on the beacon port. Once #606 lands, this PR will be rebased onto `master`.

## Summary

Ports the four beacon-related methods from [RadarSDK/RadarLocationManager.m](RadarSDK/RadarLocationManager.m) to Swift twins on `RadarLocationManagerSwift`, dispatched when `useSwiftLocationManager` is enabled. ObjC bodies are kept in place; both implementations coexist until the Swift port is trusted.

### Methods ported
- `matchBeaconIds:syncedBeacons:` — pure function, no instance state.
- `replaceSyncedBeacons:` — needs the `CLLocationManager` for `startMonitoring` / `requestState`.
- `replaceSyncedBeaconUUIDs:` — same.
- `removeSyncedBeacons` — needs the `CLLocationManager` for `stopMonitoring`.

Swift twins that need a `CLLocationManager` take it as an explicit argument; the identifier-prefix constants (`radar_beacon_`, `radar_uuid_`) are mirrored as `private static let` on `RadarLocationManagerSwift` until the ObjC bodies are deleted.

### Part A — Radar dashboard setup (one time)
1. **Enable beacons for the project.** Settings → "Beacons" section: toggle beacons **on**.
   Add your transmitter's UUID to the **iBeacon UUIDs** comma-separated list (this drives the
   `radar_uuid_` / `replaceSyncedBeaconUUIDs` path). 
2. **Create a beacon entity.** **Geofencing → Beacons** (`/geofencing/beacons/new`):
   - Type: **iBeacon**
   - UUID / Major / Minor: **match your transmitter exactly**
   - Geometry: coordinates **within ~1 km of where the device will physically be** (the
     `searchBeaconsNear` radius). This drives the `radar_beacon_` / `replaceSyncedBeacons` path.
   - Description: anything; Save. 
   - [Follow this guide should you have issues adding your beacon](https://app.notion.com/p/radarlabs/Adding-Kontakt-Beacons-to-Radar-373e42b18f9680faaab2d565f34bb523?source=copy_link)

### Part B — Example app setup

1. **Force the Swift path on** Hardcode
   [RadarSDK/RadarSdkConfiguration.swift:60](RadarSDK/RadarSdkConfiguration.swift) so server config can't flip it mid-test:
   ```swift
   useSwiftLocationManager = true   // was: dict?["useSwiftLocationManager"] as? Bool ?? true
   ```
2. Build & run on the device. Grant **"Always"** location. In the app's **Logs** tab, ensure
   debug logs are visible (`Radar.setLogLevel(.debug)` if needed).

### Part C — Verify the Swift path (flag ON)
With the device near the dashboard beacon and the transmitter broadcasting:
1. Tap **startTracking** (the beacons-enabled button), then tap **trackOnce** to force a sync.
2. In the **Logs** tab, confirm **emoji-prefixed** lines appear — this proves the Swift twin ran:
   - `🦅 Synced beacon | identifier = radar_beacon_<id>; uuid = …` (from `replaceSyncedBeacons`), and/or
   - `🦅 Synced UUID | identifier = radar_uuid_<uuid>; uuid = …` (from `replaceSyncedBeaconUUIDs`).
   - The **absence** of the emoji on these lines would mean the ObjC body ran instead → fail.
3. **Range / enter-exit:** walk into the transmitter's range. Confirm:
   - `🦅 Beacon ID matching | synced=…, ranged=…, matchedIds=[…]` appears (Swift `matchBeaconIds`,
     emoji-prefixed), and
   - a `user.entered_beacon` event shows in the **Events** tab and in the dashboard.
   Walk out of range → expect `user.exited_beacon`.
4. **Stop tracking** → re-inspect `monitoredRegions`: the `radar_beacon_*` / `radar_uuid_*` regions
   should be gone (`removeSyncedBeacons`).

### Part D — Regression check (flag OFF, ObjC path)
1. Flip [RadarSdkConfiguration.swift:60](RadarSDK/RadarSdkConfiguration.swift) to
   `useSwiftLocationManager = false`, rebuild, and repeat Part C.
2. Behavior should be **identical**, except the same log lines now appear **without** the `🦅`
   prefix (`Synced beacon | …`, `Beacon ID matching | …`) — confirming the ObjC fallback is
   doing the same work.

### Cleanup
Revert the three temporary edits (AppDelegate key, RadarSdkConfiguration hardcode) before committing.


alanjcharles14:
@stevepopovich love the emoji choice lol but probably something we'll want to cleanup before going to prod? We need to clean up logging generally across the SDK so maybe your swift conversion work is a good place to agree on a convention? Or do you want to keep logging unique until you're done with the conversion so it's easier to verify?

stevepopovich:
Ah thats a good point. Ideally, I would like to keep this around so we have very clear signal as to which implementation is being used. We could make it a clean up item at this end of this project to remove it?

Right now, I would like to keep the functionality (including logging behavior) the same as it is currently. Is that fair?

alanjcharles14:
hmm, my only thing is I don't think we want these logs showing up with emojis in customer dashboards. For testing/validating i think it's completely fine. So, depending on how you want to handle the project overall, I think we can leave it for now but do need to clean them up before we start turning this on for customers. When/how that cleanup happens is entirely your call. Thoughts?

stevepopovich:
Ill leave them in for now, [made a follow up issue!](https://linear.app/radarlabs/issue/FENCE-2873/rollout-swift-radarlocationmanager)

alanjcharles14 (approved):

ShiCheng-Lu (approved):

PR #608, opened 2026-06-01
FENCE-2857: Add NSError parameter to ipGeocode callback

stevepopovich:
Ports [radarlabs/radar-sdk-android#525](https://github.com/radarlabs/radar-sdk-android/pull/525) to iOS.

## Summary

- Adds a new four-arg `+ipGeocodeWithErrorCompletionHandler:` (typedef `RadarIPGeocodeWithErrorCompletionHandler`) that exposes the underlying `NSError` when the failure originated from a caught network/parse/exception error, so integrators can forward it to error collectors (Sentry, Crashlytics, Datadog).
- Threads the caught `NSError` through `RadarAPIHelper`'s three error paths (`NSURLSession` error, `NSJSONSerialization` error, `@catch NSException`) and through the internal `RadarIPGeocodeAPICompletionHandler`. The parse and exception paths also gain a `RadarLogger` error log that they were previously missing.
- Swift consumers see only the new four-arg block under the existing `Radar.ipGeocode { ... }` call site (via `NS_SWIFT_NAME`).

## Backwards compatibility

- The legacy three-arg `+ipGeocodeWithCompletionHandler:` is preserved, marked `__attribute__((deprecated))`, and routes through the new method (ignoring the error). Existing ObjC integrators keep compiling and keep receiving callbacks.
- The internal generic `RadarAPICompletionHandler` block typedef gains an `NSError *` (third arg). Every callsite in `RadarAPIClient.m` was updated to take and pass-through the new parameter even though only `ipGeocode` consumes it — mirrors the Android PR and primes future extension to other callbacks.
- A `test_Radar_ipGeocode_legacy_three_arg_callback_still_called` test locks the BC behavior in.

Scope is intentionally limited to `ipGeocode` as the pilot. Version bump / MIGRATION will be handled in a separate PR.

## Test plan

- [ ] Manually: airplane-mode the Example app, tap `ipGeocode`, confirm the console prints `error = The Internet connection appears to be offline.`

### DIFF
--- a/RadarSDK/RadarLogger.swift
+++ b/RadarSDK/RadarLogger.swift
@@ -26,6 +26,14 @@ final class RadarLogger: NSObject, @unchecked Sendable {
         logLevelOverride ?? RadarSettings.logLevel
     }
 
+    // Test-only bookkeeping: each `log(...)` call spawns a detached Task that delivers the
+    // message to the delegate asynchronously. Tests need to deterministically await that
+    // delivery instead of racing a wall-clock timeout (which is flaky under CI load). When
+    // `logLevelOverride` is set (test mode) we retain the in-flight tasks so `awaitPendingLogs()`
+    // can await them. In production `logLevelOverride` is nil, so this stays a no-op.
+    private let pendingLogTasksLock = NSLock()
+    private var pendingLogTasks = [Task<Void, Never>]()
+
     @MainActor
     let device = {
         UIDevice.current.isBatteryMonitoringEnabled = true
@@ -65,7 +73,7 @@ final class RadarLogger: NSObject, @unchecked Sendable {
             return
         }
 
-        Task {
+        let task = Task {
             let log = RadarLog(level: level, message: message, type: type, createdAt: Date(), includeDate: includeDate, battery: includeBattery ? await self.device.batteryLevel : nil)
 
             await RadarLogBuffer.shared.log(log)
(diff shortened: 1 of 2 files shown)


Example/Example/TestsView.swift
```diff
@@ -259,9 +259,9 @@ struct TestsView: View {
             }
 
             StyledButton("ipGeocode") {
-                Radar.ipGeocode { (status, address, proxy) in
+                Radar.ipGeocode { (status, address, proxy, error) in
                     print(
-                        "IP geocode: status = \(Radar.stringForStatus(status)); country = \(String(describing: address?.countryCode)); city = \(String(describing: address?.city)); proxy = \(proxy); full address: \(String(describing: address?.dictionaryValue()))"
+                        "IP geocode: status = \(Radar.stringForStatus(status)); country = \(String(describing: address?.countryCode)); city = \(String(describing: address?.city)); proxy = \(proxy); full address: \(String(describing: address?.dictionaryValue())); error = \(String(describing: error?.localizedDescription))"
```
nickdonnelly: is `error?.localizedDescription` not already a a string? why do we need to wrap in `String(describing: ...)`?

nickdonnelly (commented):
Looks like CI is unhappy.

Example/Example/TestsView.swift
stevepopovich: mm I was just copying, lemme see about that

ShiCheng-Lu: optional string needs to be wrapped in describing to produce "nil" for nil, otherwise compiler will warn.

ShiCheng-Lu: error?.locationzedDescription ?? "nil" is equiv.

ShiCheng-Lu (dismissed):

Example/Example/TestsView.swift
nickdonnelly: ~@ShiCheng-Lu you shouldn't need `describing` for plain old interpolation on a `nil`-able.~
Monday brain, ignore

stevepopovich: yeah that was cleaner, good call (need to step up my iOS game!)

stevepopovich:
Superseded by #611. This PR was auto-closed as "merged" due to a branch mix-up that briefly pointed its head and base at the same commit — no code actually merged into `master`. The ipgeocode work continues in #611 (stacked on #610, the flaky-logger-test fix).

PR #615, opened 2026-06-04
FENCE-2804: Port bubble-geofence + region-removal methods to Swift

stevepopovich:
## Summary

Continues the `RadarLocationManager` ObjC→Swift migration by porting the **bubble-geofence + region-removal** cluster, following the established beacon-sync seam pattern.

Ported as `@objc static` twins on `RadarLocationManagerSwift`:
- `replaceBubbleGeofence(locationManager:location:radius:)`
- `removeBubbleGeofence(locationManager:)`
- `removeAllRegions(locationManager:)`

Each ObjC method dispatches to its twin when `useSwiftLocationManager` is enabled and `return`s; both implementations coexist until the Swift port is trusted. Instance state (`CLLocationManager`) is passed in explicitly.

## Changes
- `RadarLocationManager+Swift.swift` — three Swift twins + mirrored `radar_` / `radar_bubble_` prefix constants.
- `RadarLocationManagerSwift.h` — ObjC-visible declarations.
- `RadarLocationManager.m` — `useSwiftLocationManager` dispatch shims (ObjC bodies kept for coexistence).
- `project.pbxproj` — registers the new test file.
- `RadarLocationManagerSwiftRegionTests.swift` — behavior tests: prefix-scoped removal, tracking-gated add, and the new region's center/radius.
- `AGENTS.md` — note that `make format` reformats the entire repo (separate commit).

### Part A — Example app setup
1. **Force the Swift path on.** Hardcode [RadarSdkConfiguration.swift:60](RadarSDK/RadarSdkConfiguration.swift) so server config can't flip it mid-test:
   ```swift
   useSwiftLocationManager = true   // was: dict?["useSwiftLocationManager"] as? Bool ?? false
   ```
2. Build & run on a **real device** (bubble geofences need real Core Location). Grant **"Always"** location. In the app's **Logs** tab, ensure debug logs are visible (`Radar.setLogLevel(.debug)` if needed).

### Part B — Verify the Swift path (flag ON)
1. Tap **startTracking** with the **Responsive** preset (it enables both the stopped and moving bubble geofences at a 100 m radius).
2. On the first location update, confirm the **emoji-prefixed** line appears — this proves the Swift twin ran:
   - `🦅 Successfully added bubble geofence | latitude = …; longitude = …; radius = 100; identifier = radar_bubble_<uuid>` (from `replaceBubbleGeofence`).
   - The **absence** of the `🦅` on this line would mean the ObjC body ran instead → fail.
3. **Move / re-evaluate:** walk ≳150 m (beyond the 100 m radius) so the SDK syncs and replaces the bubble. Confirm:
   - `🦅 Removed bubble geofences` (from `removeBubbleGeofence`), immediately followed by a new
   - `🦅 Successfully added bubble geofence | …` at the updated location.

### Part C — Regression check (flag OFF, ObjC path)
1. Flip [RadarSdkConfiguration.swift:60](RadarSDK/RadarSdkConfiguration.swift) to `useSwiftLocationManager = false`, rebuild, and repeat Part B.
2. Behavior should be **identical**, except the same log lines now appear **without** the `🦅` prefix (`Successfully added bubble geofence | …`, `Removed bubble geofences`) — confirming the ObjC fallback does the same work.

### Cleanup
Revert the `RadarSdkConfiguration.swift` hardcode before committing.


RadarSDK/RadarLocationManager+Swift.swift
```diff
@@ -25,6 +25,8 @@ final class RadarLocationManagerSwift: NSObject {
 
     // Mirror of the identifier prefix constants in RadarLocationManager.m. Kept in sync by
     // hand until that file is fully ported.
+    private static let identifierPrefix = "radar_"
```
nickdonnelly, line 28: Maybe we make these `@objc static let`s and ref them from the objective-c codebase so we don't have to keep them in sync?

nickdonnelly: or reference them in swift via a shared header that exposes them

RadarSDK/RadarLocationManager+Swift.swift
nickdonnelly: What are the implications of this defaulting to zero if the int cast fails?

nickdonnelly, line 126: might be worth adding a debug log here so it's obvious when this gets skipped, same for the guard below

nickdonnelly, line 141: I'd prefer filtering before the loop instead of using `continue` to skip iterations.

```swift
let validUUIDs = uuids.prefix(numUUIDs).compactMap { uuid -> (String, UUID)? in
    guard let proximityUUID = UUID(uuidString: uuid) else {
	   // log here
	   return nil
   }
   (uuid, proximityUUID)
}

// Then this, or use a `.forEach` since you know you won't need to bail on any of the iterations
for (uuid, proximityUUID) in validUUIDs {
	// work
}
```

RadarSDK/RadarLocationManager+Swift.swift
stevepopovich: I think this is ok. Sharing them makes it a bit more work to clean it up and I would like to keep the implementations as separate as possible. Usually I would agree with you but in this transitional case, I think it is ok to have some duplication so this file is fully testable without using obj c code

RadarSDK/RadarLocationManager+Swift.swift
stevepopovich: Tomato, tomato 🙃 

I feel like having having 2 loops is marginally slower and unneeded

nickdonnelly: I'd have more sympathy for the performance argument if this wasn't bounded to a 9 elements 😅 
But you're right, it's purely a personal preference nit - functional composition over collections reads easier to me at a glance.

stevepopovich: > I'd have more sympathy for the performance argument if this wasn't bounded to a 9 elements

Good point 😆

stevepopovich: thats good call, we would use a beacon with 0,0, so guarded against that in https://github.com/radarlabs/radar-sdk-ios/pull/616

stevepopovich: 👍🏼 Added a bit more logging in https://github.com/radarlabs/radar-sdk-ios/pull/616

nickdonnelly (dismissed):

ShiCheng-Lu (dismissed):
LGTM.
we should look into the new geofencing APIs after this migration, should be easier to refactor in swift, (just easier to read everything + async stuff).

stevepopovich:
That is the main point of doing this refactor first 🙃

nickdonnelly (approved):
```
 __     ___  ____  _  _ 
(  )   / __)(_  _)( \/ )
/ (_/\( (_ \  )(  / \/ \
\____/ \___/ (__) \_)(_/
```

ShiCheng-Lu (approved):

PR #616, opened 2026-06-04
Harden Swift beacon syncing: guard major/minor + log skipped syncs

stevepopovich:
## Summary

Follow-up to [#607](https://github.com/radarlabs/radar-sdk-ios/pull/607) addressing review feedback from @nickdonnelly on [#615](https://github.com/radarlabs/radar-sdk-ios/pull/615). Two small hardening fixes to the ported Swift beacon-sync methods in `RadarLocationManager+Swift.swift`.

## Changes

1. **Guard against non-numeric beacon `major`/`minor`** ([#615 comment](https://github.com/radarlabs/radar-sdk-ios/pull/615#discussion_r3358227506))
   - `RadarBeacon.major`/`.minor` are `String`s. The previous `Int(beacon.major) ?? 0` silently fell back to a `0/0` `CLBeaconRegion` when a value wasn't a parseable integer — monitoring the wrong region. Now a malformed major/minor skips the beacon and logs an error, mirroring the existing `proximityUUID` guard.

2. **Log skipped syncs** ([#615 comment](https://github.com/radarlabs/radar-sdk-ios/pull/615#discussion_r3358241374))
   - The three `useRadarModifiedBeacon` early returns and the tracking/options guard in `replaceSyncedBeaconUUIDs` returned silently. Added debug logs so it's clear when/why beacon syncing is skipped, matching the existing log in `replaceSyncedBeacons`.

## Notes
- Behavior change is limited to the malformed-input path (previously synced a bogus `0/0` region; now skips it). Valid input is unchanged.

### DIFF
--- a/RadarSDK/RadarLocationManager+Swift.swift
+++ b/RadarSDK/RadarLocationManager+Swift.swift
@@ -91,8 +91,17 @@ final class RadarLocationManagerSwift: NSObject {
                 continue
             }
 
-            let major = CLBeaconMajorValue(truncatingIfNeeded: Int(beacon.major) ?? 0)
-            let minor = CLBeaconMinorValue(truncatingIfNeeded: Int(beacon.minor) ?? 0)
+            // A non-numeric major/minor would otherwise silently monitor a 0/0 region, which
+            // collides with other beacons. Skip the beacon instead of syncing bad values.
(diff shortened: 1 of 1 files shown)


nickdonnelly (approved):
💯

ShiCheng-Lu (approved):

PR #617, opened 2026-06-08
FENCE-2856: Deregister CSGNs based on operating hours

alanjcharles14:
## Summary

Adds client-side support for restricting geofence notifications (CSGNs) to a business's operating hours. The SDK now caches each geofence's `operatingHours`, suppresses registering a CSGN while the geofence is closed, and on a new `radar:refreshNotifications` silent push it re-evaluates its cached geofences against the current time to prune notifications that have since closed (and re-register ones that have reopened). 

## New SDK Components

**`RadarOperatingHoursEvaluator`** : A port of the server's `DateUtils.isWithinOperatingHours`: given a geofence's `operatingHours`, the current time, and a timezone, returns whether it's currently open. Matches the server's day-of-week indexing, strict open/close comparison, `"24:00"` handling, and "no hours ⇒ always open" semantics. Evaluates in the device's local timezone (the user is physically near the geofence for a CSGN).

**`RadarGeofenceSwift.operatingHours`**: New optional field on the cached geofence model, decoded from the `/track` response and included in the `Codable` round-trip so it survives on-device caching.

**Operating-hours gate in `toNotificationRequest(now:)`**:  When a geofence's metadata carries `radar:restrictToOperatingHours == true` and the evaluator reports closed at `now`, registration is suppressed (returns `nil`). Reads an optional `radar:operatingHoursCloseBufferMinutes` (defaults to `0`), so it's safe before the server sends it.

**Durable geofence cache (`RadarFileStorageObject<[RadarGeofenceSwift]>`)**: On every registration, the full synced geofence set (incl. `metadata` + `operatingHours`) is persisted to `radar_notification_geofences.json`, mirroring the `RadarSyncManager` sync-region pattern. This lets the refresh push re-evaluate without a track, and lets a reopened geofence re-register.

**`refreshGeofenceNotifications()` + `radar:refreshNotifications` handler**: A new silent-push type dispatched from `Radar.didReceivePushNotificationPayload` to `RadarNotificationHelper`. It reloads the cached geofences and re-runs registration against the current time (replace-all-then-add), pruning now-closed CSGNs and re-registering still-open/reopened ones — with no location request or `/track`.

 ### QA details
Doing QA with a physical device and produced the following
- `trackOnce` to pull down `CSGN` but it's < 30 mins until it closes -> `CSGN` doesn't sync down
- `trackOnce` to pull down `CSGN` before buffer window, see it come in, `trackOnce` after `closeBuffer` window and see it be removed
- removed by server on `/track` response 
- silent push with type `radar:refreshNotifications` while `CSGN` is still within business hours does not remove `CSGN`
silent push with type `radar:refreshNotifications` while `CSGN` is outside business hours removes `CSGN`


RadarSDK/Radar.m
```diff
@@ -1761,6 +1761,10 @@ + (void)didReceivePushNotificationPayload:(NSDictionary *)payload completionHand
         [Radar trackOnceWithCompletionHandler:^(RadarStatus status, CLLocation * _Nullable location, NSArray<RadarEvent *> * _Nullable events, RadarUser * _Nullable user) {
             completionHandler();
         }];
+    } else if ([payload[@"type"] isEqual:@"radar:refreshNotifications"]) {
+        [[RadarNotificationHelper_Swift shared] refreshGeofenceNotificationsWithCompletionHandler:^() {
```
stevepopovich, line 1765: Hmm what am I missing...where is this function defined?

stevepopovich (dismissed):
Looks great to me!

RadarSDK/Radar.m
alanjcharles14: [RadarNotificationHelper.h](https://github.com/radarlabs/radar-sdk-ios/pull/617/changes#diff-5b902fbe93cc51848ae80aecd82977e80a8dbdbe14c5f243edd50f2bb77b05a7) right here!

stevepopovich (dismissed):
I don't fully understand why we need a durable cache for the geofences, but I am probably just a dumb guy, looks good

alanjcharles14:
@stevepopovich we need to store the full geofence for the CSGN pruning to work correctly. `registeredNotifications` lacks operatingHours/geometry/metadata, so you can't re-run `RadarOperatingHoursEvaluator.isOpen` against it on a refresh push. If we tried to use only pending notifications we'd also lose the geofence and couldn't re-register it when the business re-opens at a later point in time.

RadarSDK/RadarNotification.swift
```diff
@@ -114,6 +118,26 @@ extension RadarGeofenceSwift {
 
         return UNNotificationRequest(identifier: identifier, content: content, trigger: trigger)
     }
+
+    private func isWithinSchedulingWindow(now: Date) -> Bool {
```
stevepopovich, line 122: nice!

stevepopovich (approved):

PR #618, opened 2026-06-09
FENCE-2370: Add campaign scheduling window support to CSGN

stevepopovich:
## Summary

- Parses `radar:startsAt` and `radar:endsAt` from geofence metadata (ISO 8601 naive local datetimes, timezone suffix stripped for wall-clock comparison)
- **Swift v2 path** (`RadarNotification.swift` → `toNotificationRequest(now:)`): returns `nil` if `now` is outside the window, skipping notification registration
- **Legacy ObjC path** (`RadarLocationManager.m` → `replaceSyncedGeofences:`): `continue`s past out-of-window geofences before calling `extractContentFromMetadata`
- Both paths share the same formatter format (`yyyy-MM-dd'T'HH:mm:ss.SSS`, device local timezone) and window semantics: `[startsAt, endsAt)` — start inclusive, end exclusive

Companion PRs: radarlabs/dashboard#725, radarlabs/server#7902

### Setup

1. In the dashboard, create a geofence campaign with a notification message and attach it to a test geofence.
2. Use the server's campaign scheduling API (or dashboard) to set `startsAt` and `endsAt` on the campaign.

I ended up running the server PR locally and using ngrok to send requests from physical device to my local machine

### Before window (notification should NOT fire)

- [ ] Set `endsAt` to a time in the past (e.g. yesterday). Trigger a `/v1/track` call that enters the geofence. Confirm no local notification is scheduled on the device.

### After window (notification should NOT fire)

- [ ] Set `startsAt` to a time in the future (e.g. tomorrow). Trigger a `/v1/track` call that enters the geofence. Confirm no local notification is scheduled on the device.

### Within window (notification SHOULD fire)

- [ ] Set `startsAt` to a time in the past and `endsAt` to a time in the future. Trigger a `/v1/track` call that enters the geofence. Confirm the local notification fires as expected.

### No scheduling window (notification SHOULD fire)

- [ ] Create a campaign with no `startsAt` or `endsAt`. Trigger a `/v1/track` call that enters the geofence. Confirm existing notification behavior is unchanged.

### Boundary (end exclusive)

- [ ] Set `endsAt` to the current time (as close as possible). Trigger a `/v1/track` call immediately. Confirm no notification fires (end of window is exclusive).


RadarSDKTests/RadarSyncManagerTests.swift
```diff
@@ -22,6 +22,8 @@ extension RadarSerializedTests {
 
         init() {
             Radar.initialize(publishableKey: "prj_test_pk_0000000000000000")
+            RadarSyncManager.stop()
+            RadarUserDefaults.set(nil, forKey: .LastLocation)
```
stevepopovich, line 26: Unrelated to the scheduling window feature. It fixes a pre-existing race condition where `Radar.initialize` triggers an async `fetchSyncRegion` task that could overwrite test state and cause `getGeofences_insideCircle` to flake in CI. Bundled here since it was blocking the PR from going green.

RadarSDK/RadarLocationManager.m
```diff
@@ -619,6 +619,32 @@ - (void)replaceSyncedGeofences:(NSArray<RadarGeofence *> *)geofences {
...
+                static NSDateFormatter *windowFormatter;
+                static dispatch_once_t windowFormatterOnce;
+                dispatch_once(&windowFormatterOnce, ^{
+                    windowFormatter = [[NSDateFormatter alloc] init];
+                    windowFormatter.locale = [NSLocale localeWithLocaleIdentifier:@"en_US_POSIX"];
+                    windowFormatter.dateFormat = @"yyyy-MM-dd'T'HH:mm:ss.SSS";
+                });
+                NSDate *now = [NSDate date];
+                NSString *startsAtString = [metadata objectForKey:@"radar:startsAt"];
+                if ([startsAtString isKindOfClass:[NSString class]] && startsAtString.length >= 23) {
```
nickdonnelly: this `.length >= 23` check warrants an explanation

RadarSDK/RadarLocationManager.m
```diff
@@ -619,6 +619,32 @@ - (void)replaceSyncedGeofences:(NSArray<RadarGeofence *> *)geofences {
...
+                    NSDate *startsAt = [windowFormatter dateFromString:[startsAtString substringToIndex:23]];
+                    if (startsAt && [now compare:startsAt] == NSOrderedAscending) {
+                        [[RadarLogger sharedInstance] logWithLevel:RadarLogLevelDebug
+                            message:[NSString stringWithFormat:@"Skipping notification: before window | geofenceId = %@", geofenceId]];
+                        continue;
+                    }
+                }
+                NSString *endsAtString = [metadata objectForKey:@"radar:endsAt"];
+                if ([endsAtString isKindOfClass:[NSString class]] && endsAtString.length >= 23) {
```
nickdonnelly: same here, and in the swift code

RadarSDK/RadarNotification.swift
```diff
@@ -58,7 +58,28 @@ struct RadarNotificationContent: Sendable, Hashable {
...
+        let formatter = RadarGeofenceSwift.schedulingWindowFormatter
+        if let startsAtStr = metadata?["radar:startsAt"]?.string() {
+            let datePart = String(startsAtStr.prefix(23))
+            if let startsAt = formatter.date(from: datePart), now < startsAt {
+                return nil
+            }
+        }
+        if let endsAtStr = metadata?["radar:endsAt"]?.string() {
+            let datePart = String(endsAtStr.prefix(23))
+            if let endsAt = formatter.date(from: datePart), now >= endsAt {
```
nickdonnelly: Why is the `endsAt` comparison inclusive and the `startsAt` comparison exclusive?

stevepopovich: Good catch! This drifted from what is on the server, we match now!

RadarSDK/RadarLocationManager.m
stevepopovich: Made it a constant and added an explanation, good call 👍🏼

nickdonnelly (approved):

PR #619, opened 2026-06-09
FRA-2021: Escape non-ascii characters

alanjcharles14:
Escapes any non-ASCII character in the header value to a JSON `\uXXXX` sequence so the value is always ASCII-safe, while remaining valid JSON that the backend decodes identically.

### DIFF
--- a/RadarSDKTests/RadarUtilsTests.swift
+++ b/RadarSDKTests/RadarUtilsTests.swift
@@ -50,4 +50,31 @@ struct RadarUtilsTests {
         let dict: [String: Any] = ["nested": inner]
         #expect(RadarUtils.dictionaryToJson(dict) == "{}")
     }
+    
+    @Test func escapeNonAsciiLeavesAsciiUntouched() {
+        #expect(RadarUtils.escapeNonAsciiCharacters("{\"name\":\"Cafe\"}") == "{\"name\":\"Cafe\"}")
+    }
+    
+    @Test func escapeNonAsciiEscapesAccentedCharacters() {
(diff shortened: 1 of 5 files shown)


RadarSDKTests/RadarUtilsTests.swift
```diff
@@ -50,4 +50,29 @@ struct RadarUtilsTests {
...
+
+    @Test func escapeNonAsciiResultIsValidJson() throws {
+        let json = RadarUtils.escapeNonAsciiCharacters(RadarUtils.dictionaryToJson(["name": "Café"]))
+        // Every character is now ASCII.
+        #expect(json.allSatisfy { $0.isASCII })
+        let decoded = try JSONSerialization.jsonObject(with: Data(json.utf8)) as? [String: String]
+        #expect(decoded?["name"] == "Café")
+    }
+
+    @Test func escapeNonAsciiHandlesEmojiSurrogatePairs() throws {
```
alanjcharles14, line 71: probably not strictly necessary, but fun

lmeier:
@alanjcharles14 did we test against the server with an escaped name? Just to confirm e2e behavior? Otherwise lgtm

alanjcharles14:
@lmeier seems unnecessary since the server is just parsing the json which is functionally equivalent to what the tests here are doing. 

```
    let applicationInfo;
    try {
      applicationInfo = headers['x-radar-app-info'] ? JSON.parse(headers['x-radar-app-info']) : undefined;
    } catch {
      applicationInfo = undefined;
    }
```

running this in a terminal shows that it was handled correctly on the server already and handles the new fix though: 
image

lmeier (approved):

PR #621, opened 2026-06-16
FENCE-2804: Port `removeSyncedGeofences` to Swift

stevepopovich:
## Summary

Continues the `RadarLocationManager` ObjC→Swift migration, following the seam pattern from #615. Ports the clean-leaf **`removeSyncedGeofences`** as an `@objc static` twin on `RadarLocationManagerSwift`. The ObjC method dispatches to the twin when `useSwiftLocationManager` is enabled and `return`s; both implementations coexist until the Swift port is trusted. Instance state (`CLLocationManager`) is passed in explicitly.

> Stacked on top of #615 (`stevepopovich/port-bubble-geofence-methods`) — this PR is based against that branch, so the diff shows only this change.

### Why only `removeSyncedGeofences` (and not `replaceSyncedGeofences`)?
`replaceSyncedGeofences` is **not** a clean leaf — it calls two *internal* ObjC `RadarNotificationHelper` methods (`extractContentFromMetadata:identifier:`, `updateClientSideCampaignsWithPrefix:notificationRequests:`) that the framework's own Swift cannot reach: `RadarNotificationHelper.h` is not in the public umbrella (`RadarSDK/RadarSDK.h`) and has no Swift twin. (Corroborated by `RadarSettings.swift`'s `// TODO: call swift RadarReplayBuffer when implemented` — the migration deliberately avoids calling internal ObjC from Swift.) This is almost certainly why synced-geofences was left behind while the no-notification bubble/beacon clusters were ported.

### Part A — Example app setup
1. **Force the Swift path on** so server config can't flip it mid-test. Hardcode `useSwiftLocationManager` in `RadarSDK/RadarSdkConfiguration.swift`:
   ```swift
   useSwiftLocationManager = true   // was: dict?["useSwiftLocationManager"] as? Bool ?? false
   ```
2. Build & run the Example app on a **real device** (region monitoring needs real Core Location). Grant **"Always"** location.
3. In the app's **Logs** tab, ensure debug logs are visible (`Radar.setLogLevel(.debug)` if needed).

### Part B — Verify the Swift path (flag ON)
`removeSyncedGeofences` runs whenever synced geofences are cleared. The still-ObjC `replaceSyncedGeofences` calls `[self removeSyncedGeofences]` internally on every geofence sync, so the Swift twin is exercised through the normal tracking flow:

1. Tap **startTracking** with the **Responsive** preset and let the SDK sync geofences from the server (you should see `Synced geofence | …` lines as `radar_geofence_*` regions are registered).
2. Trigger a re-sync (move location / re-evaluate, or stop+start tracking). On each sync, `replaceSyncedGeofences` first removes the existing synced geofences. Confirm the **emoji-prefixed** line appears — this proves the Swift twin ran:
   - `🦅 Removed synced geofences`
   - The **absence** of the `🦅` on this line (i.e. a plain `Removed synced geofences`) would mean the ObjC body ran instead → fail.
3. **Confirm scoping:** after the removal, only `radar_geofence_*` regions are gone — any active `radar_bubble_*` bubble geofence and `radar_beacon_*` / `radar_uuid_*` beacon regions must remain monitored (the SDK keeps re-adding/keeping them). Geofence enter/exit events should keep working after re-sync, confirming nothing else was torn down.

### Part C — Regression check (flag OFF, ObjC path)
1. Flip `useSwiftLocationManager = false` in `RadarSdkConfiguration.swift`, rebuild, and repeat Part B.
2. Behavior should be **identical**, except the log line now appears **without** the `🦅` prefix (`Removed synced geofences`) — confirming the ObjC fallback does the same work.

### Cleanup
Revert the `RadarSdkConfiguration.swift` hardcode before merging.


RadarSDK/RadarLocationManager+Swift.swift
```diff
@@ -198,6 +199,16 @@ final class RadarLocationManagerSwift: NSObject {
         RadarLogger.shared.debug("🦅 Removed bubble geofences")
     }
 
+    @objc(removeSyncedGeofencesOnLocationManager:)
+    static func removeSyncedGeofences(locationManager: CLLocationManager) {
+        for region in locationManager.monitoredRegions
+        where region.identifier.hasPrefix(syncGeofenceIdentifierPrefix) {
```
ShiCheng-Lu, line 207: that's cool, didn't know this syntax existed!

ShiCheng-Lu (dismissed):

RadarSDK/RadarLocationManager+Swift.swift
stevepopovich: me either, I actually really like it!

RadarSDK/RadarLocationManager+Swift.swift
```diff
@@ -164,4 +167,53 @@ final class RadarLocationManagerSwift: NSObject {
...
         }
     }
+
+    @objc(replaceBubbleGeofenceOnLocationManager:location:radius:)
+    static func replaceBubbleGeofence(locationManager: CLLocationManager, location: CLLocation, radius: Int32) {
+        removeBubbleGeofence(locationManager: locationManager)
+
+        guard RadarSettings.tracking else {
+            return
+        }
```
nickdonnelly, line 179: Why do we call `removeBubbleGeofence()` _before_ the guard?
Seems like a slightly unexpected behavior to call something named "replace", only to have it remove the bubble geofence and return early.

stevepopovich: It's intentional I think. When tracking is off, the desired end state is no bubble geofence, so we clear the existing one and then bail out. 

Worth noting this also matches the existing ObjC behavior from before. I'm just porting it as-is (the ObjC version does the same remove-then-guard). In practice the guard is largely defensive anyway, since every caller already gates on if (tracking) at the call site in updateTracking.

I've added a comment in both the Swift method and the ObjC fallback to make the ordering's intent clear, it is actually a good question and worth noting about IMO

nickdonnelly (dismissed):

ShiCheng-Lu (dismissed):

alanjcharles14 (dismissed):

stevepopovich:
just need another review from the merge conflict

alanjcharles14 (approved):

PR #622, opened 2026-06-17
FENCE-2888: CI: consolidate into GitHub Actions (remove CircleCI)

stevepopovich:
## Summary

Consolidates the two overlapping PR gates (CircleCI + GitHub Actions) into a single GitHub Actions gate.

- **Removes CircleCI entirely** (`.circleci/config.yml`). No README badge or other references existed.
- **Adds the two CircleCI-only checks the team wants kept** to `.github/workflows/build-and-test.yml`:
  - **Build example app** — `make build-example-pretty`, with `DESTINATION` overridden to the dynamically-picked simulator UDID (the Makefile default `iPhone 17, OS=26.4.1` won't exist on the runner; this matches how the existing `ci-*` steps pass `CI_DESTINATION`).
  - **Lint podspecs** — `make lint` (`pod lib lint` over `RadarSDK.podspec` + `RadarSDKMotion.podspec`; CocoaPods is preinstalled on `macos-15`).

After this change, GitHub Actions covers: Swift lint, format check, build + static analyze, tests, swift tests, **podspec lint**, and **example build**.

## Not carried over
- **Jazzy docs generation** (CircleCI-only) is intentionally dropped — no longer used

## Follow-up (manual, outside this repo)
- If CircleCI is wired in as a **required status check** or via the GitHub app, that integration needs to be disabled in repo branch-protection / CircleCI settings. Deleting the config stops new CircleCI runs but doesn't remove the check requirement.

## Verification
- [ ] GitHub Actions "Xcode - Build, Analyze, & Test" runs green on this PR, including the new **Build example app** and **Lint podspecs** steps.
- [ ] CircleCI no longer reports a check.


alanjcharles14 (dismissed):

ShiCheng-Lu (dismissed):

Makefile
```diff
@@ -139,9 +139,16 @@ ci-test-swift:
 build-example-pretty:
 	set -o pipefail && xcodebuild $(XC_EXAMPLE_ARGS) | xcpretty
 
+ci-build-example:
+	@set -o pipefail; \
+	  xcodebuild -sdk iphonesimulator -project $(PROJECT_EXAMPLE).xcodeproj -scheme $(SCHEME_EXAMPLE) -destination "$(CI_DESTINATION)" ONLY_ACTIVE_ARCH=NO OTHER_CFLAGS="-fembed-bitcode" 2>&1 \
```
ShiCheng-Lu, line 144: by the style of the current make file, we should extract this to something like CI_XC_EXAMPLE_ARGS, 'iphonesimulator' should be $(SDK),

but we should just consolidate the builds so instead of having ci-build-example and build-example, we just have one command that's used everywhere. Let's (I can) do that in a different PR though because that'll be quite a bit more change + we don't want to keep blocking CI

ShiCheng-Lu (approved):

ShiCheng-Lu:
does the CircleCI Pipeline block PR? maybe we still need to do something to remove the dependency

Makefile
stevepopovich: yeah I totally agree and if you commit to cleaning that up, I would be ecstatic!

stevepopovich:
> does the CircleCI Pipeline block PR? maybe we still need to do something to remove the dependency

It does. We have to:
- Remove the CircleCI check as a merge requirement
- Go into CircleCI and remove the integration with our repo

PR #623, opened 2026-06-17
FENCE-2884: Fix campaign metadata in open app events

alanjcharles14:
## Summary

Fixes campaign metadata not being attached to `opened_app` conversion events when a user taps a client-side geofence notification (CSGN).

## Background

CSGN notifications are now built via the Swift path (`RadarNotification.swift`). Unlike the legacy Objective-C builder (which copied the full geofence metadata into the notification `userInfo`), the Swift path only carried `campaignId`, `url`, and `campaignMetadata`. As a result, when a CSGN was tapped, the resulting `opened_app` conversion was missing the campaign fields (`radar:campaignId`, `radar:campaignType`, `radar:notificationText`, `radar:notificationTitle`, etc.).

## Changes

- `RadarNotification.swift`: seed the notification `userInfo` with the full geofence metadata so all `radar:*` campaign fields flow through to the tap conversion (restores parity with the old Objective-C behavior).
- `RadarGeofence.swift`: add an `anyValue` accessor to `RadarMetadataValue` so metadata values can be written into `userInfo` as plain types.
- `Radar.m`: strip the internal, non-serializable `geofenceData` key from the metadata in `logConversionWithNotification` before logging the conversion, fixing the `ERROR_BAD_REQUEST` serialization failure.

## Testing

- Registered a CSGN on device, walked into the geofence to fire the notification, and tapped it.
- Confirmed the `opened_app` conversion is logged successfully with `conversionSource: radar_notification` and the campaign metadata fields present.


RadarSDK/Radar.m
```diff
@@ -702,7 +702,8 @@ + (void)logConversionWithNotification:(UNNotificationRequest *)request
                        deliveredAfter:(NSDate *)deliveredAfter {
     
     NSMutableDictionary *metadata = [[NSMutableDictionary alloc] initWithDictionary:request.content.userInfo];
-    
+    [metadata removeObjectForKey:@"geofenceData"];
```
stevepopovich, line 709: I would love to see some tests or even a code comment as to why we do `[metadata removeObjectForKey:@"geofenceData"];`. I don't think this code is self explanatory in that sense and first looking at it was ???

RadarSDK/RadarNotification.swift
```diff
@@ -69,12 +69,15 @@ extension RadarGeofenceSwift {
...
-        let userInfo: [String: Any] = [
+        var userInfo: [String: Any] = [
             "registeredAt": now.timeIntervalSince1970,
             "identifier": identifier,
             "geofenceId": id,
             "geofenceData": geofenceData,
         ]
+        for (key, value) in metadata {
+            userInfo[key] = value.anyValue
+        }
```
stevepopovich: Are we sure we want to arbitrarily map all of these? What if the contents of `userInfo` changes or ends up with private data we don't sent in requests?

Maybe we should just add to the above map with the fields we care about?

stevepopovich (commented):
Couple of questions but seems like the right idea!

RadarSDK/Radar.m
alanjcharles14: yeah fair enough lol. Added a comment.

RadarSDK/RadarNotification.swift
alanjcharles14: Also fair, tweaked the for loop to only include values prefixed with `radar:` so only the notification data is grabbed and maintains parity with the obj-c implementation.

stevepopovich (approved):
nice!

PR #627, opened 2026-06-19
FENCE-2887: Add day-of-week scheduling filter to CSGN

stevepopovich:
## Summary

Adds a day-of-week filter to client-side notifications (CSGN — geofence and beacon), building on the recently added campaign scheduling window (`radar:startsAt`/`radar:endsAt`) and operating-hours work.

A new metadata field **`radar:daysOfWeek`** holds a comma-separated list of day abbreviations (`"Sun"…"Sat"`, e.g. `"Tue,Thu"`). When present, the notification only fires on the listed days, evaluated in **device-local** time. Absent or empty → fires every day, matching the existing window convention.

> Requires the server to forward `radar:daysOfWeek` (and `radar:startsAt`/`radar:endsAt`) in the notification metadata — see [radarlabs/server#7980](https://github.com/radarlabs/server/pull/7980).

## Drive-by fix: flaky `RadarTrackingOptions` date equality

CI intermittently failed on the pre-existing `test_Radar_startTracking_custom`, unrelated to the day-of-week feature. Root cause is in `-[RadarTrackingOptions isEqual:]`:

- `dictionaryValue` serializes `startTrackingAfter`/`stopTrackingAfter` as `timeIntervalSince1970 * 1000`, and `initWithDictionary:` reads them back as `doubleValue / 1000`. `startTrackingWithOptions:` → `getTrackingOptions` round-trips through this.
- `isEqual:` compared the dates with a `< DBL_EPSILON` (~2.2e-16) tolerance. `DBL_EPSILON` is the gap between representable doubles near `1.0`; at timestamp magnitudes (~1.78e9) the real spacing is ~`1.78e9 * DBL_EPSILON` ≈ 4e-7s, so the comparison effectively demanded bit-exact equality.
- The `×1000 ÷1000` round-trip isn't bit-exact for ~1.9% of timestamps, so whether the test passed depended purely on the sub-second value of `[NSDate new]` — hence the flake.

Fix: widen the tolerance to `0.001` (1ms), matching the millisecond precision the dates are actually serialized at.

## Notes

- Metadata values are scalar (`RadarMetadataValue` is string/int/double/bool), so the day list is encoded as a single comma-separated string rather than an array.
- The shared predicate lives on the Objective-C `RadarNotificationHelper` (not the Swift bridge, which is an `actor`) and is tested from `RadarSDKTests.m`; the geofence ObjC block it replaces was previously untested.

## Companion PRs

- Server: [radarlabs/server#7980](https://github.com/radarlabs/server/pull/7980)
- Dashboard: [radarlabs/dashboard#745](https://github.com/radarlabs/dashboard/pull/745)

### DIFF
--- a/RadarSDK/RadarBeaconManager.m
+++ b/RadarSDK/RadarBeaconManager.m
@@ -154,6 +154,13 @@ static NSString *const kBeaconNotificationIdentifierPrefix = @"radar_beacon_noti
         
         if (region) {
             NSString *notificationId = [NSString stringWithFormat:@"%@%@", kBeaconNotificationIdentifierPrefix, uuid];
+            // Honor the campaign scheduling window / day-of-week filter, mirroring the geofence path, so a
+            // beacon notification isn't scheduled outside its window or on a non-selected day of the week.
+            if (![RadarNotificationHelper isNotificationActiveForMetadata:metadata now:[NSDate date]]) {
+                [[RadarLogger sharedInstance] logWithLevel:RadarLogLevelDebug
+                    message:[NSString stringWithFormat:@"Skipping beacon notification: outside campaign schedule window or day | identifier = %@", notificationId]];
+                continue;
+            }
             // Extract notification content from metadata
             UNMutableNotificationContent *content = [RadarNotificationHelper extractContentFromMetadata:metadata identifier:notificationId];
             if (content) {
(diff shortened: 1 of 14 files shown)


RadarSDK/RadarBeaconManager.m
```diff
@@ -154,6 +154,13 @@ - (void)registerBeaconRegionNotificationsFromArray:(NSArray<NSDictionary<NSStrin
         
         if (region) {
             NSString *notificationId = [NSString stringWithFormat:@"%@%@", kBeaconNotificationIdentifierPrefix, uuid];
+            // Honor the campaign scheduling window / day-of-week filter, mirroring the geofence path, so a
+            // beacon notification isn't scheduled outside its window or on a non-selected day of the week.
+            if (![RadarNotificationHelper isNotificationActiveForMetadata:metadata now:[NSDate date]]) {
```
stevepopovich, line 159: This was a bug I missed from campaign scheduling. Frankly, this was an edge case I missed. Campaigns that used a scheduling window and beacons would not honor the window and always fire regardless of the schedule 

However, campaign scheduling is not rolled out and the chance of someone wanting to use beacon based, CSGN based, campaigns with scheduling soon is low enough that it did not warrant a hot fix.

PR #628, opened 2026-06-24
FENCE-2887: Test CSGN scheduling-window deregistration on refresh

stevepopovich:
## Summary
Adds test coverage locking in **scheduling-window deregistration** for client-side geofence notifications (CSGNs). The behavior already exists end-to-end across two merged PRs:
- `RadarGeofenceSwift.toNotificationRequest(now:)` gates registration on `isWithinSchedulingWindow`, which reads `radar:startsAt`/`radar:endsAt` from campaign metadata (#618).
- `RadarNotificationHelper.refreshGeofenceNotifications()` re-runs that over the durable cache on a `radar:refreshNotifications` push (#617), and `RadarGeofenceSwift` persists `metadata`, so the window survives in the cache.

The only gap was a test proving a refresh actually **prunes** a window-expired campaign (previously only operating-hours refresh was covered).

Pairs with the server change that emits `radar:startsAt`/`radar:endsAt` and sends the `radar:refreshNotifications` nudge when a campaign's window ends — companion server PR: https://github.com/radarlabs/server/pull/8016

Also adds a new tab for inspecting CSGNs!

## Changes
- `RadarNotificationHelperTest.swift`: extend `makeGeofenceDict` with `startsAt`/`endsAt`; add tests that a refresh prunes a campaign whose `radar:endsAt` has passed and keeps one still within its window.
- `RadarNotificationHelper.swift`: clarify the cache-persist comment (refresh re-evaluates the scheduling window alongside operating hours). No behavior change.

<details><summary>CSGN Inspector</summary>
<p>

| | |
|--------|--------|
| Screenshot 2026-06-26 at 1 52 30 PM | Screenshot 2026-06-26 at 1 53 08 PM | 

</p>
</details>

PR #629, opened 2026-06-25
FENCE-2887: Add day-of-week filter to CSGN (geofence path)

stevepopovich:
## Summary
Re-implements the day-of-week CSGN filtering from [#627](https://github.com/radarlabs/radar-sdk-ios/pull/627) (CLOSED) on top of [#628](https://github.com/radarlabs/radar-sdk-ios/pull/628), **geofence path only** (no beacon).

A campaign's `radar:daysOfWeek` metadata (comma-separated abbreviations, e.g. `"Tue,Thu"`) restricts a client-side geofence notification to those local weekdays. The check runs in `RadarGeofenceSwift.toNotificationRequest`, which `RadarNotificationHelper.refreshGeofenceNotifications()` re-runs over the durable cache on a `radar:refreshNotifications` push — so when the server nudges on a day rollover, the device prunes the now-inactive CSGN without a `/track`.

## Companion PR
- Server: [radarlabs/server#8022](https://github.com/radarlabs/server/pull/8022) — emits `radar:daysOfWeek`, day-gates enrichment, and nudges on a day rollover. See testing steps there.

### DIFF
--- a/RadarSDK/RadarNotificationHelper.m
+++ b/RadarSDK/RadarNotificationHelper.m
@@ -29,6 +29,9 @@ static dispatch_semaphore_t notificationSemaphore;
 }
 
 + (BOOL)isNotificationActiveForMetadata:(NSDictionary *)metadata now:(NSDate *)now {
+    // No (or non-dictionary) metadata means no scheduling constraint, so the notification is
+    // active — the same fail-open default as the absent-value cases below and the Swift path's
+    // isActiveOnDayOfWeek. Callers (e.g. RadarLocationManager) already guard against nil.
     if (![metadata isKindOfClass:[NSDictionary class]]) {
         return YES;
     }
--- a/RadarSDK/RadarNotificationHelper.m
+++ b/RadarSDK/RadarNotificationHelper.m
@@ -28,6 +28,64 @@ static dispatch_semaphore_t notificationSemaphore;
     }
 }
 
++ (BOOL)isNotificationActiveForMetadata:(NSDictionary *)metadata now:(NSDate *)now {
+    if (![metadata isKindOfClass:[NSDictionary class]]) {
+        return YES;
+    }
+
+    // "yyyy-MM-dd'T'HH:mm:ss.SSS" is 23 chars; prefix to this length strips any trailing timezone suffix so the string is parsed as wall-clock local time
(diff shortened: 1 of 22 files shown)


RadarSDK/RadarNotificationHelper.m
```diff
@@ -28,6 +28,64 @@ + (void)initialize {
     }
 }
 
++ (BOOL)isNotificationActiveForMetadata:(NSDictionary *)metadata now:(NSDate *)now {
+    if (![metadata isKindOfClass:[NSDictionary class]]) {
+        return YES;
+    }
```
nickdonnelly, line 37: Is this a condition we ever expect to hit? This seems like a case of invalid input, which I would expect to throw an error rather than silently mark the notification as active

RadarSDK/RadarOperatingHoursEvaluator.swift
```diff
@@ -46,6 +46,19 @@ enum RadarOperatingHoursEvaluator {
...
+        var calendar = Calendar(identifier: .gregorian)
+        calendar.timeZone = timeZone
+
+        let weekdayIndex = calendar.component(.weekday, from: date) - 1
+        guard weekdayIndex >= 0, weekdayIndex < daysOfWeekAbbr.count else {
+            return nil
+        }
+        return daysOfWeekAbbr[weekdayIndex]
+    }
+
```
nickdonnelly, line 64: Could we just use the localized version and not have to maintain `daysofWeekAbbr`?

```swift
static func weekdayAbbreviation(
    for date: Date,
    timeZone: TimeZone = .current
) -> String {
    let formatter = DateFormatter()
    formatter.timeZone = timeZone
    formatter.dateFormat = "E"

    return formatter.string(from: date)
}
```

RadarSDK/RadarNotificationHelper.m
stevepopovich: Not invalid input, added a code comment explaining, lemme know what you think 👍🏼

RadarSDK/RadarOperatingHoursEvaluator.swift
stevepopovich: Need to match the server, added a comment, does that make sense?

alanjcharles14 (approved):

nickdonnelly (approved):

PR #630, opened 2026-06-29
FENCE-2892: Add recursive JSON sanitizer

alanjcharles14:
## Summary
Hardens JSON serialization in `RadarUtils` so a non-JSON value (e.g. `NSData`, `NSDate`, `NaN/Inf`) anywhere in a request payload can never crash the host app or silently drop the entire payload.

## Manual test steps
just a json serializer covered by the unit tests. 

## Checklist

### DIFF
--- a/RadarSDK/RadarAPIHelper.m
+++ b/RadarSDK/RadarAPIHelper.m
@@ -130,9 +130,9 @@ static NSTimeInterval RadarAPIHelperExtendedNetworkTimeoutInterval(NSTimeInterva
                         requestParams[@"replays"] = updatedReplays;
                     }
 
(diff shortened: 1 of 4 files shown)


stevepopovich (dismissed):
Seems like the original person had repro steps, were you able to repro consistently? Looks ok though

alanjcharles14:
> Seems like the original person had repro steps, were you able to repro consistently? Looks ok though

Yeah, I'm not sure the report was correct. It should be fixed in `3.31.0` based on everything I've looked into but if there is invalid json we just drop the entire metadata obj and return `{}`. This should make it so we only drop invalid values instead of blowing up the entire obj.

stevepopovich (approved):
fair enough then 👍🏼

PR #633, opened 2026-06-30
Guard json getter

### LINKED ISSUE #613: Crash in `RadarUtils.dictionaryToJson` when `radar-registeredNotifications` UserDefaults cache contains non-JSON userInfo values
**Affected versions:** 3.33.1

### Crash

```
*** Terminating app due to uncaught exception 'NSInvalidArgumentException',
    reason: 'Invalid type in JSON write (__NSCFData)'

0  CoreFoundation     __exceptionPreprocess
1  libobjc            objc_exception_throw
2  Foundation         _writeJSONValue
…  Foundation         _writeJSONObject -> _writeJSONArray -> _writeJSONObject
16 Foundation         +[NSJSONSerialization dataWithJSONObject:options:error:]
17 RadarSDK           specialized static RadarUtils.dictionaryToJson(_:)   RadarUtils.swift:200
20 RadarSDK           -[RadarAPIHelper requestWithMethod:...:logPayload:...]_block_invoke   RadarAPIHelper.m:61
```

### Root cause

1. Radar's `useNotificationDiff` feature persists a snapshot of pending iOS local notifications under the `radar-registeredNotifications` `UserDefaults` key. Each entry stores the notification's `userInfo` dictionary verbatim.
2. `UNNotificationContent.userInfo` is a plist-typed dictionary, not a JSON-typed one. Any host app that puts a `Data` value into `userInfo` (a common pattern — e.g. an archived `NSKeyedArchiver` payload) produces a snapshot entry whose `info` key is an `NSData`.
3. On the next `/v1/track`, Radar reads that snapshot back, includes it in the request params (via `notificationDiff`), and calls `RadarUtils.dictionaryToJson(_:)`, which calls `+[NSJSONSerialization dataWithJSONObject:options:error:]`. JSONSerialization throws `NSInvalidArgumentException ("Invalid type in JSON write (__NSCFData)")` because `Data` is not a valid JSON type.
4. The exception is thrown synchronously through `RadarUtils.dictionaryToJson` (Swift). Swift `do/try/catch` cannot catch ObjC `NSException`, so the "FENCE-2791: Add guard for NSException in dictionaryToJson" change from 3.33.1 has no effect on this path — the exception unwinds straight through and aborts the host app.
5. Because the crash happens *before* Radar can overwrite the snapshot with a new (clean) diff, the cache stays poisoned forever. Every subsequent track call crashes again. This is a permanent wedge — affected users have to be migrated by the host app explicitly removing the `radar-registeredNotifications` key.

### Evidence

Inspecting the device's `UserDefaults` on an affected device shows the snapshot under `radar-registeredNotifications` contains an `<array>` of `<dict>` entries where `info` is `<data>…</data>`.

alanjcharles14:
## Summary
Guards `RadarState.registeredNotifications`'s getter against a crash when the `radar-registeredNotifications` UserDefaults snapshot contains non-JSON values. A host app can put a `Data` value into a local notification's `userInfo`. Builds prior to 3.29 persisted that `userInfo` verbatim, so devices that upgraded still carry legacy snapshots containing `Data`. On read, the getter calls `JSONSerialization.data(withJSONObject:)` directly on that snapshot, which raises an Objective-C `NSException` ("Invalid type in JSON write"). Swift's `try?` can't catch an `NSException`, so the app crashes on every read until the host app clears the key which causes a permanent wedge that persists across SDK upgrades (reported in #613).

## Changes
- Add a `JSONSerialization.isValidJSONObject(obj)` check before serializing in the
  getter. A poisoned snapshot now returns `nil` instead of crashing, and is
  overwritten with a clean value on the next notification registration.
- Add `RadarStateTests` covering the poisoned-cache case (no crash, returns `nil`)
  and a valid snapshot round-trip.

## Related issue

Resolves #613

## Manual test steps
More json guarding, covered in unit tests.

## Checklist

### DIFF
--- a/RadarSDK.xcodeproj/project.pbxproj
+++ b/RadarSDK.xcodeproj/project.pbxproj
@@ -155,18 +155,14 @@
 		B0A0F0142FBC000100111111 /* RadarLocationManagerSwiftTestHelpers.swift in Sources */ = {isa = PBXBuildFile; fileRef = B0A0F0152FBC000100111111 /* RadarLocationManagerSwiftTestHelpers.swift */; };
 		B0A0F0162FBC000100111111 /* TrackingCLLocationManager.swift in Sources */ = {isa = PBXBuildFile; fileRef = B0A0F0172FBC000100111111 /* TrackingCLLocationManager.swift */; };
 		B0A0F0182FBC000100111111 /* RadarLocationManagerSwiftBeaconSyncTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = B0A0F0192FBC000100111111 /* RadarLocationManagerSwiftBeaconSyncTests.swift */; };
+		B0A0F0202FBC000100111111 /* RadarLocationManagerSwiftRegionTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = B0A0F0212FBC000100111111 /* RadarLocationManagerSwiftRegionTests.swift */; };
 		BA353F092F8826B500E553A1 /* RadarSyncTestHelper.m in Sources */ = {isa = PBXBuildFile; fileRef = BA353F082F8826AF00E553A1 /* RadarSyncTestHelper.m */; };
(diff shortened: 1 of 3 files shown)


stevepopovich (approved):
Yeah fun. I guess we could consider writing logic that say `if the JSON in valid, delete it`.  SO the cache gets unbroked for next time

PR #634, opened 2026-06-30
FENCE-2911: Diff CSGN notifications instead of tearing down and re-adding

stevepopovich:
## Problem

A client-side geofence notification (CSGN) syncs down and registers, but never fires on a real walk-through — even with `Always` location auth, notifications authorized, no scheduling-window/day-of-week/operating-hours filters, and only one region registered.

`replaceSyncedGeofences` runs on the completion of **every track**. Both the V2 Swift path (`registerNotifications`) and the legacy ObjC path (`updateClientSideCampaignsWithPrefix`) removed *all* pending geofence notifications and re-added them, even when nothing changed.

Re-adding a `UNLocationNotificationTrigger` re-arms it, and **iOS will not fire an entry for a trigger scheduled while the device is already inside the region**. During a walk-through, frequent location updates drive frequent tracks, so a re-registration almost always lands while the device is inside the geofence — silently killing the pending entry so the notification never fires.

Basically, this is probably an edge case but it:
1. Makes testing locally more reliable
2. Will make CSGNs more reliable if integrators are spamming `track`

I would like to release this as a patch version. Wdyt?
## Fix

Per review, this is now **Swift-only**: the diff lives entirely in the V2 path (`RadarNotificationHelper.registerNotifications`), gated behind the existing `useNotificationDiffV2` SDK config flag (default off). The legacy ObjC `updateClientSideCampaignsWithPrefix:` path (used when the flag is off, and for beacon campaigns, which the Swift path doesn't handle) is unchanged from its original tear-down/re-add behavior, so it doubles as a kill switch if the diff logic misses something.

The Swift path diffs desired vs. already-pending requests **by direct field comparison**, not a string signature:

- for each desired notification, look up any pending request with the same identifier,
- compare content (title/subtitle/body), the location trigger's circular region (center, radius, entry/exit, repeats), and `userInfo`,
- unchanged ones are left in place (preserving their arm point); gone/changed ones are removed and re-added.

`userInfo` (which carries the campaign metadata) is compared directly, excluding two volatile keys that are rebuilt on every call and would otherwise make every request look changed: `registeredAt` (fresh timestamp) and `geofenceData` (a `JSONEncoder` blob whose key order isn't byte-stable across app launches — its meaningful fields are already covered by the flat `radar:*` keys, content, and region). This means a changed campaign metadata value now correctly triggers a re-registration.


alanjcharles14:
makes sense to me, Happy to approve but I messaged @ShiCheng-Lu asking him to take a look just to make sure there's no additional context we're missing since he did most of the original notification diff work.

RadarSDK/RadarNotificationHelper.m
```diff
@@ -367,39 +368,86 @@ + (void)logConversionWithNotificationResponse:(UNNotificationResponse *)response
     }
 }
 
++ (NSString *)notificationUniqueIdentifierForRequest:(UNNotificationRequest *)request {
```
ShiCheng-Lu: UNNotificationRequest and String should both map perfectly between obj-c + swift, so I think we should be able to use swift for this function, unless there's something.

ShiCheng-Lu: oh I see, for new functions, we don't need to keep an obj-c copy.

RadarSDK/RadarNotificationHelper.m
```diff
@@ -367,39 +368,86 @@ + (void)logConversionWithNotificationResponse:(UNNotificationResponse *)response
...
+            [parts addObject:region.proximityUUID.UUIDString ?: @""];
+            [parts addObject:region.major ? region.major.stringValue : @""];
+            [parts addObject:region.minor ? region.minor.stringValue : @""];
+            [parts addObject:region.notifyOnEntry ? @"1" : @"0"];
+            [parts addObject:region.notifyOnExit ? @"1" : @"0"];
+            [parts addObject:trigger.repeats ? @"1" : @"0"];
+        }
+    }
+    return [parts componentsJoinedByString:@"|"];
+}
```
ShiCheng-Lu: why do we use a string encoding instead of comparing the notification request? you can get the request from the pending list.

ShiCheng-Lu (commented):
looks good, I think the string encoding is a bit sus compared to just comparing request object, get by ID then compare rest of the fields. 

I think we should only do this for swift + leave obj-c as is. so that if there is something that we missed, we can swap it back to obj-c behaviour with the SDK config flag, so we don't need to add another flag for this, otherwise I think we might want to add a flag.

RadarSDK/RadarNotificationHelper.swift
```diff
@@ -155,6 +166,27 @@ actor RadarNotificationHelper: NSObject {
...
+                parts.append(region.proximityUUID.uuidString)
+                parts.append(region.major?.stringValue ?? "")
+                parts.append(region.minor?.stringValue ?? "")
+                parts.append(region.notifyOnEntry ? "1" : "0")
+                parts.append(region.notifyOnExit ? "1" : "0")
+                parts.append(trigger.repeats ? "1" : "0")
+            }
+        }
+        return parts.joined(separator: "|")
+    }
```
ShiCheng-Lu, line 231: I think we also might need to compare userInfo (the campaign metadata).

if someone changes the campaign metadata, we should re-register the notification with the new metadata.

ShiCheng-Lu: metadata is always string: int/string/bool, so don't need to worry about nested obj comparison.

RadarSDK/RadarNotificationHelper.m
stevepopovich: Fair, changed it

RadarSDK/RadarNotificationHelper.m
stevepopovich: fair, didn't know it was all rolled out

RadarSDK/RadarNotificationHelper.swift
stevepopovich: Done, lemme know what you think of the comment

alanjcharles14 (approved):

PR #611, opened 2026-06-01
FENCE-2857: Add NSError parameter to ipGeocode callback

stevepopovich:
Non-broken version of #608 

Ports [radarlabs/radar-sdk-android#525](https://github.com/radarlabs/radar-sdk-android/pull/525) to iOS.

## Summary

- Adds a new four-arg `+ipGeocodeWithErrorCompletionHandler:` (typedef `RadarIPGeocodeWithErrorCompletionHandler`) that exposes the underlying `NSError` when the failure originated from a caught network/parse/exception error, so integrators can forward it to error collectors (Sentry, Crashlytics, Datadog).
- Threads the caught `NSError` through `RadarAPIHelper`'s three error paths (`NSURLSession` error, `NSJSONSerialization` error, `@catch NSException`) and through the internal `RadarIPGeocodeAPICompletionHandler`. The parse and exception paths also gain a `RadarLogger` error log that they were previously missing.
- Swift consumers see only the new four-arg block under the existing `Radar.ipGeocode { ... }` call site (via `NS_SWIFT_NAME`).

## Backwards compatibility

- The legacy three-arg `+ipGeocodeWithCompletionHandler:` is preserved, marked `__attribute__((deprecated))`, and routes through the new method (ignoring the error). Existing ObjC integrators keep compiling and keep receiving callbacks.
- The internal generic `RadarAPICompletionHandler` block typedef gains an `NSError *` (third arg). Every callsite in `RadarAPIClient.m` was updated to take and pass-through the new parameter even though only `ipGeocode` consumes it — mirrors the Android PR and primes future extension to other callbacks.
- A `test_Radar_ipGeocode_legacy_three_arg_callback_still_called` test locks the BC behavior in.

Scope is intentionally limited to `ipGeocode` as the pilot. Version bump / MIGRATION will be handled in a separate PR.

## Test plan

- [ ] Manually: airplane-mode the Example app, tap `ipGeocode`, confirm the console prints `error = The Internet connection appears to be offline.`

PR #636, opened 2026-07-01
FENCE-2913: Fix "Ambiguous use of ipGeocode()" for Swift async callers

alanjcharles14:
## Summary
Restores the Swift `async` form of `Radar.ipGeocode()`, which broke in **3.34.1**. Callers of `await Radar.ipGeocode()` were hitting a compile error: *"Ambiguous use of `ipGeocode()`"*.

## Root cause
#611 (FENCE-2857, shipped in 3.34.1) added `+ipGeocodeWithErrorCompletionHandler:` alongside the existing `+ipGeocodeWithCompletionHandler:`. Both map to the same Swift name, so Swift synthesized two `ipGeocode() async` overloads that differ only by `throws`, making every async call site ambiguous and uncompilable.

## Fix
- Add `NS_SWIFT_DISABLE_ASYNC` to `+ipGeocodeWithErrorCompletionHandler:` so only the original non-throwing `ipGeocode() async` is synthesized (restores the pre-3.34.1 signature, no `try` needed).
- Remove the `deprecated` attribute from `+ipGeocodeWithCompletionHandler:` so async callers don't get a deprecation warning (async now resolves to this method).

Both closure APIs are unchanged; the 4-arg error variant remains usable as a closure. No compiling code relied on the async form, since it hasn't compiled since 3.34.1.

## Testing
- New `RadarIPGeocodeAsyncTests.swift` regression guard calls `await Radar.ipGeocode()` (ambiguity = build failure).
- Existing ObjC closure/error/legacy tests still pass.
- `make test-pretty`

## Checklist

### DIFF
--- a/RadarSDK.xcodeproj/project.pbxproj
+++ b/RadarSDK.xcodeproj/project.pbxproj
@@ -186,6 +186,7 @@
 		BA8647B92F6C7DB700F1A199 /* RadarSyncManagerTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = BA8647B82F6C7DAD00F1A199 /* RadarSyncManagerTests.swift */; };
 		BA8647BB2F6C8F0E00F1A199 /* SyncRegionResponse.swift in Sources */ = {isa = PBXBuildFile; fileRef = BA8647BA2F6C8F0600F1A199 /* SyncRegionResponse.swift */; };
 		BAC932F02FF43E8100503C2C /* RadarStateTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = BAC932EF2FF43E7400503C2C /* RadarStateTests.swift */; };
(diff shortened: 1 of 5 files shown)


stevepopovich (approved):
ugh Im so sorry. I had no idea. Thanks for writing a regression test.

I am still green on this, but could we prompt AI to write a lint rule maybe to catch this in the future? I really have no idea.

alanjcharles14:
@stevepopovich no stress man! I'm honestly not sure how we would've caught this one, your original PR had multiple approvals so no one else caught it either. Just one of those things we'll have to keep an eye on as we move forward on the swift conversion. Also it's not necessarily a bad change, it's more that we released it as a patch without calling out the "new" public API.

stevepopovich:
Yeah I am bad with semvar, gotta get better

PR #637, opened 2026-07-02
Remove `useNotificationDiff` and `useNotificationDiffV2` flags

stevepopovich:
## Summary
- Removes `useNotificationDiffV2`, which has been at 100% rollout for 5+ months: collapses the V1/V2 branches in `RadarAPIClient.m`, `RadarNotificationHelper.m`, and `RadarLocationManager.m` down to the V2 (Swift actor) notification-diff path, and adds an `XCTestCase` guard to the two call sites that were newly-unconditional on `RadarNotificationHelper_Swift.shared` (touching `UNUserNotificationCenter` crashes in the test host without it).
- Removes `useNotificationDiff`, also fully rolled out: the notification diff is now always computed before a track request instead of being gated behind the flag.
- Removes both flags from `RadarSdkConfiguration` (Swift + ObjC header) and from the Example app's debug settings display.

## Flaky-test fix (surfaced by this branch's CI run)
CI failed on `test_Radar_startTracking_custom`, but the failure was a pre-existing flake unrelated to the flag removal — the notification changes are test-neutral (the newly-unconditional notification-diff path is guarded by `XCTestCase`).

`RadarTrackingOptions.isEqual` compared the `startTrackingAfter` / `stopTrackingAfter` dates with `fabs(diff) < DBL_EPSILON`. Those dates round-trip through millisecond serialization in `-dictionaryValue` (`timeIntervalSince1970 * 1000`, then `/ 1000` on the way back). `DBL_EPSILON` (~2.2e-16) is far smaller than a single ULP at Unix-timestamp magnitudes (~2.4e-7 near 1.75e9), so a serialized-then-deserialized copy compared unequal on ~1.9% of timestamps per field — flaking this test (the only tracking test that sets dates) on ~3.8% of runs, purely depending on the wall-clock time when it ran. Fixed by comparing at the millisecond precision the dates are actually stored with (1 ms tolerance).

I have run into this multiple times, just haven't made the fix yet.

## Manual testing (iOS Simulator)
Verifies the notification-diff path works in a real app run (not just under `XCTest`, where the new `RadarNotificationHelper_Swift.shared` call sites are guarded out).

1. On the **Debug** tab, confirm debug logs are streaming.
2. On the **Map** tab → Tracking panel, tap **startTracking (continuous)** and confirm a track request fires in the Debug logs.
3. **Diff always computed:** confirm the track request params in the Debug logs include a `notificationDiff` key on every track (previously gated behind `useNotificationDiff`).
4. **Registration:** simulate a location near your CSGN geofence (Xcode → Debug → Simulate Location, or the bundled `Example/Example/waypoints.gpx`). After the next track/sync, open the **CSGN** tab and refresh — confirm the geofence's notification is listed as registered (title/body/campaign shown; "inside/outside" reflects the simulated location).
5. **Delivery:** exit and re-enter the geofence region (iOS only fires an entry trigger after you leave and re-enter). Verify the local notification is delivered.
6. **Diff, not teardown/re-add:** trigger another sync (move location and let it re-sync, or stopTracking → startTracking). Confirm in the Debug logs that unchanged CSGNs are diffed rather than all removed and re-registered — the `registeredAt` timestamps for unchanged notifications on the **CSGN** tab should stay stable across syncs.
7. **No crash:** confirm the app does not crash at the now-unconditional `RadarNotificationHelper_Swift.shared` call sites (geofence registration on sync, notification removal on replayed tracks) — these run in the real app but are guarded out under `XCTest`.


alanjcharles14 (dismissed):
looks correct to me, no notes

RadarSDK/RadarAPIClient.m
```diff
@@ -588,16 +574,20 @@ - (void)makeTrackRequestWithParams:(NSDictionary *)params
...
 
-                                    if ([RadarSettings sdkConfiguration].useNotificationDiffV2) {
+                                    // Skip notification removal under XCTest. RadarNotificationHelper_Swift.shared resolves its
+                                    // notification center from UNUserNotificationCenter.current(), which has no app bundle or
+                                    // notification entitlements in the unit-test host and misbehaves there. This call used to be
+                                    // gated behind the useNotificationDiffV2 flag (false in tests); removing the flag made it
+                                    // unconditional, so the guard preserves the prior "skip in tests" behavior. This matches the
+                                    // existing XCTestCase guards elsewhere in the SDK. TODO: route through the injectable
+                                    // NotificationCenterProtocol seam instead so tests can use a mock and this guard can be deleted.
+                                    if (NSClassFromString(@"XCTestCase") == nil) {
```
nickdonnelly, line 584: Not a fan of production code checking if we're running inside of a test.

RadarSDK/RadarLocationManager.m
```diff
@@ -595,13 +595,20 @@ - (void)replaceSyncedGeofences:(NSArray<RadarGeofence *> *)geofences {
...
 
-    if ([RadarSettings sdkConfiguration].useNotificationDiffV2) {
+    // Skip notification registration under XCTest. RadarNotificationHelper_Swift.shared resolves its
+    // notification center from UNUserNotificationCenter.current(), which has no app bundle or
+    // notification entitlements in the unit-test host and misbehaves there. This call used to be
+    // gated behind the useNotificationDiffV2 flag (false in tests); removing the flag made it
+    // unconditional, so the guard preserves the prior "skip in tests" behavior. This matches the
+    // existing XCTestCase guards elsewhere in the SDK. TODO: route through the injectable
+    // NotificationCenterProtocol seam instead so tests can use a mock and this guard can be deleted.0
+    if (NSClassFromString(@"XCTestCase") == nil) {
```
nickdonnelly, line 605: same

nickdonnelly (approved):

RadarSDK/RadarAPIClient.m
stevepopovich: I agree but this is already a prescedent. Something worth looking at when we move to swift. Big red flag to me too

PR #515, opened 2025-12-17
ML beacons

ShiCheng-Lu:

PR #641, opened 2026-07-07
FENCE-2904: Add RadarIndoors ML engine and Swift delegate/invoke plumbing

stevepopovich:
## What this is

Second of the smaller PRs carved out of `shicheng/ml-beacons` (existing PR #515), **stacked on top of #640**. Review/merge #640 first — this PR's base is `stevepopovich/indoors-protocol-and-submodule`, not `master`, and it depends on the `activeIndoorModelId` geofence property that #640 adds.

This adds the **`RadarIndoors` ML engine wrapper** — the Swift-side glue that drives the optional `RadarSDKIndoors` framework. It is **inert at runtime**: nothing calls it yet. Wiring it into the tracking pipeline (`RadarLocationManager` / `Radar` / `RadarAPIClient`) is the next PR.

No manual test steps needed

## Fixes made while extracting (both pre-existing)

1. `RadarDelegateActor.shared` was mistakenly initialized as `RadarIndoorsActor()` (copy-paste) — it "compiled" but conflated two global actors. Now `RadarDelegateActor()`.
2. The `getModelData` callback bridged async→sync with a `Task {}` writing a shared mutable `var` read after `semaphore.wait()` — a data race the **Swift 6 compiler rejects** (hard build error under this repo's `swift_version = 6.0`). Reworked to a detached `Task` writing into a reference box, with the read ordered after the write by the semaphore. The synchronous framework contract and the cache-miss laziness are both preserved. **Follow-up:** the real fix is an async data-provider API on `RadarSDKIndoors` so the blocking wait can go away entirely — noted for a later change to that framework.

## Follow-up PRs

3. **Wire `RadarIndoors` into tracking** — `RadarLocationManager`/`Radar`/`RadarAPIClient` hot-path changes.
4. **Example app** — indoor survey/calibration tooling.


RadarSDK/RadarIndoors.swift
```diff
@@ -0,0 +1,179 @@
...
+@available(iOS 13.0, *)
+public actor RadarIndoorsActor {
+    public static let shared = RadarIndoorsActor()
+}
+
+@RadarIndoorsActor
+@available(iOS 13.0, *)
+class RadarSDKIndoors {
+    let instance: NSObject?
+    init? () {
```
alanjcharles14: The else branch sets `instance = nil` but the initializer still succeeds, so `let sdk = RadarSDKIndoors()`below is always non-nil. That makes the `guard let sdk` in `updateTracking` dead code when the framework isn't linked and `useIndoorScan` is on, the intended `"RadarIndoors class is nil" `warning never fires, and we fall through and set `currentModelId` anyway. Either make it truly fail-able `(else { return nil })` so the guard means what it says, or drop the `?` and gate on `instance != nil`explicitly.

RadarSDK/RadarIndoors.swift
```diff
@@ -0,0 +1,179 @@
...
+    public func getLocation() async -> CLLocation? {
+        guard let sdk else {
+            return nil
+        }
+        return await sdk.getLocation()
+    }
+
+    // Downloads the mlmodel asset for a geofence's active indoor model to a temp file.
+    // nonisolated so it can run on a detached Task without hopping onto RadarIndoorsActor.
+    nonisolated static func downloadModelData(modelId: String) async -> URL? {
```
alanjcharles14, line 164: These `.mlmodel` blobs get written to `NSTemporaryDirectory()` with a unique uuid name and never deleted. iOS only purges tmp opportunistically, not within a session, so repeated model switches accumulate sizable files. Is it possible to delete a model once the framework's done consuming it, or write to a deterministic per-model path and overwrite?

RadarSDK/RadarIndoors.swift
```diff
@@ -0,0 +1,179 @@
...
+        if user.geofences?.contains(where: { $0.activeIndoorModelId != nil && ($0.activeIndoorModelId == currentModelId) }) == true {
+            RadarLogger.shared.debug("model already in use")
+            return
+        }
+        guard let modelId = user.geofences?.first(where: { $0.activeIndoorModelId != nil })?.activeIndoorModelId else {
+            // no model id in current geofences
+            RadarLogger.shared.debug("found no model id in current geofences")
+            return
+        }
+        currentModelId = modelId
```
alanjcharles14: Set `currentModelId` only after `start()` succeeds and clear it to `nil` after `stop()` so we don't get false positives on the next `updateTracking` and it actually restarts when a geofence is re-enabled.

stevepopovich: Yeah AI caught this for me. I think in practice it won't actually matter but willing to fix it

RadarSDK/RadarIndoors.swift
stevepopovich: We can. In practice, I doubt the models will switch very often, but worth purging I suppose

alanjcharles14 (approved):

PR #642, opened 2026-07-08
FENCE-2905: Wire RadarIndoors into the tracking pipeline

stevepopovich:
## What this is

Third of the smaller PRs carved out of `shicheng/ml-beacons` (existing PR #515), **stacked on top of #641**. Base is `stevepopovich/radar-indoors-engine`, not `master` — review/merge #640 → #641 first.

This is the PR that actually **wires the `RadarIndoors` ML engine into the track pipeline**, replacing the old beacon indoor-scan path with the ML indoor-location path. It's the highest-risk PR of the stack because it touches the hot path every SDK integrator runs, so I hand-reconciled each change against current `master`'s structure rather than copying from the source branch.

## Manual test steps

Highest-risk PR — verify on a real device / staging project, not just unit tests.
1. **Regression, indoor off** (no `activeIndoorModelId` / `useIndoorScan = false`): run continuous tracking start/stop, a background update, `trackOnce()`, and `trackVerified`. Confirm `/track` requests are unchanged from `master` — no `indoorMLLatitude`/`indoorMLLongitude` in `locationMetadata`, coordinates untouched. This is the case every existing customer hits.
2. **Indoor on** (`useIndoorScan = true`, geofence with a real `activeIndoorModelId`): walk into the geofence (or simulate) and confirm via a logging proxy / breakpoint in `RadarAPIClient.m` that the `/track` `locationMetadata` contains `indoorMLLatitude`/`indoorMLLongitude` **and** `deviceLatitude`/`deviceLongitude`, and the top-level `latitude`/`longitude` are the indoor-predicted values. Confirm the app's `RadarDelegate.didUpdateClientLocation(_:stopped:source:)` fires with source `.indoors`.
3. Leave the geofence / flip `useIndoorScan` off mid-session — confirm `RadarIndoors` stops cleanly and tracking falls back to device location, no crash.
4. Background/foreground during an active indoor session — no crash; scanning suspends/resumes.

## Follow-up PR

4. **Example app** — indoor survey/calibration tooling (MapLibre, AR, survey upload).


stevepopovich:
@ShiCheng-Lu would love a review here as I close in on the hot path and get this closer to being merged

ShiCheng-Lu:
> @ShiCheng-Lu would love a review here as I close in on the hot path and get this closer to being merged

kinda busy with reveal rn, I'll review tmr if that's ok

stevepopovich:
Absolutely, no rush at all @ShiCheng-Lu

RadarSDK/RadarAPIClient.m
```diff
@@ -453,7 +450,17 @@ - (void)trackWithLocation:(CLLocation *_Nonnull)location
         }
     }
 
-    if (options.usePressure || options.useMotion) {
+    if (options.useIndoorScan && indoorLocation) {
+        locationMetadata[@"indoorMLLatitude"] = @(indoorLocation.coordinate.latitude);
```
alanjcharles14, line 456: we should also wrap these in a conditional to make sure the coords we're replacing are valid

```
    CLLocationCoordinate2DIsValid(indoorLocation.coordinate) &&
    (indoorLocation.coordinate.latitude != 0.0 || indoorLocation.coordinate.longitude != 0.0)) {
    locationMetadata[@"indoorMLLatitude"] = @(indoorLocation.coordinate.latitude);
    locationMetadata[@"indoorMLLongitude"] = @(indoorLocation.coordinate.longitude);
  .... etc.
}

File: RadarSDK/RadarAPIClient.m [hunk 2]
```diff
@@ -453,7 +450,17 @@ - (void)trackWithLocation:(CLLocation *_Nonnull)location
...
+        locationMetadata[@"indoorMLLongitude"] = @(indoorLocation.coordinate.longitude);
+        locationMetadata[@"deviceLatitude"] = params[@"latitude"];
+        locationMetadata[@"deviceLongitude"] = params[@"longitude"];
+        // CRTICAL: Overwrite the reported coordinates with the indoor ML model's prediction.
```
alanjcharles14: NIT: `CRITICAL` missing `i`

RadarSDK/RadarLocationManager.m
```diff
@@ -1047,31 +1048,23 @@ - (void)handleLocation:(CLLocation *)location source:(RadarLocationSource)source
...
-                        withKnownLocation:location
-                        completionHandler:^(NSString *_Nullable indoorScanResult, CLLocation *_Nullable locationAtStartOfScan) {
-            [[RadarLogger sharedInstance] logWithLevel:RadarLogLevelDebug
-                               message:[NSString stringWithFormat:@"Indoor scan completed: %lu chars", (unsigned long)(indoorScanResult ? indoorScanResult.length : 0)]];
-            completionHandler(beacons, indoorScanResult);
+// Fetches the indoor ML location (nil when indoor positioning is not configured or the
+// RadarSDKIndoors framework is absent), then invokes the completion with the passed-through
+// beacons so callers keep the beacon variable they already resolved. RadarIndoors decides
+// internally whether indoor positioning is active based on trackingOptions/model state.
+- (void)getIndoorLocationIfConfigured:(CLLocation *)location
```
alanjcharles14: we should add a timeout to this like `performIndoorScanIfConfigured` had so we don't get stuck if the completion is never invoked

alanjcharles14: do we still need to pass in the location? I don't think it's ever forwarded to the model

alanjcharles14 (commented):
the non-indoor track path seems behaviorally equivalent, just a couple of smaller notes but otherwise looks good

RadarSDK/RadarAPIClient.m
stevepopovich: this is why I let the AI type for me

RadarSDK/RadarAPIClient.m
stevepopovich: fair, can add 👍🏼

RadarSDK/Radar.m
```diff
@@ -392,20 +395,23 @@ + (void)trackOnceWithDesiredAccuracy:(RadarTrackingOptionsDesiredAccuracy)desire
...
                                          completionHandler:^(RadarStatus status, NSDictionary *_Nullable res, NSArray<RadarEvent *> *_Nullable events, RadarUser *_Nullable user,
                                                              NSArray<RadarGeofence *> *_Nullable nearbyGeofences, RadarConfig *_Nullable config, RadarVerifiedLocationToken *_Nullable token) {
-                                            if (status == RadarStatusSuccess && config != nil) {                                    
-                                                [[RadarLocationManager sharedInstance] updateTrackingFromMeta:config.meta];                                            
+                                            if (status == RadarStatusSuccess && config != nil) {
+                                                [[RadarLocationManager sharedInstance] updateTrackingFromMeta:config.meta];
+                                            }
+                                            if (status == RadarStatusSuccess && user) {
+                                                [[RadarIndoors shared] updateTrackingWithUser:user completionHandler:^{}];
```
nickdonnelly: nit: we're making the same `status == RadarStatusSuccess` check here twice.

RadarSDK/RadarLocationManager.m
stevepopovich: hmmmm yeah I could see that be a good improvement

RadarSDK/RadarIndoors.swift
```diff
@@ -83,20 +87,33 @@ class RadarSDKIndoors {
...
+    let onLocationUpdate: @Sendable @convention(block) (CLLocation) -> Void
+
+    // Initialized entirely from nonisolated, Sendable expressions so the singleton can be built
+    // off the actor. The stored properties are only read again from actor-isolated methods.
+    nonisolated override init() {
+        self.sdk = RadarSDKIndoors()
+        self.onLocationUpdate = { location in
+            Task {
+                await RadarDelegateHolder.didUpdateClientLocation(location: location, stopped: false, source: .indoors)
+                RadarLogger.shared.debug("indoor location update")
```
nickdonnelly: i know this is carryover but can we either enrich this log statement or remove it

RadarSDK/RadarLocationManager.m
```diff
@@ -1047,31 +1048,23 @@ - (void)handleLocation:(CLLocation *)location source:(RadarLocationSource)source
...
+                              beacons:(NSArray<RadarBeacon *> *_Nullable)beacons
+                    completionHandler:(void (^)(NSArray<RadarBeacon *> *_Nullable, CLLocation *_Nullable))completionHandler {
+    [[RadarIndoors shared] getLocationWithCompletionHandler:^(CLLocation *_Nullable indoorLocation) {
+        // RadarIndoors resolves the location on the RadarIndoorsActor's (background) executor, so
+        // its completion fires off the main thread. The downstream track pipeline touches
+        // main-actor-isolated state (e.g. RadarInAppMessageManager), and the real RadarAPIHelper
+        // already delivers its callbacks on the main queue, so marshal back to main here to keep
+        // that contract and avoid a main-actor executor assertion.
+        [RadarUtilsDeprecated runOnMainThread:^{
```
nickdonnelly: `RadarUtilsDeprecated`? 👀

stevepopovich: yeah this is one I need help with 🤔 I wonder what ShiCheng thinks here

RadarSDK/RadarIndoors.swift
stevepopovich: can make it better 👍🏼

RadarSDK/Radar.m
stevepopovich: good find

RadarSDK/RadarLocationManager.m
stevepopovich: Seems like it is deprecated just in favor the Swift imp, so still have to use this

RadarSDK/RadarLocationManager.m
stevepopovich: Good  call on this one!

alanjcharles14 (dismissed):

RadarSDK/RadarLocationManager.m
ShiCheng-Lu: I think this whole function could be removed, update the swift function to run with a timeout. Also why did we add a time out, the indoor SDK should not be waiting for anything, it should return immediately.

RadarSDK/Radar.m
```diff
@@ -336,10 +339,9 @@ + (void)trackOnceWithDesiredAccuracy:(RadarTrackingOptionsDesiredAccuracy)desire
                          };
 
                          void (^performIndoorScanThenTrack)(NSArray<RadarBeacon *> *_Nullable) = ^(NSArray<RadarBeacon *> *_Nullable beacons) {
-                            [[RadarLocationManager sharedInstance] performIndoorScanIfConfigured:location
-                                                                                          beacons:beacons
-                                                                                completionHandler:^(NSArray<RadarBeacon *> *_Nullable beacons, NSString *_Nullable indoorScan) {
-                                callTrackAPI(beacons, indoorScan);
+                            [[RadarLocationManager sharedInstance] getIndoorLocationIfConfiguredWithBeacons:beacons
+                                                                                          completionHandler:^(NSArray<RadarBeacon *> *_Nullable beacons, CLLocation *_Nullable indoorLocation) {
```
ShiCheng-Lu: we don't need to be passing beacons along.

RadarSDK/RadarIndoors.swift
```diff
@@ -83,20 +87,35 @@ class RadarSDKIndoors {
...
-            RadarLogger.shared.debug("indoor location update")
+    let sdk: RadarSDKIndoors?
+
+    let onLocationUpdate: @Sendable @convention(block) (CLLocation) -> Void
+
+    // Initialized entirely from nonisolated, Sendable expressions so the singleton can be built
+    // off the actor. The stored properties are only read again from actor-isolated methods.
+    nonisolated override init() {
+        self.sdk = RadarSDKIndoors()
+        self.onLocationUpdate = { location in
```
ShiCheng-Lu, line 116: I think this still needs to be marked `@Sendable @convention(block)` in order to work when passed through to the Indoors SDK

RadarSDK/RadarIndoors.swift
```diff
@@ -83,20 +87,35 @@ class RadarSDKIndoors {
...
+                await RadarDelegateHolder.didUpdateClientLocation(location: location, stopped: false, source: .indoors)
+                RadarLogger.shared.debug(
+                    "Indoor location update | latitude = \(location.coordinate.latitude); longitude = \(location.coordinate.longitude); horizontalAccuracy = \(location.horizontalAccuracy); floor = \(location.floor.map { String($0.level) } ?? "nil"); timestamp = \(location.timestamp)"
+                )
+            }
         }
+        super.init()
     }
 
     public func updateTracking(user: RadarUser) async {
```
ShiCheng-Lu: let's update this param to be geofences[] instead of user.

ShiCheng-Lu: I think we only access the geofences within user.

RadarSDK/RadarIndoors.swift
ShiCheng-Lu, line 51: this is me looking back thinking I should've done this differently. But maybe we should update the IndoorSDK's interface to all be one param. that way we can call with NSObject.perform(arg1, arg2) in swift (arg2 is the async callback auto generated).

I think we're in a spot where we can still make this update since it's not being used RN.

this way we don't have to go through the swift bridge for .invoke

ShiCheng-Lu: and update it to be the same pattern as RadarSDKFraud (in the main sdk)

Maybe we can discuss this and standardise the interface between our sub sdks.

RadarSDK/RadarLocationManager.m
```diff
@@ -1113,9 +1120,12 @@ - (void)sendLocation:(CLLocation *)location stopped:(BOOL)stopped source:(RadarL
                             [RadarSyncManager rollbackSyncState];
                         }
                     }
-                    
+
                     [self updateTrackingFromMeta:config.meta];
                     [self replaceSyncedGeofences:nearbyGeofences];
+                    if (user) {
+                        [[RadarIndoors shared] updateTrackingWithUser:user completionHandler:^{}];
```
ShiCheng-Lu: RadarIndoors updateTracking should also be ran from user initiated tracks, I don't think those go through sendLocation, but maybe I'm wrong. Let's double check

RadarSDK/RadarIndoors.swift
stevepopovich: Already marked `@Sendable @convention(block) `on onLocationUpdate 👍

RadarSDK/Radar.m
stevepopovich: good call, removed

RadarSDK/RadarIndoors.swift
stevepopovich: I think that makes sense, fixed

RadarSDK/RadarLocationManager.m
stevepopovich: I think it is covered 🤔 double check me

RadarSDK/RadarIndoors.swift
stevepopovich: gave it a shot, matches the pattern now I think

stevepopovich: It broke CI so gonna leave this for now

ShiCheng-Lu: I think we need to make a new release for indoor sdk (before we merge this PR) that has the new interface for combining the args for this function. 

I do think we should change it before we make any release, otherwise it's a back-compat we have to maintain in prod instead of just a broken CI

RadarSDK/RadarLocationManager.m
stevepopovich: still worth it just to be safe IMO

RadarSDK/RadarIndoors.swift
stevepopovich: https://github.com/radarlabs/radar-sdk-ios-indoors/pull/16

RadarSDK/RadarIndoors.swift
```diff
@@ -26,77 +30,111 @@ class RadarSDKIndoors {
...
                 continuation.resume()
             }
-            let selector = NSSelectorFromString("useModelWithName:getModelData:completionHandler:")
-            bridge.invoke(target: instance, selector: selector, args: [model, getModelData, completion])
+            let config: [String: Any] = ["name": model, "getModelData": getModelData]
+            let selector = NSSelectorFromString("useModelWithConfig:completionHandler:")
+            guard instance.responds(to: selector) else {
+                continuation.resume()
+                return
+            }
```
ShiCheng-Lu: you can check the instance responds before the withCheckedContinuation. (unless there's some "passing selector race condition")

also maybe check for all the selectors in the initiation, but tomato tomato.

RadarSDK/RadarLocationManager.m
ShiCheng-Lu: really don't think we should have a timeout because the getLocation function is synchronous. 

I think we can remove this from radar location manager. just call the RadarIndoor function directly.

If you really need a timeout, can add it in RadarIndoors in swift, make a `withTimeout(time, async closure)` helper.

stevepopovich:
@ShiCheng-Lu I think this is good. I would want to merge https://github.com/radarlabs/radar-sdk-ios-indoors/pull/16 too

RadarSDK/RadarAPIClient.m
```diff
@@ -453,7 +450,19 @@ - (void)trackWithLocation:(CLLocation *_Nonnull)location
         }
     }
 
-    if (options.usePressure || options.useMotion) {
+    if (options.useIndoorScan && indoorLocation &&
+        CLLocationCoordinate2DIsValid(indoorLocation.coordinate) &&
+        (indoorLocation.coordinate.latitude != 0.0 || indoorLocation.coordinate.longitude != 0.0)) {
```
ShiCheng-Lu: nit: this could probably be checked in swift (returns valid coord or nil)

actually probably just use the objc `CLLocationCoordinate2DIsValid` function for now

ShiCheng-Lu: and CLLocationCoordinate2DIsValid does != 0.0 check already

RadarSDK/RadarIndoors.swift
```diff
@@ -26,77 +30,111 @@ class RadarSDKIndoors {
         instance = cls.init()
```
ShiCheng-Lu: lets check the "responds to selector" here to match fraud sdk. Also more stricter verificatioin on the object is our object.

RadarSDK/RadarIndoors.swift
```diff
@@ -16,8 +16,12 @@ public actor RadarIndoorsActor {
 
 @RadarIndoorsActor
 class RadarSDKIndoors {
```
ShiCheng-Lu, line 18: I think the RadarSDKIndoors is sendable, so maybe a sendable class works better instead of the actor and nonisolated init and instance. But it's kind of tomatoes tomatoes.

RadarSDK/RadarIndoors.swift
ShiCheng-Lu: bump

ShiCheng-Lu (dismissed):
Couple minor things, some are just my opinion vs others.

* still don't think we need a time out for get location because it's 'synchronous'. If we do, run it in swift.
* checking selector responds to during init.

RadarSDK/RadarLocationManager.m
stevepopovich: removed!

RadarSDK/RadarAPIClient.m
stevepopovich: The guard's already here: CLLocationCoordinate2DIsValid(coord) && (lat != 0.0 || lon != 0.0). Heads up though — CLLocationCoordinate2DIsValid does not cover the != 0.0 case: (0,0) ("null island") passes it as a valid coordinate, so the explicit zero check is deliberate to avoid overwriting with a null-island prediction. Left as-is for now

RadarSDK/RadarIndoors.swift
stevepopovich: leaving for now

RadarSDK/RadarIndoors.swift
stevepopovich: Fixed!

RadarSDK/RadarAPIClient.m
ShiCheng-Lu: Oh, I was thinking of `CLLocation.isValid` an extension function we added in CLLocation+Radar.h

ShiCheng-Lu: lets use that since we're getting the CLLocation back

I think `indoorLocation.isValid` should work

ShiCheng-Lu (dismissed):

RadarSDK/RadarIndoors.swift
```diff
@@ -149,11 +177,28 @@ internal class RadarIndoors: NSObject {
...
+    }
+
+    nonisolated static func validIndoorLocation(_ location: CLLocation?) -> CLLocation? {
+        guard let location else {
+            return nil
+        }
+        let coordinate = location.coordinate
+        guard CLLocationCoordinate2DIsValid(coordinate),
+            coordinate.latitude != 0.0 || coordinate.longitude != 0.0
+        else {
```
ShiCheng-Lu, line 198: I was suggesting using `location.isValid` in objective-C since we have a custom defined variable called `isValid` for the CLLocation class checks for lat / lng is within epsilon of 0. but this works.

ShiCheng-Lu (approved):

PR #644, opened 2026-07-10
FRA 2079 reveal risk API

ShiCheng-Lu:

https://linear.app/radarlabs/issue/FRA-2079/reveal-risk-ios-sdk-interface

## Manual test steps

Screenshot 2026-07-14 at 10 49 23 AM

1.
2.
3.

## Checklist

- [X] New code is written in Swift (the SDK is migrating off Objective-C)
- [ ] Added or updated unit tests (`make test`)
- [X] `make lint-swift` and `make format-check` pass locally
- [ ] For breaking changes: added a `MIGRATION.md` entry and bumped the version with `./set-version.sh`
- [ ] Updated README / public docs if the public API changed
- [ ] No real API keys committed in the example app
- [ ] Only fixed lint/format violations on lines I changed; removed any now-stale `.swiftlint-baseline.json` entries


RadarSDK/RadarAPIClient.swift
```diff
@@ -115,9 +115,45 @@ public final class RadarAPIClient: Sendable {
...
+            "sdkVersion": RadarUtils.sdkVersion,
+            "deviceModel": RadarUtils.deviceModel,
+            "deviceOS": await RadarUtils.deviceOS,
+            "country": RadarUtils.country,
+            "timeZoneOffset": RadarUtils.timeZoneOffset,
+            "lang": RadarSettings.userLanguage,
+            "fraudPayload": fraudPayload,
+            "appId": Bundle.main.bundleIdentifier,
+            "appName": Bundle.main.object(forInfoDictionaryKey: "CFBundleName"),
+            "appVersion": Bundle.main.object(forInfoDictionaryKey: "CFBundleShortVersion"),
```
alanjcharles14: shouldn't this be `CFBundleShortVersionString`

RadarSDK/RadarRevealRiskToken.swift
```diff
@@ -0,0 +1,149 @@
...
+
+@objc(RadarRevealRiskTokenNetworkIpAddress) @objcMembers
+final class RadarRevealRiskTokenNetworkIpAddress: NSObject, Decodable, Sendable {
+    let countryCode: String?
+    let country: String?
+    let countryFlag: String?
+    let state: String?
+    let city: String?
+    let postalCode: String?
+    let lattiude: Double?
```
alanjcharles14: `lattiude` -> `latitude`

RadarSDK/RadarRevealRiskManager.swift
```diff
@@ -0,0 +1,71 @@
...
+    func revealRisk(
+        useSecondaryVerifiedHost: Bool,
+        completionHandler: @escaping @Sendable (RadarStatus, RadarRevealRiskToken?) -> Void
+    ) {
+        Task {
+            do {
+                let token = try await self.revealRisk(useSecondaryVerifiedHost: useSecondaryVerifiedHost)
+                completionHandler(.success, token)
+            } catch {
+                if let radarError = error as? RadarError {
```
alanjcharles14, line 60: the `else if let _ = error as? RadarAPIClient.APIError` branch and the final `else` both return `.errorServer`, making the middle branch dead code right?. Also none of these paths log the error before calling the completion handler, so a failed `revealRisk` doesn't produce a log line.

RadarSDK/RadarSwiftBridge.h
```diff
@@ -35,7 +31,7 @@
 - (RadarTripOptions * _Nullable)getTripOptions;
 - (RadarUser * _Nullable)radarUser;
 - (void)didReceiveEvents:(NSArray<RadarEvent *> * _Nonnull)events user:(RadarUser * _Nonnull)user;
-- (void)didUpdateClientLocation:(CLLocation * _Nonnull)location stopped:(BOOL)stopped source:(RadarLocationSource)source;
+- (void)invokeWithTarget:(NSObject * _Nonnull)target selector:(SEL _Nonnull)selector args:(NSArray * _Nonnull)args;
```
alanjcharles14: is anything calling this? I don't see it anywhere else. It looks like `RadarSDKFraud` is using raw `perform` calls.

RadarSDK/Radar.m
```diff
@@ -1778,4 +1786,7 @@ + (void)didReceivePushNotificationPayload:(NSDictionary *)payload completionHand
     }
 }
 
++ (void)revealRisk {
```
alanjcharles14: leftover? Should we remove?

RadarSDK/Radar.m
```diff
@@ -429,6 +430,16 @@ + (void)trackVerifiedWithBeacons:(BOOL)beacons desiredAccuracy:(RadarTrackingOpt
     [[RadarVerificationManager sharedInstance] trackVerifiedWithBeacons:beacons desiredAccuracy:desiredAccuracy reason:reason transactionId:transactionId completionHandler:completionHandler];
 }
 
++ (void)revealRiskWithCompletionHandler:(RadarRevealRiskCompletionHandler)completionHandler {
```
alanjcharles14, line 433: should we also log the result?

RadarSDK/RadarSDKFraud.swift
```diff
@@ -0,0 +1,54 @@
...
+            return nil
+        }
+        guard let instance = RadarSDKFraudClass.value(forKey: "sharedInstance") as? NSObject else {
+            return nil
+        }
+        return RadarSDKFraud(instance: instance)
+    }()
+    
+    public func initialize(options: [String: Any]) {
+        let selector = NSSelectorFromString("initializeWithOptions:")
```
alanjcharles14: The existing ObjC code in `RadarVerificationManager.m` uses `NSClassFromString` to dynamically load the fraud framework, but then calls methods through `RadarSDKFraudProtocol` so you get dynamic loading with compile-time method signature checking.  This and line ~38 bypass the protocol entirely and uses raw `NSSelectorFromString + perform(_:with:)`, which means zero compile-time safety on method signatures. If the fraud SDK ever renames a selector, this crashes at runtime instead of failing at compile time.
The protocol already exists at `RadarSDKFraudProtocol.h`. We should probably be casting through it instead:
```
swiftguard let fraudClass = NSClassFromString("RadarSDKFraud") as? (NSObject & RadarSDKFraudProtocol).Type else {
    return nil
}
let instance = fraudClass.sharedInstance()
```
If you keep the current approach, at minimum add `instance.responds(to: selector)` guards before each perform call and return `.errorPlugin` / `.errorUnknown` on failure. Right now, a missing selector is an unrecoverable crash, which defeats the purpose of dynamic loading.

alanjcharles14 (changes requested):
I think we should also add unit tests where applicable

RadarSDK/RadarSDKFraud.swift
ShiCheng-Lu: yeah, I'm going to verify the instance responds to all the selectors on initialization of RadarSDKFraud which results in nil if the passed in instance for RadarSDKFraud doesn't match, errorPlugin will be returned if shared = nil.

ShiCheng-Lu: I think currently the exiting initialize with options will crash if sharedInstance is not real. I've added the "sharedInstance" check to those as well.

RadarSDK/RadarSwiftBridge.h
ShiCheng-Lu: thought I got rid of this...

RadarSDK/RadarAPIClient.swift
alanjcharles14: @ShiCheng-Lu don't think I saw anything in your recent commits for this. Thoughts?

RadarSDK/Radar.m
ShiCheng-Lu: logged in reveal manager

RadarSDK/RadarRevealRiskManager.swift
ShiCheng-Lu: updated, we want to handle APIError a little different since we know it has a valid message. we have logs on errors now

Makefile
```diff
@@ -103,7 +103,7 @@ ci-build-analyze:
 
 test-pretty:
 	@set -o pipefail; \
-	  xcodebuild $(XC_TEST_ARGS) test -skip-testing:RadarSDKTests/InAppMessageTest -skip-testing:RadarSDKTests/RadarSettingsTest -skip-testing:RadarSDKTests/RadarNotificationHelperTest 2>&1 \
+	  xcodebuild $(XC_TEST_ARGS) test -skip-testing:RadarSDKTests/InAppMessageTest -skip-testing:RadarSDKTests/RadarSettingsTest -skip-testing:RadarSDKTests/RadarNotificationHelperTest -skip-testing:RadarSDKTests/RadarRevealRiskTests 2>&1 \
```
stevepopovich, line 106: why skip the new tests?

RadarSDK/RadarDelegateHolder.swift
```diff
@@ -5,17 +5,15 @@
 //  Copyright © 2026 Radar Labs, Inc. All rights reserved.
 //
 
-import Foundation
 import CoreLocation
+import Foundation
 
 @globalActor
-@available(iOS 13.0, *)
```
stevepopovich, line 12: Whats up with dropping these everywhere? Did we bump a min version or something?

RadarSDK/RadarRevealRiskManager.swift
```diff
@@ -0,0 +1,73 @@
...
+        )
+        return revealRisk
+    }
+
+    @objc
+    func revealRisk(
+        useSecondaryVerifiedHost: Bool,
+        completionHandler: @escaping @Sendable (RadarStatus, RadarRevealRiskToken?) -> Void
+    ) {
+        Task {
```
stevepopovich, line 54: Do we have to wrap in a `Task`? Maybe just my iOS ignorance here

stevepopovich (commented):
Some questions from a noob

RadarSDK/RadarDelegateHolder.swift
ShiCheng-Lu: yeah, we're already min iOS 13

Makefile
ShiCheng-Lu: this is the XCTest vs SwiftTest, it's skipped in the XCTest section and -only-testing on the SwiftTest section, because the output format is different. so we get readable output in CI.

RadarSDK/RadarRevealRiskManager.swift
ShiCheng-Lu: yeah, because the stuff inside the Task is async, however, this function is technically synchronous, it returns immediately after it's called. The completionHandler is called at a later time when the function actually completes.

stevepopovich: So maybe this is not something to worry about on iOS (it is on Android) but if you just fire off a task without attaching it to anything, could it possibly get randomly cancelled? Though it is probably fine

Makefile
stevepopovich: ooooo oh yeah. Cool 👍🏼

RadarSDK/RadarRevealRiskManager.swift
ShiCheng-Lu: yeah, iOS Tasks are just "run this async" there's no attaching to an owner or anything.

alanjcharles14 (approved):

stevepopovich (approved):

PR #645, opened 2026-07-10
FENCE-2804: Port `locationManager:didUpdateHeading`: to Swift

stevepopovich:
Continues the `RadarLocationManager` ObjC→Swift migration behind the `useSwiftLocationManager` seam (follows #615 / #621 / #631).

## What changed

**1. Port `locationManager:didUpdateHeading:`** — the one **pure leaf** among the remaining `CLLocationManagerDelegate` callbacks. Its only side effect is transforming a `CLHeading` into a dictionary and storing it; the other delegate callbacks route into the large unported `handleLocation:` / completion-handler / `RadarBeaconManager` machinery, so this is the natural next chunk.

- **`RadarLocationManager+Swift.swift`** — new `@objc(didUpdateHeading:)` static twin. It takes the heading rather than the manager, so it uses a plain selector instead of the `...OnLocationManager:` convention.
- **`RadarLocationManager.m`** — `useSwiftLocationManager` guard-and-return shim at the top of the ObjC method; the ObjC body is left intact so both implementations coexist until the port is trusted.
- **`RadarLocationManagerSwift.h`** — declares the twin in the hand-maintained header.
- **`RadarState.swift`** — new `lastHeadingData` computed property backed by `RadarUserDefaults`, mirroring the existing `registeredNotifications` precedent. The Swift `RadarState` shadows the ObjC class in-module, so the twin can't call the ObjC `setLastHeadingData:`.

**2. Route the ObjC `RadarState` heading accessors through `RadarUserDefaults`** — so the storage funnel is consistent across the ObjC/Swift boundary (see below).

- **`RadarUserDefaults.swift`** — now `@objc(RadarUserDefaults)` with an `@objc static var sharedUserDefaults` exposing the app-group-aware backing store.
- **`RadarUserDefaults.h`** — new hand-maintained bridge header (Project visibility, like `RadarLocationManagerSwift.h`), since internal Swift classes aren't in the public `RadarSDK-Swift.h`.
- **`RadarState.m`** — `+lastHeadingData` and `+setLastHeadingData:` now read/write `[RadarUserDefaults sharedUserDefaults]` instead of `standardUserDefaults`.

**Tests** — new `StubCLHeading` (a `CLHeading` subclass mock, same technique as `TrackingCLLocationManager`) and a `RadarLocationManagerSwiftHeadingTests` behavior suite; both registered in the `RadarSDKTests` target.

## Why the RadarUserDefaults routing matters

The Swift twin writes heading through `RadarUserDefaults` (which resolves to the **app-group suite** when a customer configures one via `Radar.initializeWithAppGroup:`). The sole reader — `RadarAPIClient` → `[RadarState lastHeadingData]` — plus the flag-off ObjC writer previously used `standardUserDefaults` directly. So with `useSwiftLocationManager` on **and** an app group configured, the Swift write landed in the shared suite while the ObjC read came from `standard`, silently dropping the heading from the location payload.

Both the getter and setter had to move together: migrating only the reader would have broken the flag-**off** path under an app group. After this change, all four paths — Swift twin write (flag-on), ObjC write (flag-off), the ObjC reader, and the Swift test getter — resolve to the same store in both app-group and non-app-group modes.

## Manual test steps (Example app)

These verify heading is captured and flows into the `/v1/track` payload as `locationMetadata.heading` — the behavior this PR touches. The Example app already configures an app group (`Radar.setAppGroup("group.waypoint.data")` in `Example/Example/AppDelegate.swift`, matching `Example.entitlements`), so it exercises the exact app-group condition this fix addresses.

**Prerequisites**
- A **physical device** — the Simulator has no magnetometer, so `didUpdateHeading:` doesn't fire (the `heading` key comes back null; Simulator is only good for confirming the plumbing, not real values).
- A signing team that can use the `group.waypoint.data` app group (or change the id in both `Example/Example/Example.entitlements` and the `Radar.setAppGroup(...)` call).

**Setup**

1. **Valid publishable key:** there's no key field in the UI, so set `SettingsStore.defaultPublishableKey` (`Example/Example/Services/SettingsStore.swift:44`) to a real test key. The placeholder won't sync to the dashboard.
2. **Force the Swift path on.** This PR's code only runs when `useSwiftLocationManager` is set, which is server-driven. Force it locally by flipping the default at `RadarSdkConfiguration.swift:60`:
   ```swift
   useSwiftLocationManager = dict?["useSwiftLocationManager"] as? Bool ?? true
   ```
   (Or hardcode the ObjC guard in `RadarLocationManager.m:1508` to `if (YES)`.)
4. **Debug logging:** Debug builds default to `.debug`; to be safe, add `Radar.setLogLevel(.debug)` right after `Radar.initialize(...)` in `AppDelegate.swift`.

**Run**
1. Open `Example/Example.xcodeproj`, run the **Example** scheme on the device.
2. Grant **Always** location and **Motion & Fitness** permissions.
3. Ensure `useMotion` is on for local or server tracking options
4. `startTracking()`
5. Move / rotate the device so heading updates fire (`headingFilter` is 5°).

**Verify**
- Open the **Debug** tab → **Logs** filter → expand the `📍 Radar API request | method = POST; url = …/v1/track` entry → confirm `locationMetadata.heading` contains real values (`magneticHeading`, `trueHeading`, `x`, `y`, `z`, `headingAccuracy`, `timestamp`), not null. Use the entry's Copy/Share button to grab the full body.
- Optional: with a valid key, cross-check the location event's heading in the Radar dashboard.

### DIFF
--- a/RadarSDK/RadarState.swift
+++ b/RadarSDK/RadarState.swift
@@ -25,4 +25,16 @@ class RadarState {
             }
         }
     }
+
+    public var lastHeadingData: [String: Double]? {
+        get {
+            guard let dict = RadarUserDefaults.dictionary(forKey: .LastHeadingData) else {
+                return nil
+            }
+            return dict.compactMapValues { ($0 as? NSNumber)?.doubleValue }
+        }
+        set {
+            RadarUserDefaults.set(newValue, forKey: .LastHeadingData)
+        }
+    }
 }
--- a/RadarSDK/RadarState.swift
+++ b/RadarSDK/RadarState.swift
@@ -31,7 +31,7 @@ class RadarState {
             guard let dict = RadarUserDefaults.dictionary(forKey: .LastHeadingData) else {
                 return nil
             }
-            return dict.compactMapValues { ($0 as? NSNumber)?.doubleValue }
+            return dict.compactMapValues { heading in (heading as? NSNumber)?.doubleValue }
         }
         set {
             RadarUserDefaults.set(newValue, forKey: .LastHeadingData)
--- a/RadarSDK.xcodeproj/project.pbxproj
+++ b/RadarSDK.xcodeproj/project.pbxproj
@@ -149,6 +149,7 @@
 		B0A0F0022FBC000100111111 /* RadarLocationManager+Swift.swift in Sources */ = {isa = PBXBuildFile; fileRef = B0A0F0062FBC000100111111 /* RadarLocationManager+Swift.swift */; };
 		B0A0F0032FBC000100111111 /* RadarLocationManagerSwiftSeamTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = B0A0F0072FBC000100111111 /* RadarLocationManagerSwiftSeamTests.swift */; };
 		B0A0F0082FBC000100111111 /* RadarLocationManagerSwift.h in Headers */ = {isa = PBXBuildFile; fileRef = B0A0F0092FBC000100111111 /* RadarLocationManagerSwift.h */; };
+		B0A0F0262FBC000100111111 /* RadarUserDefaults.h in Headers */ = {isa = PBXBuildFile; fileRef = B0A0F0272FBC000100111111 /* RadarUserDefaults.h */; };
 		B0A0F0142FBC000100111111 /* RadarLocationManagerSwiftTestHelpers.swift in Sources */ = {isa = PBXBuildFile; fileRef = B0A0F0152FBC000100111111 /* RadarLocationManagerSwiftTestHelpers.swift */; };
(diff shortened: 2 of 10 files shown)


RadarSDK/RadarState.swift
```diff
@@ -25,4 +25,16 @@ class RadarState {
             }
         }
     }
+
+    public var lastHeadingData: [String: Double]? {
+        get {
+            guard let dict = RadarUserDefaults.dictionary(forKey: .LastHeadingData) else {
+                return nil
+            }
+            return dict.compactMapValues { ($0 as? NSNumber)?.doubleValue }
```
nickdonnelly: can we name the param instead of `$0`

nickdonnelly (dismissed):

alanjcharles14 (dismissed):
One comment for posterity:

We will probably want to either convert `RadarAPIClient` to swift before we roll out location manager or add tests to make sure we're testing the path between write via `RadarLocationManagerSwift.didUpdateHeading` and read via `[RadarState  lastHeadingData]`

Otherwise looks great and is not something we need to address in this PR.

alanjcharles14 (approved):

PR #646, opened 2026-07-13
FENCE-2804: Port `locationManager:didChangeAuthorizationStatus:` to Swift

stevepopovich:
Continues the `RadarLocationManager` ObjC→Swift migration behind the `useSwiftLocationManager` seam.

`didChangeAuthorizationStatus:` is the last `CLLocationManagerDelegate` leaf that doesn't route into the large unported `handleLocation:` / completion-handler / `RadarBeaconManager` machinery.

## `RadarUserDefaults.Key` camelCase rename (lint fix)

Adding the new key surfaced a latent lint problem: every case in `RadarUserDefaults.Key` is PascalCase, which violates SwiftLint's `identifier_name` rule. The existing ~48 cases only pass because they're grandfathered into `.swiftlint-baseline.json`, so any newly added case needs an inline `swiftlint:disable`. Rather than add another disable, this PR fixes the root cause:

- Renamed all 49 PascalCase cases to camelCase. **The rawValue strings are unchanged**, so nothing persisted in `UserDefaults` is affected — this is a pure Swift-identifier rename.
- Updated the 88 call sites across `RadarSettings.swift`, `RadarState.swift`, and the touched test files. Compiler-verified — the old case names no longer exist, so any missed reference is a build error.
- Relabeled `clone(from:to:)` internals to `source`/`target` (external argument labels unchanged) to clear the file's one remaining `identifier_name` violation.
- Removed all 47 `RadarUserDefaults.swift` entries from `.swiftlint-baseline.json`; the file is now lint-clean with no baseline crutch. Other files' baseline entries are untouched.

## Reviewer notes
- **No behavior change on the flag-off path** — the ObjC body is intact.
- **App-group upgrade transient (known, self-healing):** for a user who already had an app group configured on a prior SDK version, the historical auth status sits in `standardUserDefaults` and isn't in the app-group suite the accessors now read. On the first `didChangeAuthorizationStatus` after upgrade, `previousStatus` reads `.notDetermined`, so an unchanged status is seen as a change and may fire one extra `Radar.trackOnce` (and `startTracking` if `startTrackingOnInitialize`). It self-heals after the first write, and firing `trackOnce` on app open when authorized is the intended behavior anyway. This is the same accepted tradeoff as the merged `lastHeadingData` change — not adding out-of-scope migration code.

## Manual testing (Example app)

The flag-on path only does something observable when the server-fetched `sdkConfiguration` has **`useSwiftLocationManager: true`** *and* at least one of **`trackOnceOnAppOpen: true`** / **`startTrackingOnInitialize: true`** — all three default to `false`. It also only fires on an actual authorization *change* (`status != previousStatus`). So a valid run needs those flags set for the Example app's publishable key and a fresh permission grant.

1. In the dashboard/server config for the key the Example app uses, enable `useSwiftLocationManager` and (`trackOnceOnAppOpen` or `startTrackingOnInitialize`). (No in-app toggle exists; the flag is server-driven. For a source build you can instead force it in the debugger / flip the default in `RadarSdkConfiguration.swift`.)
2. Reset the Example app's location permission — delete the app, or iOS **Settings → Example → Location → Never** — so the next grant is a real status change.
3. Launch the Example app. It runs `Radar.initialize` and fetches the config on open. Confirm the flag is live via **Tests tab → gear → "SDK configuration"** disclosure (shows `RadarSdkConfiguration.current()`); `useSwiftLocationManager` should read `true`.
4. Grant location permission — accept the system prompt, or **Tests tab → Permissions → "Request" / "Request Always"**. This drives `CLLocationManager` → `didChangeAuthorizationStatus` into the Swift twin.
5. Open the **Logs (Debug) tab** and confirm the `"Location services authorized"` with the emoji info log fires, followed by a `trackOnce` (surfaced as `didUpdateLocation` / event entries); if `startTrackingOnInitialize` is set, tracking starts. (Ensure log level is `info`/`debug`.)
6. **Flag-off regression:** set `useSwiftLocationManager: false` (or clear the override), reset permission, and repeat — the logs/behavior must be identical, confirming the ObjC path is unchanged.


RadarSDK/RadarLocationManager+Swift.swift
```diff
@@ -277,4 +277,28 @@ final class RadarLocationManagerSwift: NSObject {
...
         ]
     }
+
+    @objc(didChangeAuthorizationStatus:)
+    static func didChangeAuthorizationStatus(_ status: CLAuthorizationStatus) {
+        let state = RadarState()
+        let previousStatus = state.locationAuthorizationStatus
+        state.locationAuthorizationStatus = status
+
+        guard status != previousStatus else {
```
ShiCheng-Lu: don't think we need to be using a guard here, there's no type narrowing or anything happening here. 

I think just an if is more readable

RadarSDK/RadarLocationManager+Swift.swift
```diff
@@ -277,4 +277,28 @@ final class RadarLocationManagerSwift: NSObject {
...
+            return
+        }
+
+        let config = RadarSettings.sdkConfiguration
```
ShiCheng-Lu: this is a good place to guard let config. so you don't have to ? call every time for config access

RadarSDK/RadarLocationManager+Swift.swift
```diff
@@ -277,4 +277,28 @@ final class RadarLocationManagerSwift: NSObject {
             "timestamp": heading.timestamp.timeIntervalSince1970,
         ]
     }
+
+    @objc(didChangeAuthorizationStatus:)
+    static func didChangeAuthorizationStatus(_ status: CLAuthorizationStatus) {
```
ShiCheng-Lu, line 282: I think for these didUpdateXXX delegate functions with shapes defined by Apple, we'd want to eventually switch to using the swift class as the delegate, so maybe it's better to keep the shape the same as what it would be, but I guess also fine as it is.

ShiCheng-Lu: or maybe we'll just end up with a separate swift delegate that calls these functions, so it's fine.

ShiCheng-Lu (dismissed):

alanjcharles14 (dismissed):
solid and clean imo. Can re-approve if you decide to take @ShiCheng-Lu 's style notes and push another commit or two, but the functionality is correct to me.

RadarSDK/RadarLocationManager+Swift.swift
stevepopovich: I think once the conversion is complete, we should re-approach this with the new APIs in mind

alanjcharles14 (approved):

ShiCheng-Lu (approved):

PR #648, opened 2026-07-14
FENCE-2923: CI: run build-and-test on stacked branches, better concurrency, don't run on master

stevepopovich:
## Summary

Reworks the `build-and-test` GitHub Actions workflow so CI runs on **stacked branches**. This was my main impetuous and it is an easy fix. Ended up doing some other fixes. The team already builds stacked branches by hand with plain git but CI never ran on them: the `pull_request` trigger only fired when the PR's *base* was `master`/`develop`, so a PR based on an intermediate feature branch got no CI until it was re-pointed at `master`. `develop` is not a branch that is actually used

Changes:

1. **Trigger on any base branch** — removed the `pull_request: branches: [ master, develop ]` filter so PRs targeting any base (an intermediate stack branch) run CI.
2. **PR-only — dropped the `push` trigger entirely.** Open PRs already cover CI for every branch, and releases build/test on their own (`release-sdk.yml`), so `master` doesn't need a separate post-merge run in this workflow.
3. **Format-check the PR's changed Swift files via `gh pr diff --name-only`** — GitHub computes the diff server-side, so it needs no local git history or merge-base and is correct on stacked PRs by construction (each PR already knows its base). This replaces the previous `git diff origin/<base>...HEAD`, which required a full-depth checkout to resolve the merge-base on a stack. `make format-check` gained a `FORMAT_FILES` override for CI; with it empty (local use) it still falls back to diffing against `FORMAT_BASE`.
4. **Read-only `contents`/`pull-requests` permissions** for the `gh` call. (swift formatting stuff)
5. **PR-scoped concurrency** — cancels superseded runs when a PR branch is re-pushed (restacking a stack).

The `swift-format` check stays scoped to changed files (not whole-repo) because the tree isn't fully `swift-format`-clean and `swift-format` has no baseline mechanism like SwiftLint's. No source or Swift files change — only the workflow and the Makefile target.


alanjcharles14 (approved):
this is cool, nice work

PR #649, opened 2026-07-14
FENCE-2823: Convert `RadarBeaconManager` to swift

alanjcharles14:
## Summary

Migrates `RadarBeaconManager` from Objective-C to Swift, replacing `@synchronized` locking with `@MainActor` isolation and introducing an injectable permissions helper for testability.

## Notable design decisions

- **`lazy var locationManager`** instead of eager init. Avoids `CLLocationManager` requiring a host app bundle in the test target
- **`constraintIdentifierMap`**  The constraint-based ranging API (`startRangingBeacons(satisfying:)`) doesn't carry region identifiers in its callbacks, unlike the deprecated region-based API. A lookup map from constraint key → beacon `_id` is built at range-start and resolved in delegate callbacks to preserve per-beacon identity for early completion detection in `handleBeacons()`
- **Guard reordering in `rangeBeacons`/`rangeBeaconUUIDs`** Empty-input check moved before `addCompletionHandler` (diverges from ObjC). This is intentional: the ObjC ordering has a latent double-invocation bug where an empty-input handler gets called immediately AND queued with a 5s timeout

## Improvements over ObjC

- Nil-checked UUID/major/minor parsing (ObjC passed unchecked values to `CLBeaconRegion`)
- Unconditional return on permissions failure in `rangeBeaconUUIDs` (ObjC had a fall-through bug when completionHandler was nil)
- `nearbyBeaconIdentifiers` insertion moved inside the ranging loop to match ObjC semantics , empty callbacks from CoreLocation don't count toward completion

## Manual test steps

1. Setup beacons in project
2. Set `beacons: true` in RTOs
3. Build and run example app
4. check logs for beacon ranging and beacon event generation
5. safety check: disable bluetooth and see alert in app to enable bluetooth for better performance

## Checklist

## Screenshots / recordings (optional)
image


stevepopovich:
No code to remove here?

alanjcharles14:
@stevepopovich ah yeah... we can delete `RadarBeaconManager.m` just kept it for the review incase folks want to compare. Good callout though

stevepopovich:
I would make the argument it should get deleted in this PR so we don't have dead code hanging around and then I can see the deleted diff in the GitHub PR view, but open to doing it later I suppose.

RadarSDK/RadarBeaconManager+Delegate.swift
```diff
@@ -0,0 +1,253 @@
...
+
+extension RadarBeaconManagerSwift {
+
+    nonisolated func locationManager(
+        _ manager: CLLocationManager,
+        monitoringDidFailFor region: CLRegion?,
+        withError error: Error
+    ) {
+        let identifier = region?.identifier
+        MainActor.assumeIsolated {
```
stevepopovich, line 23: Do you mind explaining why we are using `nonisolated` and then `assumeIsolated`? Probably missing something iOS related

RadarSDK/RadarBeaconManager+Delegate.swift
```diff
@@ -0,0 +1,253 @@
...
+        }
+    }
+
+    nonisolated func locationManager(
+        _ manager: CLLocationManager,
+        didRange beacons: [CLBeacon],
+        satisfying beaconConstraint: CLBeaconIdentityConstraint
+    ) {
+        let constraintUUID = beaconConstraint.uuid.uuidString
+        let constraintMajor = beaconConstraint.major.map { "\($0)" }
```
stevepopovich, line 40: nit: 
```suggestion
        let constraintMajor = beaconConstraint.major[0].toString()
```
something like this may be a bit more readable

RadarSDK/RadarLocationManager.m
```diff
@@ -1382,15 +1382,19 @@ - (void)locationManager:(CLLocationManager *)manager didEnterRegion:(CLRegion *)
     }
 
     if ([region.identifier hasPrefix:kSyncBeaconUUIDIdentifierPrefix]) {
-        [[RadarBeaconManager sharedInstance] handleBeaconUUIDEntryForRegion:(CLBeaconRegion *)region
-                                                          completionHandler:^(RadarStatus status, NSArray<RadarBeacon *> *_Nullable nearbyBeacons) {
-                                                              [self handleLocation:location source:RadarLocationSourceBeaconEnter beacons:nearbyBeacons];
-                                                          }];
+        [RadarUtilsDeprecated runOnMainThread:^{
```
stevepopovich, line 1385: why do these need wrapped and runOnMainThread now?

RadarSDK/RadarPermissionsHelper.swift
```diff
@@ -0,0 +1,29 @@
...
+import CoreLocation
+import Foundation
+
+@MainActor
+protocol RadarPermissionsHelping {
+    func locationAuthorizationStatus() -> CLAuthorizationStatus
+    func isRangingAvailable() -> Bool
+}
+
+@MainActor
```
stevepopovich, line 18: why does this need `@MainActor`? Just learning

RadarSDK/RadarBeaconManagerSwift.h
stevepopovich, line 1: Am i dumb (yes probably)

Why do we need an .h file for this?

stevepopovich (commented):
sorry for all the questions, feel free to tell me to ask AI on them

RadarSDK/RadarBeaconManager+Delegate.swift
nickdonnelly: Couldn't it just be `"\(beaconConstraint.major)"`?

RadarSDK/RadarBeaconManager+Delegate.swift
```diff
@@ -0,0 +1,253 @@
...
+// MARK: - Entry/Exit Handlers
+
+extension RadarBeaconManagerSwift {
+
+    @objc(handleBeaconEntryForRegion:completionHandler:)
+    func handleBeaconEntry(
+        for region: CLBeaconRegion,
+        completionHandler: @escaping RadarBeaconCompletionHandler
+    ) {
+        guard !RadarSettings.useRadarModifiedBeacon else { return }
```
nickdonnelly: I think `if` is clearer than `guard` here because of the negation.
```swift
if RadarSettings.useRadarModifiedBeacon { return }
```

RadarSDK/RadarBeaconManager+Delegate.swift
```diff
@@ -0,0 +1,253 @@
...
+            completionHandler(.success, Array(nearbyBeacons))
+        }
+    }
+
+    @objc(handleBeaconUUIDEntryForRegion:completionHandler:)
+    func handleBeaconUUIDEntry(
+        for region: CLBeaconRegion,
+        completionHandler: @escaping RadarBeaconCompletionHandler
+    ) {
+        guard !RadarSettings.useRadarModifiedBeacon else { return }
```
nickdonnelly: same

RadarSDK/RadarBeaconManager+Delegate.swift
```diff
@@ -0,0 +1,253 @@
...
+                RadarLogger.shared.log(level: .error, message: "Missing or invalid metadata for beacon notification | uuid = \(uuidString)")
+                continue
+            }
+
+            let major = beaconDict["major"] as? String
+            let minor = beaconDict["minor"] as? String
+
+            let constraint: CLBeaconIdentityConstraint
+            if let majorString = major, let majorValue = CLBeaconMajorValue(majorString),
+                let minorString = minor, let minorValue = CLBeaconMinorValue(minorString)
```
nickdonnelly: these major/minor assignments can be moved into the `if let`

```swift
if let major = beaconDict["major"] as? String, let majorValue = CLBeaconMajorValue(major), /* same for minor */ {
	...
}
```

RadarSDK/RadarBeaconManager.swift
```diff
@@ -0,0 +1,271 @@
...
+
+    private override init() {
+        super.init()
+    }
+
+    // MARK: - Region Helpers
+
+    private func region(for beacon: RadarBeacon) -> CLBeaconRegion? {
+        guard let uuid = UUID(uuidString: beacon.uuid) else { return nil }
+        guard let major = CLBeaconMajorValue(beacon.major),
```
nickdonnelly: these guards can be combined

RadarSDK/RadarBeaconManager.swift
```diff
@@ -0,0 +1,271 @@
...
+            beaconIdentityConstraint: constraint,
+            identifier: uuidString
+        )
+    }
+
+    func constraintKey(uuid: String, major: String? = nil, minor: String? = nil) -> String {
+        var key = uuid
+        if let major { key += "-\(major)" }
+        if let minor { key += "-\(minor)" }
+        return key
```
nickdonnelly, line 72: this could produce duplicate keys for `major = x, minor = nil` and `major = nil, minor = x`:

```
x = 5
// case 1 produces
"uuid-5"

// case 2 also produces
"uuid-5"
```

RadarSDK/RadarBeaconManager.swift
```diff
@@ -0,0 +1,271 @@
...
+    private var timeoutTask: Task<Void, Never>?
+
+    // MARK: - Completion Handler Management
+
+    private func addCompletionHandler(_ handler: @escaping RadarBeaconCompletionHandler) {
+        completionHandlers.append(handler)
+
+        if timeoutTask == nil {
+            timeoutTask = Task {
+                try? await Task.sleep(nanoseconds: 5_000_000_000)
```
nickdonnelly, line 86: `Task.sleep(for: .seconds(5))`

RadarSDK/RadarBeaconManager+Delegate.swift
alanjcharles14: `beaconConstraint.major` is an `NSNumber?` and the whole point of `.map { "\($0)" }` is to preserve the optionality.

`"\(beaconConstraint.major)"` would stringify the optional wrapper, giving you `Optional(5)` or `nil`. This matters because `constraintKey` branches on whether major/minor are nil.

`beaconConstraint.major[0].toString()` doesn't apply bc major isn't an array, it's a single optional number.

RadarSDK/RadarBeaconManager+Delegate.swift
alanjcharles14: cleaned it up in a slightly different way so i don't have to repeat the same lookup.

RadarSDK/RadarBeaconManager.swift
alanjcharles14: yeah i guess technically correct but not something we'd ever hit in practice. `CLBeaconIdentityConstraint` only supports three shapes, `uuid`, `uuid+major`, or `uuid+major+minor`. There's no "minor without major" constraint in CoreLocation, and every caller here derives the key from a constraint's .major/.minor. So you can never actually hit the (major: nil, minor: x) case. I added explicit labels so they can't collide though.

RadarSDK/RadarBeaconManager.swift
alanjcharles14: this won't work with our deployment target (iOS 13). `Task.sleep(for:)` takes `Duration` which is only available in iOS 16+. https://developer.apple.com/documentation/swift/task/sleep(for:tolerance:clock:)

RadarSDK/RadarBeaconManager+Delegate.swift
alanjcharles14: So, we have: 
```
@MainActor
@objc(RadarBeaconManagerSwift)
class RadarBeaconManagerSwift: NSObject, CLLocationManagerDelegate {
```

This means every method and stored property is isolated to the main actor by default but the `CLLocationManagerDelegate` protocol is an old-scool objC protocol. Its requirements are **not** main actor isolated (nonisolated). Swift's concurrency checker won't let a `@MainActor` method satisfy a nonisolated protocol requirement bc it can't prove it will be called on the main thread

RadarSDK/RadarLocationManager.m
alanjcharles14: totally related to your previous q:

`RadarLocationManager`'s own `CLLocationManagerDelegate` callbacks (like `didEnterRegion:`) are not guaranteed to run on the main thread. So if you called the beacon manager directly from a background thread, you'd violate the main actor assumption and the` MainActor.assumeIsolated` inside the the code in the q above would trip its precondition and crash. The old objC beacon manager had no isolation requirement, so callers didn't have to care. Now that it's `@MainActor Swift`, the objC callers are responsible for hopping to main themselves

RadarSDK/RadarPermissionsHelper.swift
alanjcharles14: mmm good catch/question. tl;dr for testing lol

Marking the protocol @MainActor propagates the isolation to every conformer (including test mocks), which does two useful things:

- It lets `RadarBeaconManagerSwift` call `permissionsHelper.locationAuthorizationStatus()` synchronously with no await and no `Sendable ` warnings, since both sides share the main actor isolation. A stored `any RadarPermissionsHelping` on a main actor class is much friendlier to the concurrency checker when the protocol is itself main actor.
- It keeps the whole beacon subsystem on one isolation domain, so a mock injected in tests has the same rules.

RadarSDK/RadarBeaconManagerSwift.h
alanjcharles14: no, you're not dumb lol. It's basically how you bridge swift files to objc files that need to call them. 

The beacon manage class is internal, not public and swift auto-generates an objC interface header (`RadarSDK-Swift.h`), but for a framework target that header has two flavors, and the objC files pick one via ` __has_include`:


i.e. RadarLocationManager.m 

```
#if __has_include(<RadarSDK/RadarSDK-Swift.h>)
#import <RadarSDK/RadarSDK-Swift.h>
#elif __has_include("RadarSDK-Swift.h")
#import "RadarSDK-Swift.h"
#endif
```


The angle bracket `<RadarSDK/RadarSDK-Swift.h>` is the public interface. It only contains public @objc declarations. Sooooo in any build configuration that resolves to the public header, the objC code would have no declaration for the class and couldn't call `[[RadarBeaconManagerSwift shared] ...].`

RadarSDK/RadarLocationManager.m
stevepopovich: wow interesting

RadarSDK/RadarBeaconManager+Delegate.swift
stevepopovich: my bad, thank you

RadarSDK/RadarBeaconManagerSwift.h
stevepopovich: heard that 👍🏼

stevepopovich (approved):
thanks for clearing my questions up (I know I have done it myself already but didn't fully grasp all the ins and outs)

alanjcharles14:
@stevepopovich no problem at all!

PR #657, opened 2026-07-22
FENCE-2929: Convert RadarNotificationHelper to Swift

alanjcharles14:
## Summary

Migrates `RadarNotificationHelper.m` to Swift, splitting it into focused single-responsibility files.

## Changes

**New Swift files:**
- `RadarEventNotifications.swift`: campaign content extraction + legacy event notification text
- `RadarNotificationSwizzling.swift`: swizzle setup and deep link handling
- `RadarNotificationUtils.swift`: permissions, notification diff, client-side campaign management

**New ObjC helper:**
- `RadarSwizzleHelper.h/.m`:  houses the swizzled delegate handlers. These *must* stay in ObjC: the `[self swizzled_...]` callback pattern relies on `objc_msgSend` dispatch, which Swift can't reliably reproduce (static dispatch causes infinite recursion; raw IMP casting breaks on the closure-vs-block ABI and async bridge thunks).

**Tests:**
- `RadarEventNotificationsTest.swift`: exhaustive coverage of `extractCampaignContent`, `isCampaign`, and `legacyNotificationText` (all 6 event types + edge cases)
- Removed two obsolete ObjC tests in `RadarSDKTests.m` whose target method no longer exists; equivalent coverage already lives in `RadarNotificationTest.swift`

## Manual test steps
To be honest, there's a lot of manual setup involved here. 
1. setup push notification func in radar dashboard
2. add some logging to various lifecycle delegates
3. create a campaign/geofences in dashboard
4.  run all of these different QA tests 
5. profit

Here's a list of the testing I actually did: 

Unit tests pass. Manually QA'd against the Example app on device:

- ✅ Device token capture + sync
- ✅ Silent push → `didReceivePushNotificationPayload`
- ✅ Deep link handling (`autoHandleNotificationDeepLinks`)
- ✅ Conversion logging on notification tap
- ✅ Host app's own notification delegates still fire (swizzle chain preserved)
- ✅ Campaign geofence notifications

## Checklist


.gitignore
```diff
@@ -18,3 +18,6 @@ Example/Example.xcodeproj/project.xcworkspace/xcshareddata/swiftpm/
 .claude/settings.local.json
 .tools/
 
+
+# AI semantic search index
+.ai-index/
```
stevepopovich, line 23: wah I have never seen this??

RadarSDK/RadarEventNotifications.swift
```diff
@@ -0,0 +1,137 @@
...
+//  RadarSDK
+//
+//  Created by Alan Charles on 7/17/26.
+//  Copyright © 2026 Radar Labs, Inc. All rights reserved.
+//
+
+import Foundation
+import UserNotifications
+
+private let kEventNotificationIdentifierPrefix = "radar_event_notification_"
```
stevepopovich: nit: I think using `k` prefix is not a typical thing in Swift

RadarSDKTests/RadarEventNotificationsTest.swift
stevepopovich, line 1: seems like pretty excellent test coverage!

RadarSDK/RadarEventNotifications.swift
```diff
@@ -0,0 +1,137 @@
...
+        else { return nil }
+
+        return (metadata, text)
+    }
+
+    private static func postNotification(identifier: String, content: UNNotificationContent) {
+        let request = UNNotificationRequest(identifier: identifier, content: content, trigger: nil)
+        UNUserNotificationCenter.current().add(request) { error in
+            if let error {
+                RadarLogger.shared.log(level: .debug, message: "Error adding local notification | identifier = \(identifier); error = \(error)")
```
stevepopovich: should we log level error here?

stevepopovich (commented):
"raw IMP casting breaks on the closure-vs-block ABI and async bridge thunks)." maybe I am a noob, but I do not know what this means 🤔

alanjcharles14:
> Member

yeah it's just a lot of objc/swift jargon to be honest. 

When you swizzle, you sometimes need to call the original implementation that was replaced. in objc we did that with `[self swizzled_...]`, which routes through `objc_msgSend`. The runtime figures out the right function and bridges all the argument types for you. I tried to do this from swift by grabbing the original impl as a raw function pointer (`IMP`) and calling it directly via `unsafeBitCast` to a `@convention(c)` function type. But it broke things in two different ways

1. closure-vs-block ABI: the notification APIs take a completion handler. In swift that's a closure and in objc it's a block. They have different memory layouts at the binary level but`objc_msgSend` bridges between them automatically. A raw function-pointer does not....the block arrived malformed and crashes 
2. Async bridge thunks: the app's delegate method is declared async. Swift compiles an async method into a hidden "thunk" that expects swift concurrency runtime context. Calling that thunk as if it were a plain synchronous C function skips all that setup, so it crashes.

.gitignore
alanjcharles14: lol just a search tool I'm using to try and save some tokens/get better results

RadarSDK/RadarEventNotifications.swift
alanjcharles14: ah good catch

RadarSDKTests/RadarEventNotificationsTest.swift
alanjcharles14: thanks! still want to do some more QA haha

RadarSDK/RadarEventNotifications.swift
alanjcharles14: trueeee- just changed

RadarSDK/RadarNotificationUtils.swift
```diff
@@ -0,0 +1,133 @@
...
+            guard granted else {
+                RadarLogger.shared.log(level: .debug, message: "Notification permissions not granted. Skipping adding notifications.")
+                semaphore.signal()
+                return
+            }
+
+            let center = UNUserNotificationCenter.current()
+            let group = DispatchGroup()
+            let collectQueue = DispatchQueue(label: "com.radar.notificationRegister")
+            nonisolated(unsafe) var added: [NotificationValue] = []
```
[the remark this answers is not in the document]
alanjcharles14: mm this is incorrect.

RadarSDK/RadarNotificationUtils.swift
```diff
@@ -0,0 +1,133 @@
...
+//  Created by Alan Charles on 7/17/26.
+//  Copyright © 2026 Radar Labs, Inc. All rights reserved.
+//
+
+import Foundation
+@preconcurrency import UserNotifications
+
+@objc public final class RadarNotificationUtils: NSObject {
+
+    private static let semaphore = DispatchSemaphore(value: 1)
```
stevepopovich, line 14: Should we use a `Actor` instead here? Less chance of bugs I think

alanjcharles14: I think `Actor` would be conceptually cleaner but I don't think it's the right move in this case. The class is `@objc` (actors aren't @objc-compatible), and the semaphore is serializing a multi-callback chain across `remove` and `add` that spans several `UNUserNotificationCenter` callbacks. `Actor` serializes entry points, but would release exclusivity at each await, so we'd need manual wiring to hold it across the full chain... roughly the same complexity done a different way.

stevepopovich: hmm I was just thinking about trying to move to the most modern stuff, but trust your judgement 👍🏼

alanjcharles14: to be clear I completely agree on principal. Hopefully this is the only feature that still has to rely on objc to some extent, so I'm probably being a bit extra cautious...

stevepopovich (dismissed):

stevepopovich (approved):

PR #659, opened 2026-07-28
FENCE-2947: Fix CLLocationManager-per-call memory leak in RadarUtils auth getters

stevepopovich:
## Summary

Fixes the memory leak a customer reported during their SceneDelegate integration ([FENCE-2947](https://linear.app/radarlabs/issue/FENCE-2947)).

`RadarUtils.locationAuthorization` and `RadarUtils.locationAccuracyAuthorization` each allocated a **new `CLLocationManager()` on every call** (iOS 14+, where the class method is deprecated and an instance is required). Instantiating a `CLLocationManager` purely to read authorization status leaks the underlying CoreLocation client (`CLClientCreateWithBundleIdentifier`) — its XPC connection is never torn down. These getters run on **every config fetch and every track payload build** (`RadarAPIClient`), so the leak grows with request volume. This matches the repeating small mallocs in the customer's memory-profiler traces.

The fix reuses a single cached `CLLocationManager` for these status reads (Apple's recommended pattern). Applied to both:
- `RadarUtils.swift` — the **live** path (`@objc(RadarUtils)` is the Swift class, so `[RadarUtils …]` calls from ObjC dispatch here).
- `RadarUtils.m` — the same pattern on the dead `RadarUtilsDeprecated` copy, fixed for consistency.

> Note on the customer's second screenshot (`+[RadarLocationManager sharedInstance]`): that is a `dispatch_once` singleton allocated once for the app lifetime — a **false positive**, not a growing leak. Leak tooling routinely flags long-lived singletons.

## How to reproduce the leak

**Option A — Xcode Instruments (what they did):**
1. Run any app that integrates the SDK and calls `Radar.trackOnce` / lets tracking run.
2. Profile with the **Allocations** or **Leaks** instrument.
3. Trigger repeated config fetches / track calls (each builds a payload that reads the two auth getters).
4. On the *old* code you see `CLLocationManager` / CoreLocation-client allocations accumulate one per call and never free; after this fix they don't.

**Option B — deterministic footprint measurement (used to produce the numbers below):**
Loop each pattern 20,000 times and diff `phys_footprint` (from `task_info(TASK_VM_INFO)`) before/after:
- OLD: `let m = CLLocationManager(); _ = m.authorizationStatus` inside the loop.
- NEW: allocate one `CLLocationManager` outside the loop; read `.authorizationStatus` inside.

## Proof it was there and is now gone

Measured on iPhone 17 simulator (iOS 26.4), 20,000 reads each:

| Pattern | Process footprint growth | Per call |
|---|---|---|
| **Old** (new `CLLocationManager` per call) | **+17,328 KB (~17.3 MB)** | ~0.9 KB, scales with call count |
| **New** (reuse one instance) | **+2,176 KB (~2.1 MB)** | not proportional to call count (residual is simulator/ARC noise) |

The old pattern's growth is linear in the number of calls (the leak); the new pattern flattens out.

This was done with some temporary scripts and testing, those are outlined below:

<details><summary>Testing details</summary>
<p>

## Appendix: temporary measurement tooling (not committed)

These were used only to produce the before/after numbers, then removed. Reproducible on an iPhone 17 simulator (iOS 26.4).

### 1. Throwaway test

Temporarily added to `RadarSDKTests/RadarUtilsTests.swift` (needs `import CoreLocation` at the top of the file):

```swift
// TEMP measurement — not for merge.
@Test func leakProofMeasurement() {
    // Query the process's phys_footprint — the same "real memory" number
    // Xcode's memory gauge / Instruments report.
    func footprintBytes() -> UInt64 {
        var info = task_vm_info_data_t()
        var count = mach_msg_type_number_t(
            MemoryLayout<task_vm_info_data_t>.size / MemoryLayout<integer_t>.size)
        let kr = withUnsafeMutablePointer(to: &info) { ptr in
            ptr.withMemoryRebound(to: integer_t.self, capacity: Int(count)) { intPtr in
                task_info(mach_task_self_, task_flavor_t(TASK_VM_INFO), intPtr, &count)
            }
        }
        return kr == KERN_SUCCESS ? info.phys_footprint : 0
    }

    let iterations = 20_000
    var sink = 0  // prevents the compiler from optimizing the reads away

    // OLD pattern: allocate a fresh CLLocationManager on every read (the leak).
    let oldBefore = footprintBytes()
    for _ in 0..<iterations {
        let m = CLLocationManager()
        if #available(iOS 14.0, *), m.authorizationStatus == .authorizedAlways { sink += 1 }
    }
    let oldAfter = footprintBytes()

    // NEW pattern: reuse one shared CLLocationManager (the fix).
    let shared = CLLocationManager()
    let newBefore = footprintBytes()
    for _ in 0..<iterations {
        if #available(iOS 14.0, *), shared.authorizationStatus == .authorizedAlways { sink += 1 }
    }
    let newAfter = footprintBytes()

    let oldGrowth = Int64(oldAfter) - Int64(oldBefore)
    let newGrowth = Int64(newAfter) - Int64(newBefore)
    let msg =
        "LEAKPROOF iterations=\(iterations) sink=\(sink) "
        + "OLD growth=\(oldGrowth) bytes (\(oldGrowth / 1024) KB) "
        + "NEW growth=\(newGrowth) bytes (\(newGrowth / 1024) KB)"

    // print()/Issue.record() don't surface reliably through xcodebuild, so write
    // the result into the test's sandbox and fish it off the host afterward.
    let out = FileManager.default.temporaryDirectory.appendingPathComponent("leakproof_result.txt")
    try? msg.write(to: out, atomically: true, encoding: .utf8)
}
```

### 2. Run it and read the result

`xcodebuild` clones the target device and discards it after the run, and `print` output doesn't surface — so boot a specific simulator, run the suite *on that exact device*, then read the file out of its data container:

```bash
# Boot a specific simulator and keep it alive (avoids xcodebuild's throwaway clone).
UDID=$(xcrun simctl list devices available \
  | grep -A30 "iOS 26.4" | grep "iPhone 17 (" | head -1 \
  | grep -oE "[0-9A-F-]{36}")
xcrun simctl boot "$UDID"

# Clear any stale result, then run the whole RadarUtilsTests suite on that device.
find ~/Library/Developer/CoreSimulator/Devices/"$UDID" -name leakproof_result.txt -delete 2>/dev/null
xcodebuild test \
  -project RadarSDK.xcodeproj -scheme RadarSDK \
  -destination "platform=iOS Simulator,id=$UDID" \
  -only-testing:RadarSDKTests/RadarUtilsTests

# Read the numbers the test wrote into its sandbox.
find ~/Library/Developer/CoreSimulator/Devices/"$UDID" -name leakproof_result.txt -exec cat {} \;

# Cleanup.
find ~/Library/Developer/CoreSimulator/Devices/"$UDID" -name leakproof_result.txt -delete 2>/dev/null
xcrun simctl shutdown "$UDID"
```

### 3. Output

```
LEAKPROOF iterations=20000 sink=0 OLD growth=17743872 bytes (17328 KB) NEW growth=2228248 bytes (2176 KB)
```

| Pattern | Footprint growth over 20,000 reads |
|---|---|
| Old (new `CLLocationManager` per call) | +17,328 KB (~17.3 MB) — ~0.9 KB/call, scales with call count |
| New (reuse one instance) | +2,176 KB (~2.1 MB) — not proportional to call count |

**Caveats:** measures process `phys_footprint` (a proxy), not an Instruments `leaks` graph; simulator, not device; the ~2 MB residual on the new pattern is attributed to ARC/simulator noise + CoreLocation's one-time client spin-up, since the new loop allocates nothing per iteration.

</p>
</details> 

## Severity / rollout note for the customer

Impact is low: a slow, small per-request drip of a CoreLocation client handle, not a large buffer — unlikely to have real customer-facing impact. It does require an SDK release + the customer bumping their SDK version; there is no app-side workaround since the leak is entirely inside SDK code.


nickdonnelly (approved):
LGTM!

alanjcharles14 (approved):

PR #663, opened 2026-07-31
FENCE-2804: Port shutDown and requestLocation to Swift

stevepopovich:
## Summary

- Continues the `RadarLocationManager` ObjC→Swift seam migration (behind `useSwiftLocationManager`) with two more leaf methods: `shutDown` and `requestLocation`.
- Both only touch `self.locationManager`/`self.lowPowerLocationManager` (already public properties), so they're unblocked by the `RadarState` Swift/ObjC name-collision that blocks most of the remaining backlog (`handleLocation:...`, the region/visit delegate handlers, etc.).
- Adds the matching hand-maintained declarations to `RadarLocationManagerSwift.h` and the `useSwiftLocationManager` dispatch guard in `RadarLocationManager.m`.

## Manual test steps

1. Temporarily default `useSwiftLocationManager` to `true` in `RadarSdkConfiguration.swift`.
2. Run the Example app, start/stop tracking, and confirm the 🦅-prefixed "Shutting down" / "Requesting location" debug logs appear. (Delete the example app to get the permission prompt to fire)
3. Revert the hardcoded flag and confirm the logs are absent (ObjC path).


RadarSDK/RadarLocationManager+Swift.swift
```diff
@@ -278,6 +278,21 @@ final class RadarLocationManagerSwift: NSObject {
         ]
     }
 
+    @objc(shutDownOnLocationManager:lowPowerLocationManager:)
+    static func shutDown(locationManager: CLLocationManager, lowPowerLocationManager: CLLocationManager) {
+        RadarLogger.shared.debug("🦅 Shutting down")
```
nickdonnelly, line 283: Shutting down _what_?

nickdonnelly (approved):

ShiCheng-Lu (approved):

RadarSDK/RadarLocationManager+Swift.swift
alanjcharles14: agreed we should make this more clear but otherwise it looks good to me

stevepopovich: yeah shutting down the location manager but yeah, I will consider improving these more as I go

PR #668, opened 2026-08-07
FENCE-2804: Port desiredAccuracy, didFailWithError, and effective-location fallback to Swift

stevepopovich:
## Summary
- Continues the `RadarLocationManager` ObjC → Swift seam migration, three small unblocked pieces at a time, behind `useSwiftLocationManager` (default off).
- Dedupes the `RadarTrackingOptionsDesiredAccuracy` → `CLLocationAccuracy` switch (previously duplicated in `getLocationWithDesiredAccuracy:` and `updateTracking:fromInitialize:`) into `RadarLocationManagerSwift.clLocationAccuracy(for:)`.
- Ports `locationManager:didFailWithError:`'s log + delegate-notify into `RadarLocationManagerSwift.didFail(error:)`, routed through the existing `RadarSwiftBridgeProtocol.didFail(status:)` hook. The completion-handler timeout flush stays in ObjC since it touches host state (`completionHandlers`) not yet exposed to Swift.
- Dedupes the "effective location" fallback (`manager.location.isValid ? manager.location : [RadarState lastLocation]`), duplicated verbatim across `didEnterRegion:`, `didExitRegion:`, and `didDetermineState:forRegion:`, into `RadarLocationManagerSwift.effectiveLocation(for:)`.
- All three follow the same flag-gated twin pattern as prior ports (`shutDown`, `requestLocation`, `didUpdateHeading`, etc.) — the ObjC body stays as a fallback until the Swift port is trusted.
- Also moved `RadarSDKTests/MockRadarSwiftBridge.swift` out of `RadarSDKTests/RadarReplayBufferTests.swift` because it is used in multiple files

## Manual test plan
`useSwiftLocationManager` is only ever set via remote SDK config today (`RadarSdkConfiguration.swift:56`), so exercising the Swift paths locally needs a temporary local override — **do not ship the override**. In a debug build, either flip the `?? false` default at `RadarSdkConfiguration.swift:56` to `?? true`, or set `RadarSettings.sdkConfiguration = RadarSdkConfiguration(dict: ["useSwiftLocationManager": true])` early in the Example app's launch path, then run the Example app in the Simulator. Revert the override afterward.

For each check below, run once with the flag forced **on** and once **off** (default) and confirm identical behavior:

- [ ] **Accuracy switch** — call `Radar.getLocation(desiredAccuracy:completionHandler:)` (or `trackOnce`) at `.high`, `.medium`, and `.low`; confirm the location callback still fires and `CLLocationManager.desiredAccuracy` is set to `kCLLocationAccuracyBest` / `kCLLocationAccuracyHundredMeters` / `kCLLocationAccuracyKilometer` respectively.
- [ ] **didFailWithError** — Revoke location permission in the example app, then call `Radar.getLocation` (`trackOnce`) and confirm the Example app's event log shows a `RadarStatusErrorLocation` delegate callback.
- [ ] **effectiveLocation fallback** (with simulator) — start tracking with `.presetResponsive` or `.presetEfficient` (both have `showBlueBar = false`, so `self.locationManager` never gets `startUpdatingLocation`/`requestLocation` and its `.location` stays `nil` for the whole run — no timing race to hit) with a synced geofence or beacon configured. Wait for one low-power update to land (this populates `RadarState.lastLocation` via `handleLocation:`), then use Simulator → Features → Location → Custom Location to move into/out of the geofence/beacon region. Confirm the enter/exit event still fires with a non-nil location — since `self.locationManager.location` was never populated, it can only have come from the `RadarState.lastLocation` fallback. (ended up testing this with some local code changes and break points, hard to reproduce)
- [ ] Confirm no crashes or unexpected log spam in either flag state.


nickdonnelly (approved):
```
                    ___                       ___     
                   /  /\          ___        /__/\    
                  /  /:/_        /  /\      |  |::\   
  ___     ___    /  /:/ /\      /  /:/      |  |:|:\  
 /__/\   /  /\  /  /:/_/::\    /  /:/     __|__|:|\:\ 
 \  \:\ /  /:/ /__/:/__\/\:\  /  /::\    /__/::::| \:\
  \  \:\  /:/  \  \:\ /~~/:/ /__/:/\:\   \  \:\~~\__\/
   \  \:\/:/    \  \:\  /:/  \__\/  \:\   \  \:\      
    \  \::/      \  \:\/:/        \  \:\   \  \:\     
     \__\/        \  \::/          \__\/    \  \:\    
                   \__\/                     \__\/    
```

PR #679, opened 2026-08-13
FENCE-2804: Port location visit delegate to Swift

stevepopovich:
## Summary

- Port `CLLocationManagerDelegate.locationManager(_:didVisit:)` to the Swift location-manager twin.
- Preserve the existing missing-location and tracking gates, visit logging, and arrival/departure source mapping.
- Add unit coverage for ignored, arrival, and departure visit paths.

## Manual test instructions

Core Location visit events are not deterministic: iOS infers them from real-world dwell and movement, and Apple does not publish a timing or distance threshold. `startMonitoringVisits()` enables visit delivery independently of Radar's tracking state.

1. On a physical iPhone, configure the SDK with `useVisits: true` so it calls `startMonitoringVisits()`, and enable `useSwiftLocationManager`.
2. Grant the app location permission. Use **Always** authorization and the app's `location` background mode when validating background or relaunch delivery.
3. Move to a distinct destination, remain mostly stationary for 20–30 minutes or longer, then wait for delayed Core Location delivery. These are testing heuristics, not Apple guarantees.
4. With Radar tracking enabled, confirm an arrival-only visit logs:
   - `🦅 Visit detected | arrival = …; departure = 4001-01-01 …`
   - `Handling location | source = VISIT_ARRIVAL; location = …`
5. Leave the destination and travel roughly 1 km or more. When Core Location eventually reports the completed visit, confirm:
   - `🦅 Visit detected | arrival = …; departure = <actual date> …`
   - `Handling location | source = VISIT_DEPARTURE; location = …`
6. Disable Radar tracking without stopping visit monitoring, then reproduce a visit callback. Confirm Core Location’s callback reaches the SDK via `🦅 Visit detected | …`, followed by `🦅 Ignoring visit: not tracking`; there must be no corresponding `Handling location | source = VISIT_…` log.

For repeatable coverage, rely on the unit tests; use the physical-device flow as an end-to-end smoke test.


stevepopovich:
I was able to get these to fire:

<details><summary>Details</summary>
<p>

| | | |
|--------|--------|--------|
| IMG_0002 | IMG_0003 | IMG_0004 | 

</p>
</details> 

Here's how:
- Not tracking or running track. All the correct auth and permissions
- Took a bike to a coffee shop .75 miles away (the way the crow flies)
- Waited about 5 minutes
- Got a arrival
- Left and went back on a bike
- When I parked the bike, I got the departure event

alanjcharles14 (approved):

PR #681, opened 2026-08-13
FENCE-2804: Port start tracking entry point to Swift

stevepopovich:
## Summary

- Port `RadarLocationManager.startTrackingWithOptions:` to its Swift twin when `useSwiftLocationManager` is enabled, while retaining the Objective-C fallback.
- Route authorization-status lookup and tracking updates through `RadarSwiftBridgeProtocol`.
- Isolate Swift tracking seam tests with scoped bridge and permissions mocks, and run the Swift-only paths on the main actor.
- Skip geofence notification registration only in Xcode's hostless XCTest agent; both Objective-C and Swift geofence-replacement paths continue to run.

## Manual test instructions

1. In the Example app configuration, set `useSwiftLocationManager: true`.
2. Grant the app **When In Use** or **Always** location permission, start responsive tracking, and confirm location updates continue normally.
3. Create a temporary geofence around the simulator's current location in the Radar dashboard, relaunch the Example app, and confirm its geofence behavior still works.
4. Reset or deny location permission, start tracking again, and confirm the SDK reports a permissions error without starting tracking.


RadarSDK/RadarLocationManager+Swift.swift
```diff
@@ -31,6 +31,19 @@ final class RadarLocationManagerSwift: NSObject {
     private static let syncBeaconIdentifierPrefix = "radar_beacon_"
     private static let syncBeaconUUIDIdentifierPrefix = "radar_uuid_"
 
+    @objc(startTrackingWithOptions:)
+    static func startTracking(options: RadarTrackingOptions) {
+        let authorizationStatus = RadarSwift.bridge?.locationAuthorizationStatus()
```
ShiCheng-Lu: should be using RadarPermissionsHelperSwift's locationAuthorizationStatus(), create a private var in here that holds a RadarPermissionsHelperSwift() instance, and call locationAuthorizationStatus() on it.

stevepopovich: Fixed, good call 👍🏼

RadarSDKTests/RadarSDKTests.m
```diff
@@ -283,10 +283,6 @@ - (void)assertRoutesOk:(RadarRoutes *)routes {
     AssertRouteOk(routes.car);
 }
 
-+ (void)setUp {
```
stevepopovich: `RadarSDKTests` asserted that the SDK had not been initialized before its test suite began. XCTest does not guarantee suite execution order, so an earlier suite could initialize Radar and cause this assertion to fail despite all individual tests passing. `setUp` run order is only guaranteed in the scope of this file.

Removing the assertion makes the test suite independent of global execution order while preserving each test’s existing initialization setup.

RadarSDK/RadarLocationManager+Swift.swift
```diff
@@ -30,6 +30,21 @@ final class RadarLocationManagerSwift: NSObject {
     private static let syncGeofenceIdentifierPrefix = "radar_geofence_"
     private static let syncBeaconIdentifierPrefix = "radar_beacon_"
     private static let syncBeaconUUIDIdentifierPrefix = "radar_uuid_"
+    @MainActor static var permissionsHelper: RadarPermissionsHelping = RadarPermissionsHelperSwift()
+
+    @MainActor
+    @objc(startTrackingWithOptions:)
+    static func startTracking(options: RadarTrackingOptions) {
```
ShiCheng-Lu, line 36: @alanjcharles14 why are the Permission Helper in swift marked as MainActor even though the functions do not require MainActor?

ShiCheng-Lu: this requires the `startTrackingWithOptions:` to be main thread call from objC, which I don't think is always the case. (might crash and burn)

I don't think the RadarPermissionHelperSwift needs to be @MainActor.

stevepopovich: yeah I am still not sure I quite understand this. I assume the main thread was scoped to the thread of the sdk? idk. We should probably not have this scoped to the main thread I guess

stevepopovich: oh is it because permissions may pop a dialog to ask for location permissions?

ShiCheng-Lu: there's only one MainThread for the device, and someone could call startTracking from a background thread, and I think when it gets to swift it might crash because when you have @MainActor marked, it'll assert this function is running on the main actor (or maybe not). But the PermissionHelperSwift does not need to be @MainActor so we shouldn't put it on @MainActor.

stevepopovich: yeah lemme rework this

alanjcharles14: @ShiCheng-Lu great catch. Struggling to remember exactly why i did it in the first place but iirc it was something dumb related to testing that has now leaked out the obj-c facing method. I just pushed what I think should be a correct fix. @stevepopovich also interested in what you think about my commit. Sorry to send you down the wrong path with my bad code.

stevepopovich: me likey!

alanjcharles14 (approved):
lgtm but i think you might need another approval since I made a commit

RadarSDK/RadarLocationManager+Swift.swift
```diff
@@ -30,6 +30,20 @@ final class RadarLocationManagerSwift: NSObject {
     private static let syncGeofenceIdentifierPrefix = "radar_geofence_"
     private static let syncBeaconIdentifierPrefix = "radar_beacon_"
     private static let syncBeaconUUIDIdentifierPrefix = "radar_uuid_"
+    nonisolated(unsafe) static var permissionsHelper: RadarPermissionsHelping = RadarPermissionsHelperSwift()
+
+    @objc(startTrackingWithOptions:)
+    static func startTracking(options: RadarTrackingOptions) {
+        let authorizationStatus = permissionsHelper.locationAuthorizationStatus()
+        guard authorizationStatus == .authorizedWhenInUse || authorizationStatus == .authorizedAlways else {
```
nickdonnelly, line 38: maybe out of scope for this pr but this `authorizationStatus is in use or always` is a common enough condition that we should probably have a helper for it.

nickdonnelly (approved):

RadarSDK/RadarLocationManager+Swift.swift
stevepopovich: probably fair 👍🏼

PR #684, opened 2026-08-18
added jwe encryption for fraud responses

cameron-radar:
see https://github.com/radarlabs/server/pull/8355

## Summary

Adds support for JWE-encrypted (protected mode) response tokens for verified tracking and reveal risk, controlled by a server-side project setting. When a project has protected responses enabled, the server omits `user`, `events`, and other verdict-bearing fields from the plaintext response and returns an encrypted JWE instead of a signed JWS — the token is opaque to the SDK and the customer decrypts it server-side with their private key, then verifies it against Radar's JWKS.

FRA-1913

## Changes

New public API:

- `RadarTokenFormat` enum (`jws` / `jwe`) exposed on both `RadarVerifiedLocationToken` and `RadarRevealRiskToken` via a new `format` property. Parsed from the response's `format` field, case-insensitively, defaulting to JWS so existing projects are unaffected.

`RadarVerifiedLocationToken`:

- `user` and `events` are now nullable in the initializer, since in JWE mode they only exist inside the encrypted token. Parsing still requires `token` and `expiresAt` in both modes, and still requires `user` + `events` in JWS mode.

`RadarAPIClient`:

- A verified track response with a JWE token short-circuits before user/event parsing: it fires `didUpdateToken:` and completes with success, the token, and nearby geofences, but `nil` user and events.

`RadarVerificationManager`:

- `isLastTokenValid` always returns `NO` for JWE tokens. The reuse check depends on `user.state.distanceToBorder`, which is inside the encrypted payload, so JWE tokens are never reused and every retrieval re-tracks.

`RadarRevealRiskToken`:

- `network`, `device`, and `risk.reasons` are now optional — `nil` in JWE mode for the same reason.

Tests:

- New `RadarVerifiedLocationTokenTests.swift`: JWS parsing, JWE parsing without user/events, default-to-JWS when `format` is missing, case-insensitive format parsing, and required-field validation in both modes.
- `RadarRevealRiskTests.swift`: new test for a protected response that omits reasons/network/device, plus `format` assertions on existing tests.

## Manual test steps

see server pr

## Checklist

- [x] New code is written in Swift (the SDK is migrating off Objective-C) — new tests are Swift; model/manager changes edit existing ObjC files
- [ ] `make lint-swift` and `make format-check` pass locally
- [ ] For breaking changes: added a `MIGRATION.md` entry and bumped the version with `./set-version.sh`
- [ ] Updated README / public docs if the public API changed

### DIFF
--- a/RadarSDK/RadarAPIClient.m
+++ b/RadarSDK/RadarAPIClient.m
@@ -701,6 +701,14 @@
             
                             [RadarOfflineEventManager reset];
 
+                            // nearbyBeaconRegions is not verdict-bearing, so protected (JWE) responses still
+                            // include it in plaintext — register before branching on the response format
+                            id nearbyBeaconRegionsObj = res[@"nearbyBeaconRegions"];
+                            if (nearbyBeaconRegionsObj && [nearbyBeaconRegionsObj isKindOfClass:[NSArray class]]) {
+                                NSArray<NSDictionary<NSString *, NSString *> *> *beaconRegions = (NSArray<NSDictionary<NSString *, NSString *> *> *)nearbyBeaconRegionsObj;
+                                [[RadarBeaconManagerSwift shared] registerBeaconRegionNotificationsFromArray:beaconRegions];
+                            }
+
                             if (events && user) {
                                 [RadarSettings setId:user._id];
                                 [RadarState setRadarUser:user];
@@ -730,12 +738,6 @@
                                     [[RadarDelegateHolder sharedInstance] didUpdateToken:token];
                                 }
 
-                                id nearbyBeaconRegionsObj = res[@"nearbyBeaconRegions"];
-                                if (nearbyBeaconRegionsObj && [nearbyBeaconRegionsObj isKindOfClass:[NSArray class]]) {
-                                    NSArray<NSDictionary<NSString *, NSString *> *> *beaconRegions = (NSArray<NSDictionary<NSString *, NSString *> *> *)nearbyBeaconRegionsObj;
-                                    [[RadarBeaconManagerSwift shared] registerBeaconRegionNotificationsFromArray:beaconRegions];
-                                }
-                                
                                 return completionHandler(RadarStatusSuccess, res, events, user, nearbyGeofences, config, token);
                             } else if (token && token.format == RadarTokenFormatJWE) {
                                 // protected mode: user and events are omitted from the plaintext response
--- a/RadarSDK/RadarAPIClient.m
+++ b/RadarSDK/RadarAPIClient.m
@@ -737,6 +737,12 @@
                                 }
                                 
                                 return completionHandler(RadarStatusSuccess, res, events, user, nearbyGeofences, config, token);
+                            } else if (token && token.format == RadarTokenFormatJWE) {
+                                // protected mode: user and events are omitted from the plaintext response
+                                // and only exist inside the encrypted token
+                                [[RadarDelegateHolder sharedInstance] didUpdateToken:token];
+
+                                return completionHandler(RadarStatusSuccess, res, nil, nil, nearbyGeofences, config, token);
                             } else {
                                 [[RadarLogger sharedInstance] logWithLevel:RadarLogLevelInfo message:[NSString stringWithFormat:@"Setting %lu notifications remaining", (unsigned long)notificationsRemaining.count]];
                                 [RadarState setRegisteredNotifications:notificationsRemaining];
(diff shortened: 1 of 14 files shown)


RadarSDK/RadarRevealRiskToken.swift
```diff
@@ -22,8 +22,16 @@ final class RadarRevealRiskToken: NSObject, Decodable, @unchecked Sendable {
...
+    /// `nil` when `format` is JWE, since the payload only exists inside the encrypted token.
+    let network: RadarRevealRiskTokenNetwork?
+    /// `nil` when `format` is JWE, since the payload only exists inside the encrypted token.
+    let device: RadarRevealRiskTokenDevice?
+
+    private let rawFormat: String?
+    /// The wire format of the token, controlled by a server-side project setting. When JWE, the
+    /// token is opaque to the SDK: decrypt with your private key and verify against Radar's JWKS
+    /// server-side.
+    @objc(format) var format: RadarTokenFormat { rawFormat?.uppercased() == "JWE" ? .jwe : .jws }
```
alanjcharles14, line 34: This swift impl makes `reasons`, `network`, and `device` optional for JWE and adds `format`, but the public `RadarRevealRiskToken.h` still marks those fields `nonnull` and does not expose `format`. We need to update the public header so customers receive the same API the implementation provides, and add a public API test to prevent them from drifting again.

RadarSDK/RadarAPIClient.m
```diff
@@ -744,6 +737,12 @@ - (void)makeTrackRequestWithParams:(NSDictionary *)params
                                 }
                                 
                                 return completionHandler(RadarStatusSuccess, res, events, user, nearbyGeofences, config, token);
+                            } else if (token && token.format == RadarTokenFormatJWE) {
+                                // protected mode: user and events are omitted from the plaintext response
+                                // and only exist inside the encrypted token
+                                [[RadarDelegateHolder sharedInstance] didUpdateToken:token];
+
+                                return completionHandler(RadarStatusSuccess, res, nil, nil, nearbyGeofences, config, token);
```
alanjcharles14, line 747: JWE responses can still include `nearbyBeaconRegions`, but this early return skips registering them. We should move beacon-region registration before the JWS/JWE branches and add a JWE test that includes nearby beacon regions.

RadarSDK/RadarRevealRiskToken.swift
```diff
@@ -111,7 +120,8 @@ enum RadarRevealRiskLevel: Int, Sendable, Decodable {
 @objc(RadarRevealRiskTokenRisk) @objcMembers
 final class RadarRevealRiskTokenRisk: NSObject, Decodable, Sendable {
     let level: RadarRevealRiskLevel
-    let reasons: [String]
+    /// `nil` when the token format is JWE, since reasons only exist inside the encrypted token.
+    let reasons: [String]?
```
alanjcharles14, line 132: These fields can be missing for JWE, but making them optional also allows incomplete JWS responses to decode successfully. Should we validate that they remain required when format is JWS?

RadarSDK/RadarVerifiedLocationToken.m
```diff
@@ -96,8 +98,16 @@ - (instancetype _Nullable)initWithObject:(id _Nonnull)object {
...
-    if (user && events && token && expiresAt) {
-        return [[RadarVerifiedLocationToken alloc] initWithUser:user events:events token:token expiresAt:expiresAt expiresIn:expiresIn passed:passed failureReasons:failureReasons _id:_id fullDict:dict];
+    RadarTokenFormat format = RadarTokenFormatJWS;
+    id formatObj = dict[@"format"];
+    if (formatObj && [formatObj isKindOfClass:[NSString class]] && [[(NSString *)formatObj uppercaseString] isEqualToString:@"JWE"]) {
+        format = RadarTokenFormatJWE;
+    }
+    
+    // in JWE (protected) mode the user and events only exist inside the encrypted token
+    BOOL hasRequiredFields = token && expiresAt && (format == RadarTokenFormatJWE || (user && events));
```
alanjcharles14, line 115: Protected responses omit `failureReasons`, but the token initializer defaults them to an empty array. That reports 'no failures' when the reasons are actually encrypted and unavailable, including for passed == NO. Should we preserve nil for JWE and add a failed-JWE test?

alanjcharles14 (changes requested):

stevepopovich (approved):
looks great, good comments and docs too!

PR #688, opened 2026-08-24
FENCE-2991: Improve iOS app kill detection for trip issues

stevepopovich:
## Summary

This PR improves app-killed detection for active trips and adds searchable Example logs.

Lifecycle detection: 
- Adds an SDK-owned durable lifecycle marker at `radar-appLifecycleMarker`.
- Records the main app process before initialization and emits the existing `App terminating` message on the next main-app launch when the prior process ended unexpectedly while trip options were persisted.
- Keeps the marker set through backgrounding and does not clear it from `applicationWillTerminate` or `Radar.logTermination`, because iOS does not provide a reliable clean-process-exit callback for background-capable apps.
- Excludes app extensions from lifecycle tracking. Extensions share app-group defaults but run as separate processes; allowing them to update the marker creates false or duplicate `app_killed` issues.
- Makes lifecycle initialization idempotent within one process.

Log deduplication:
- Prevents duplicate log uploads by making log flushes single-flight, removing the redundant initialization flush, and preserving logs created during an in-flight upload.
- Deduplicates identical persisted and in-memory logs using their timestamp and full log identity before upload. This addresses replayed persistent logs with identical `createdAt` values.

Debugging improvements:
- Adds live text search to the Example Logs view across summaries and details while preserving kind filters, count, copy, share, expansion, and clear behavior.

The server's existing `App terminating` message contract is unchanged. 

## What we discovered

- A backgrounded iOS app can be removed without a termination callback, so a durable marker is required to detect the next unclean launch.
- The Example app embeds a Location Push extension that initializes Radar in a separate process. Both processes share app-group defaults, which caused extension initialization to look like an app termination. (calls Radar.initalize more than once)
- The log buffer can replay persistent records after a killed process. The same termination record and other messages can therefore appear more than once before upload.
- Initialization had two independent flush triggers. Because flushing was asynchronous, both could submit the same batch before the buffer was cleared.

## Test Plan

1. Configure the Example app with a valid publishable key and enable trip issue detection and SDK logging for the project.
2. Install the Example app with fresh app data, start a trip, background the app, swipe it away from the app switcher, relaunch it, and wait for initialization and log flushing to finish.
3. Refresh the dashboard Trips view, filter by **App killed**, and confirm the trip has exactly one `app_killed` issue.
4. Start another trip, background and foreground the app without swiping it away, and confirm no inferred `app_killed` issue is added.
5. Open the Example app's Debug tab, trigger actions that produce different log kinds, and confirm All, Actions, Events, Locations, and Logs still show their documented kinds.
6. Enter a mixed-case summary term in `Search logs`, clear it, enter only spaces or newlines, and confirm the selected kind filter's full list returns.
7. Expand an entry, search for a term present only in its detail, and confirm it remains visible. Search for a term absent from both fields and confirm **No matches** appears without a crash.
8. Use Copy and Share after searching and confirm the exported content contains only visible matches. Tap Clear and confirm the complete timeline is cleared and **Nothing yet** returns.

### DIFF
--- a/RadarSDK/Radar.m
+++ b/RadarSDK/Radar.m
@@ -45,8 +45,6 @@
 + (instancetype)sharedMarker;
 - (BOOL)beginProcess;
 + (NSString *)appTerminatingMessage;
-- (void)markBackground;
-- (void)markCleanTermination;
 
 @end
 
@@ -140,11 +138,6 @@ BOOL _initialized = NO;
     [[NSNotificationCenter defaultCenter] addObserver:[self sharedInstance]
                                              selector:@selector(applicationWillEnterForeground)
                                                  name:UIApplicationWillEnterForegroundNotification
-                                               object:nil];
-
-    [[NSNotificationCenter defaultCenter] addObserver:[self sharedInstance]
-                                             selector:@selector(applicationWillTerminate)
-                                                 name:UIApplicationWillTerminateNotification
                                                object:nil];
 
     RadarSdkConfiguration *sdkConfiguration = [RadarSettings sdkConfiguration];
@@ -1480,8 +1473,7 @@ BOOL _initialized = NO;
     [RadarSdkConfiguration_ObjC updateSdkConfigurationFromServer];
 }
 
-+ (void)logTermination { 
-    [[RadarLifecycleMarker sharedMarker] markCleanTermination];
++ (void)logTermination {
     [[RadarLogger sharedInstance] logWithLevel:RadarLogLevelDebug type:RadarLogTypeNone message:[RadarLifecycleMarker appTerminatingMessage] includeDate:YES includeBattery:YES append:YES];
 }
 
@@ -1749,9 +1741,6 @@ BOOL _initialized = NO;
     return [message toDictionary];
 }
 
-- (void)applicationWillTerminate {
-    [[RadarLifecycleMarker sharedMarker] markCleanTermination];
-}
 
 - (void)applicationWillEnterForeground {
     BOOL updated = [RadarSettings updateSessionId];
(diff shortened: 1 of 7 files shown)


RadarSDK/Radar.m
```diff
@@ -102,9 +111,29 @@ + (void)initializeWithAppGroup:(NSString *)appGroup {
     [Radar initializeWithPublishableKey:[RadarSettings publishableKey]];
 }
 
++ (void)logUncleanPreviousProcessIfNeeded {
```
nickdonnelly: Could this be in swift?

alanjcharles14 (dismissed):
ignore my previous comment (deleted), it was nonsense. Looks good to me, very cool implementation

RadarSDK/Radar.m
stevepopovich: yeahhhh might as well try

stevepopovich: done

nickdonnelly (dismissed):
lgtm

RadarSDK/Radar.m
```diff
@@ -183,7 +194,6 @@ + (void)initializeWithOptions:(RadarInitializeOptions *)options {
                                                 [Radar trackOnceWithDesiredAccuracy:RadarTrackingOptionsDesiredAccuracyMedium beacons:[Radar getTrackingOptions].beacons completionHandler:nil];
                                             }
 
-                                            [self flushLogs];
```
stevepopovich, line 186: this was just unneeded, we were already flushing earlier in this call

nickdonnelly (approved):

PR #692, opened 2026-08-31
FENCE-3000 Update RadarCoodinateSwift to be compatible with RadarCoordinate in ObjC

ShiCheng-Lu:
## Manual test steps

1. https://github.com/radarlabs/radar-sdk-ios/pull/700 used by RadarRoute
2.
3.

## Checklist

- [ ] (NA) For breaking changes: added a `MIGRATION.md` entry and bumped the version with `./set-version.sh`
- [ ] (NA) Updated README / public docs if the public API changed


RadarSDK/RadarCoordinate.swift
```diff
@@ -9,10 +9,17 @@
 import CoreLocation
 import Foundation
 
-struct RadarCoordinateSwift: Codable, Sendable, Equatable {
+@objc(RadarCoordinate)
+final class RadarCoordinateSwift: NSObject, Codable, Sendable {
```
alanjcharles14, line 5: The public objc api still exposes init/new. These previously created a zero coordinate, but the swift-generated initializer now crashes. I think we should preserve the old behavior with override init()

radar-sdk-ios.xcworkspace/xcshareddata/swiftpm/Package.resolved
```diff
@@ -0,0 +1,42 @@
+{
```
alanjcharles14: I think this got added unintentionally, we should remove.

alanjcharles14 (commented):
We also should add characterization tests for this like we have for the other models just to lock down conversion behavior. ex. https://github.com/radarlabs/radar-sdk-ios/pull/698/changes

RadarSDK/RadarCoordinate.swift
ShiCheng-Lu: good catch!

RadarSDK/RadarCoordinate.swift
```diff
@@ -26,8 +32,95 @@ struct RadarCoordinateSwift: Codable, Sendable, Equatable {
...
+        self.longitude = 0
+    }
+
+    @objc
+    internal static func coordinatesFrom(object: Any) -> [RadarCoordinateSwift]? {
+        guard let array = object as? [Any] else {
+            return nil
+        }
+        return array.compactMap(RadarCoordinateSwift.init)
+    }
```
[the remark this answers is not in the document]
ShiCheng-Lu: fake news

RadarSDK/RadarCoordinate.swift
```diff
@@ -26,8 +32,95 @@ struct RadarCoordinateSwift: Codable, Sendable, Equatable {
         self.longitude = longitude
     }
```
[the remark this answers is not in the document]
ShiCheng-Lu: fake news

RadarSDK/RadarCoordinate.swift
```diff
@@ -1,18 +1,24 @@
...
+        case LatLngDictionary
+    }
+    
     let latitude: Double
     let longitude: Double
 
+    @objc
+    public var coordinate: CLLocationCoordinate2D {
+        CLLocationCoordinate2D(latitude: latitude, longitude: longitude)
+    }
```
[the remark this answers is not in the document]
ShiCheng-Lu: fake news

RadarSDK/RadarCoordinate.swift
```diff
@@ -26,8 +31,95 @@ struct RadarCoordinateSwift: Codable, Sendable, Equatable {
...
+        self.latitude = 0
+        self.longitude = 0
+    }
+
+    @objc
+    internal static func coordinatesFrom(object: Any) -> [RadarCoordinateSwift]? {
+        guard let array = object as? [Any] else {
+            return nil
+        }
+        return array.compactMap(RadarCoordinateSwift.init)
```
alanjcharles14: i think the old objc implementation just returned nil if an entry was malformed but `compactMap` with return a partial array. So like...

now we could return 
```
[valid coordinate A, malformed coordinate, valid coordinate B]
```

instead of just nil. I think that's a regression but could be convinced otherwise :thinking:

RadarSDK/RadarCoordinate.swift
```diff
@@ -26,8 +31,95 @@ struct RadarCoordinateSwift: Codable, Sendable, Equatable {
...
+        } else {  // CodingStrategy.LatLngDictionary or default
+            var container = encoder.container(keyedBy: CodingKeys.self)
+            try container.encode(latitude, forKey: .latitude)
+            try container.encode(longitude, forKey: .longitude)
+        }
+    }
+
+    // `==` on an NSObject subclass routes through `isEqual:`, which defaults to identity.
+    // Compare values instead, matching the `Equatable` synthesis of the previous `struct`.
+    override func isEqual(_ object: Any?) -> Bool {
```
alanjcharles14: This preserves the swift struct’s value equality, but changes the public objc class from identity to value equality. Are we comfortable changing its behavior in sets and dictionary keys?

stevepopovich:
Is there a ticket or manual test steps to add here?

RadarSDK/RadarCoordinate.swift
ShiCheng-Lu: fixed

RadarSDK/RadarCoordinate.swift
ShiCheng-Lu: fixed, with only == override, swift will use == for value comparison, objc will use isEqual which is still pointer comparison.

stevepopovich (dismissed):

alanjcharles14:
@ShiCheng-Lu lgtm just 1 linting issue in the ci. Happy to approve once that's resolved. I usually just run `make format-check`, then `make lint-swift-fix`, then `make lint-swift` before a commit

ShiCheng-Lu:
ran into some issue where optional comparison actually routes through isEqual, so I've just removed equatable conformance (equatable via object address now) and provided a `valueEquals` comparison function for swift, and replaced it for where value comparison is used.

stevepopovich (approved):

alanjcharles14 (approved):

PR #697, opened 2026-09-01
FENCE-2804: Port one-shot location requests to Swift

stevepopovich:
## Summary

- Port `getLocationWithCompletionHandler:` and `getLocationWithDesiredAccuracy:completionHandler:` to Swift.
- Preserve Objective-C completion storage and fallback behavior behind `useSwiftLocationManager`.
- Pass authorization status from the injected Objective-C permissions helper into the Swift twin.

## Test Plan

1. Build and launch the Example app on a device.
2. Grant the app location access and open the `CSGN` tab.
3. Tap `Refresh` and confirm that `Current location` shows coordinates instead of `unavailable`.
4. Open iOS Settings, disable location access for the Example app, return to the app, and tap `Refresh` again.
5. Confirm that `Current location` reports `unavailable` and the Debug tab records a permissions error.
6. Re-enable location access, return to the Example app, tap `Refresh`, and confirm that coordinates are shown again.
7. The Example app does not expose the desired-accuracy overload separately; the CSGN refresh exercises the shared one-shot request path.

### DIFF
--- a/RadarSDK/RadarLocationManager+Swift.swift
+++ b/RadarSDK/RadarLocationManager+Swift.swift
@@ -21,10 +21,10 @@ import Foundation
 // auto-synthesized Swift module, so we cannot extend `RadarLocationManager` from Swift.
 // Static methods on this class are called from `RadarLocationManager.m` via
 // `RadarSDK-Swift.h`. Methods that need access to the manager's CLLocationManager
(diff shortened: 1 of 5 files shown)


alanjcharles14 (approved):
looks like it needs a rebase but otherwise lgtm. Can re-approve when the branch is clean

PR #701, opened 2026-09-02
FENCE-2804: Port `didDetermineState` to Swift

stevepopovich:
## Summary

- Port `locationManager:didDetermineState:forRegion:` to Swift.
- Preserve the Objective-C implementation as the fallback when `useSwiftLocationManager` is disabled.
- Keep synced-beacon entry and exit handling on the main actor.

## Test Plan

1. In the Radar dashboard, open the **Beacons** page for the same project and environment used by the Example app. Create or verify an enabled iBeacon with the physical beacon's UUID, major, minor, and coordinates within about 1 kilometer of the test location.
2. In the dashboard's remote SDK configuration, enable the tracking option `beacons`. If the project uses Radar modified beacon mode, disable `useRadarModifiedBeacon` for this test, then save the configuration.
3. Build and launch the Example app on a physical iPhone using the matching project key. Grant location access, select **Always** when testing in the background, and grant Bluetooth access if prompted.
4. Open the Example app's Tracking section, tap **Refresh**, and confirm that **Active tracking options** shows `beacons: true`.
5. Build and run once with `useSwiftLocationManager` enabled in the test configuration. Start tracking from the Example app, then confirm the Debug tab or monitored-region information shows a synced region with a `radar_beacon_` identifier.
6. Move the phone into range of the physical beacon, or restart tracking while already in range, and confirm the Debug tab records beacon-enter location handling.
7. Move the phone out of range of the beacon and confirm the Debug tab records beacon-exit location handling.
8. Stop tracking, build and run again with `useSwiftLocationManager` disabled, and repeat the previous three steps. Confirm that beacon synchronization and enter/exit handling produce the same observable results through the Objective-C fallback.

### DIFF
--- a/RadarSDK/RadarLocationManager+SwiftDelegate.swift
+++ b/RadarSDK/RadarLocationManager+SwiftDelegate.swift
@@ -8,16 +8,6 @@
 import CoreLocation
 import Foundation
 
-private final class RadarBeaconRegionStateBox: @unchecked Sendable {
-    let location: CLLocation
-    let region: CLBeaconRegion
-
-    init(location: CLLocation, region: CLBeaconRegion) {
-        self.location = location
-        self.region = region
-    }
-}
-
 // The `CLLocationManagerDelegate` half of `RadarLocationManager`
 extension RadarLocationManagerSwift {
 
@@ -89,25 +79,47 @@ extension RadarLocationManagerSwift {
 
         let isInside = state == .inside
         let source: RadarLocationSource = isInside ? .beaconEnter : .beaconExit
-        let stateBox = RadarBeaconRegionStateBox(location: location, region: beaconRegion)
+        let beaconUUID = beaconRegion.uuid
+        let beaconMajor = beaconRegion.major?.uint16Value
+        let beaconMinor = beaconRegion.minor?.uint16Value
         RadarLogger.shared.debug("🦅 \(isInside ? "Inside" : "Outside") beacon region | identifier = \(identifier)")
 
-        Task { @MainActor [stateBox] in
+        Task { @MainActor in
             let beaconManager = RadarBeaconManagerSwift.shared
             let completionHandler: RadarBeaconCompletionHandler = { _, _ in
-                RadarSwift.bridge?.handleLocation(stateBox.location, source: source)
+                RadarSwift.bridge?.handleLocation(location, source: source)
             }
 
             if identifier.hasPrefix(syncBeaconUUIDIdentifierPrefix) {
                 if isInside {
-                    beaconManager.handleBeaconUUIDEntry(for: stateBox.region, completionHandler: completionHandler)
+                    beaconManager.handleBeaconUUIDEntry(completionHandler: completionHandler)
                 } else {
-                    beaconManager.handleBeaconUUIDExit(for: stateBox.region, completionHandler: completionHandler)
+                    beaconManager.handleBeaconUUIDExit(completionHandler: completionHandler)
                 }
-            } else if isInside {
-                beaconManager.handleBeaconEntry(for: stateBox.region, completionHandler: completionHandler)
             } else {
-                beaconManager.handleBeaconExit(for: stateBox.region, completionHandler: completionHandler)
+                let beaconRegion: CLBeaconRegion
+                if let beaconMajor, let beaconMinor {
+                    beaconRegion = CLBeaconRegion(
+                        uuid: beaconUUID,
+                        major: beaconMajor,
+                        minor: beaconMinor,
+                        identifier: identifier
+                    )
+                } else if let beaconMajor {
+                    beaconRegion = CLBeaconRegion(
+                        uuid: beaconUUID,
+                        major: beaconMajor,
+                        identifier: identifier
+                    )
+                } else {
+                    beaconRegion = CLBeaconRegion(uuid: beaconUUID, identifier: identifier)
+                }
+
+                if isInside {
+                    beaconManager.handleBeaconEntry(for: beaconRegion, completionHandler: completionHandler)
+                } else {
+                    beaconManager.handleBeaconExit(for: beaconRegion, completionHandler: completionHandler)
+                }
             }
         }
     }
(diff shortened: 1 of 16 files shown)


RadarSDK/RadarLocationManager+SwiftDelegate.swift
```diff
@@ -66,6 +76,57 @@ extension RadarLocationManagerSwift {
...
+                    beaconManager.handleBeaconUUIDEntry(for: stateBox.region, completionHandler: completionHandler)
+                } else {
+                    beaconManager.handleBeaconUUIDExit(for: stateBox.region, completionHandler: completionHandler)
+                }
+            } else if isInside {
+                beaconManager.handleBeaconEntry(for: stateBox.region, completionHandler: completionHandler)
+            } else {
+                beaconManager.handleBeaconExit(for: stateBox.region, completionHandler: completionHandler)
+            }
+        }
```
ShiCheng-Lu: I feel like 

```
Task { @MainActor [stateBox] in
  let beaconManager 
  ...

}
```

would work.

RadarSDK/RadarLocationManager+SwiftDelegate.swift
```diff
@@ -66,6 +76,57 @@ extension RadarLocationManagerSwift {
...
+        guard let location = effectiveLocation(for: locationManager), let beaconRegion = region as? CLBeaconRegion else {
+            return
+        }
+
+        let isInside = state == .inside
+        let source: RadarLocationSource = isInside ? .beaconEnter : .beaconExit
+        let stateBox = RadarBeaconRegionStateBox(location: location, region: beaconRegion)
+        RadarLogger.shared.debug("🦅 \(isInside ? "Inside" : "Outside") beacon region | identifier = \(identifier)")
+
+        runOnMainThread { [stateBox] in
```
ShiCheng-Lu: probably better to capture the location, and what's necessary of beaconRegion separately, CLLocation is Sendable, CLBeaconRegion, not so much, don't need a beacon state box class.

handleBeaconUUIDEntry don't use the region?

ShiCheng-Lu: kind of annoying because we're using deprecated beacons, the new `CLMonitor.BeaconIdentityCondition` is sendable.

RadarSDK/RadarLocationManager+SwiftDelegate.swift
stevepopovich: Is that the more modern way of doing it? Not sure

RadarSDK/RadarLocationManager+SwiftDelegate.swift
stevepopovich: Yeah, I can take a look. I will not use the new beacon APIs now. It is shaped this way to match the original implementation but I can look at cleaning it up

RadarSDK/RadarLocationManager+SwiftDelegate.swift
ShiCheng-Lu: unless you require this to be run sychronously when it's on the main thread, Task should be prefered to DispatchQueue.

RadarSDK/RadarLocationManager+SwiftDelegate.swift
ShiCheng-Lu: yeah, because CLBeaconRegion is not really sendable, but I think every param that we do end up using can just be passed in, might be a small list of strings/ints, but I think that inspires more confidence than slapping a @unchecked Sendable in there.

ShiCheng-Lu: probably ok and not work the effort though.

RadarSDK/RadarLocationManager+SwiftDelegate.swift
```diff
@@ -8,6 +8,16 @@
 import CoreLocation
 import Foundation
 
+private final class RadarBeaconRegionStateBox: @unchecked Sendable {
```
nickdonnelly: Why does this need to be `@unchecked`?

RadarSDK/RadarLocationManager+SwiftDelegate.swift
nickdonnelly: Agree w/ @ShiCheng-Lu, tasks are the modern way

RadarSDK/RadarLocationManager+SwiftDelegate.swift
```diff
@@ -66,6 +76,57 @@ extension RadarLocationManagerSwift {
...
+            } else {
+                beaconManager.handleBeaconExit(for: stateBox.region, completionHandler: completionHandler)
+            }
+        }
+    }
+
+    // Core Location may call this delegate off the main thread, while beacon state is main-actor owned.
+    private static func runOnMainThread(_ work: @escaping @MainActor @Sendable () -> Void) {
+        if Thread.isMainThread {
+            MainActor.assumeIsolated {
```
nickdonnelly: Is there ever a case where the client code could cause a crash using `assumeIsolated`? I'm not super familiar but the docs state:

> Assume that the current task is executing on this actor’s serial executor, or _*stop program execution otherwise.*_

RadarSDK/RadarLocationManager+SwiftDelegate.swift
stevepopovich: I tried removing actually but it wouldn't compile. It might not be needed if I switch to `Task`

RadarSDK/RadarLocationManager+SwiftDelegate.swift
stevepopovich: Good call, removing it, it was unused

RadarSDK/RadarLocationManager+SwiftDelegate.swift
stevepopovich: yeah getting this removed

RadarSDK/RadarLocationManager+SwiftDelegate.swift
stevepopovich: Should be fine now with the use of Tasks

PR #703, opened 2026-09-02
FENCE-2998: Convert `RadarTrackingOptions` to swift

alanjcharles14:
## Summary

Migrates the complete `RadarTrackingOptions` implementation from Objective-C to Swift, including:

- Tracking presets and default values
- Enum and string mappings
- Dictionary parsing and serialization
- Equality behavior and date handling

This preserves the existing public Objective-C and Swift APIs through the hand-maintained header, with no intended API or runtime behavior changes. The main class uses Swift’s `@objc @implementation`, while a small Objective-C category remains as a compatibility shim for `trackingOptionsFromDictionary:` because Swift cannot implement Objective-C factory convenience initializers.

The existing behavior is covered by new Swift tests for presets, defaults, mappings and fallbacks, supported dictionary values, malformed input, serialization, equality, and Objective-C compatibility.

## Linear ticket

[FENCE-2998](https://linear.app/radarlabs/issue/FENCE-2998/convert-radartrackingoptions-to-swift)

## Manual test steps

- Build and run the example app
- Start tracking with the continuous, responsive, and efficient presets
- Start tracking with custom tracking options
- Restart the app and confirm persisted tracking options load correctly
- Stop tracking and confirm tracking state is cleared

## Checklist

- [x] New code is written in Swift
- [x] Added or updated unit tests
- [x] SwiftLint passes locally
- [x] No breaking public API changes
- [x] No README or public documentation changes required
- [x] No real API keys committed
- [x] Only fixed lint/format violations on lines changed


RadarSDK/RadarTrackingOptions.swift
```diff
@@ -0,0 +1,478 @@
...
+    }
+
+    // MARK: - Enum Mappings
+
+    @objc(stringForDesiredAccuracy:)
+    public class func string(
+        for desiredAccuracy: RadarTrackingOptionsDesiredAccuracy
+    ) -> String {
+        switch desiredAccuracy {
+        case .high:
```
nickdonnelly, line 110: might it be simpler to just stringify these with `"\(desiredAccuracy)"` and combine the first 3 cases?

nickdonnelly: same for all the other stringifiers in this file

nickdonnelly (dismissed):

RadarSDK/RadarTrackingOptions.m
```diff
@@ -6,385 +6,22 @@
 //
 
 #import "RadarTrackingOptions.h"
-#import "RadarUtils.h"
 
-@implementation RadarTrackingOptions
+// Swift's @objc @implementation cannot implement Objective-C factory
+// convenience initializers. Keep this category as a compatibility shim so
+// +trackingOptionsFromDictionary: remains available to Objective-C callers
+// and continues importing into Swift as init(from:).
```
ShiCheng-Lu, line 13: does this really not work? if the swift is just the class with a static function that generates the same objective-c signature

```
@objc
class RadarTrackingOptions {
  @objc(radar_trackingOptionsFromDictionary)
  static func radar_trackingOptionsFromDictionary(dictionary: [String: ?]) -> RadarTrackingOptions? {
  
  }
}
```

the call won't convert to swift convenience initializer outside the API?

ShiCheng-Lu: wouldn't it call the same objective-c signature [RadarTrackingOptions radar_trackingOptionsFromDictionary] and end up with the same swift implementation?

or is this just a @implementation issue? but we I think don't need to use @implementation

ShiCheng-Lu: actually yeah, we probably do because of the @property markers right? hmm...

ShiCheng-Lu: but the AI just writes swift properties and it works, so...

RadarSDK/RadarTrackingOptions.swift
```diff
@@ -0,0 +1,478 @@
...
+    public var useSignificantLocationChanges = false
+    public var beacons = false
+    public var useIndoorScan = false
+    public var useMotion = false
+    public var usePressure = false
+    public var batchInterval: Int32 = 0
+    public var batchSize: Int32 = 0
+    public var type: RadarTrackingOptionsType = .default
+
+    public class var presetContinuous: RadarTrackingOptions {
```
ShiCheng-Lu, line 53: I think prefer `public static var` unless there's a need for it be be subclassed.

RadarSDK/Include/RadarTrackingOptions.h
```diff
@@ -223,11 +223,16 @@ typedef NS_ENUM(NSInteger, RadarTrackingOptionsType) {
...
-+ (RadarTrackingOptions *_Nullable)trackingOptionsFromDictionary:(NSDictionary *_Nonnull)dictionary;
 + (NSString *)stringForType:(RadarTrackingOptionsType)type;
 + (RadarTrackingOptionsType)typeForString:(NSString *)str;
 - (NSDictionary *)dictionaryValue;
 
 @end
 
+@interface RadarTrackingOptions (Dictionary)
+
++ (RadarTrackingOptions *_Nullable)trackingOptionsFromDictionary:(NSDictionary *_Nullable)dictionary;
```
stevepopovich: So the dictionary `nullable` now?

stevepopovich (dismissed):

RadarSDK/Include/RadarTrackingOptions.h
alanjcharles14: I initially made it nullable to match the old implementation’s defensive nil handling. Butttt good catch! It’s no longer necessary now that the objc factory shim is there

alanjcharles14: fixed

RadarSDK/RadarTrackingOptions.swift
alanjcharles14: Normally I’d prefer `static` here too, but these are implementations of the existing objc class properties rather than new swift-only members.

With `@implementation`, the compiler only matches them when declared as `class var`/`class func`. Using `static` produced “missing class property” and “missing class method” errors for the presets and mapping methods. `class` also preserves the objc class-method dispatch behavior already exposed by the public header

RadarSDK/RadarTrackingOptions.m
alanjcharles14: This is specifically an `@implementation` constraint. We need `@implementation` here because tracking options is already declared in the objc header and is also used by public swift apis.

I tried using a normal swift class first. Making it public produced a duplicate interface in `RadarSDK-Swift.h`; making it internal created two distinct imported types with the same name and caused `RadarSyncManager.shouldTrack` to lose its public visibility.

The header’s `+trackingOptionsFromDictionary:` factory imports into swift as `init(from:)`, and swift’s `@implementation` proposal explicitly lists objc factory convenience initializers as unsupported. Implementing it as an initializer compiled but emitted an instance selector, causing `+[RadarTrackingOptions trackingOptionsFromDictionary:]: unrecognized selector` at runtime.

The small objc category keeps the existing class selector and forwards into the swift implementation while preserving one shared public type.

RadarSDK/RadarTrackingOptions.swift
alanjcharles14: I checked this, but these are imported objc `NS_ENUM` values. String interpolation produces the type and raw value, ie `RadarTrackingOptionsDesiredAccuracy(rawValue: 0)`, rather than `"high"`.

The explicit switches also preserve the existing string format and fallback behavior for unknown raw values, so I think we should keep these mappings explicit for now anyway

RadarSDK/RadarTrackingOptions.m
ShiCheng-Lu:
```
@objc(RadarTrackingOptions)
class RadarTrackingOptionsSwift {

}
```
should generate the implementation for ObjC, and `RadarTrackingOptions` should still be available to swift as a objc header class. And `RadarTrackingOptionsSwift` should be unused by swift or objective-c, basically only responsible for generating the definitions.

could this work?

alanjcharles14: mmmm just tested this... With `RadarTrackingOptionsSwift` internal, swift emits a unique objc class symbol, so existing objc references to `_OBJC_CLASS_$_RadarTrackingOptions` fail to link. Making it public emits a second `@interface RadarTrackingOptions` in `RadarSDK-Swift.h`, which conflicts with the header.  

@objc @implementation avoids both problems by treating the imported declaration and swift implementation as the same class, but it can't currently implement factory convenience initializers...

I think the objc shim is really the only compatibility preserving option for `+trackingOptionsFromDictionary:`. We will likely remove it whenever we're ready to cut over to a v 4.0 of the sdk and have some flexibility to change the public interface and make the public types natively swift-owned

ShiCheng-Lu: https://github.com/radarlabs/radar-sdk-ios/pull/707

ShiCheng-Lu: the class fail to link was because when it tries to compile RadarTrackingOptionsSwift, it references RadarTrackingOptions in some places, which requires the implementation, which don't exist. but since RadarTrackingOptionsSwift is RadarTrackingOptions, replacing all the names should work. I think this works, but didn't do a deep dive testing on it.

alanjcharles14: ah nice dude! You're definitely right, it seems to work when i replace all of the references with `RadarTrackingOptionsSwift`. I'm not sure which is the better approach tbh. I think my approach gives use us compiler-verified header compat but yours has some definite advantages and i'm not sure header compatibility is a strong enough reason to keep mine. Let me sleep on it 🤔

stevepopovich (dismissed):

stevepopovich (approved):

PR #710, opened 2026-09-04
FENCE-2804: Port region entry and exit delegates to Swift

stevepopovich:
## Summary

- Port the Core Location region-entry and region-exit delegate callbacks to Swift.
- Preserve geofence and beacon event sources, effective-location fallback behavior, and nearby-beacon payloads.
- Route through Swift only when `useSwiftLocationManager` is enabled and keep the Objective-C implementations as fallbacks.

## Test Plan

1. In the Radar dashboard, create a geofence around your physical test location and configure an enabled synced iBeacon that matches a physical beacon.
2. Build and launch the Example app on a physical iPhone, then grant location and Bluetooth permissions.
3. Start tracking and confirm the geofence and synced beacon region appear in the app's Debug information.
4. Move into and out of the configured geofence and confirm the Debug tab records geofence-enter and geofence-exit handling.
5. Move into and out of the beacon's range and confirm the Debug tab records beacon-enter and beacon-exit handling.
6. Repeat the previous steps with `useSwiftLocationManager` disabled and confirm the Objective-C fallback produces equivalent behavior.

### DIFF
--- a/RadarSDK/RadarLocationManager+SwiftDelegate.swift
+++ b/RadarSDK/RadarLocationManager+SwiftDelegate.swift
@@ -157,19 +157,31 @@ extension RadarLocationManagerSwift {
                 beaconRegion = CLBeaconRegion(uuid: beaconUUID, identifier: identifier)
             }
 
-            if identifier.hasPrefix(syncBeaconUUIDIdentifierPrefix) {
-                if isEntry {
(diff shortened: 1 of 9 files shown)


ShiCheng-Lu (approved):
lgtm

nickdonnelly (approved):

PR #717, opened 2026-09-08
Migrate RadarVerificationManager to Swift

ShiCheng-Lu:
## Summary

Migrate some of RadarVerificationManager to swift. The Swift Manager uses the objective c instance as variable/property access. Swift behaviour is gated with the sdkConfiguration `useSwiftVerificationManager`

https://linear.app/radarlabs/issue/FENCE-3014/convert-radarverificationmanager-to-swift

## Manual test steps

1. Spin up local server to serve the sdk configuration.
2. Launch app and set expected country and state code.
3. a debug print statement in the function shows the swift function was called.

Unit tests to check Swift updated ObjC vars.

## Checklist

- [X] New code is written in Swift (the SDK is migrating off Objective-C)
- [X] Added or updated unit tests (`make test`)
- [X] `make lint-swift` and `make format-check` pass locally
- [-] For breaking changes: added a `MIGRATION.md` entry and bumped the version with `./set-version.sh`
- [-] Updated README / public docs if the public API changed
- [X] No real API keys committed in the example app
- [X] Only fixed lint/format violations on lines I changed; removed any now-stale `.swiftlint-baseline.json` entries


RadarSDK/RadarAPIHelper.swift
```diff
@@ -112,19 +108,36 @@ final class RadarAPIHelper: Sendable {
...
-    func radarRequest(method: String, url: String, query: [String: String] = [:], headers: [String: String] = [:], body: [String: Any?] = [:]) async throws -> (Data, HTTPURLResponse) {
-
+    
+    enum RadarHost {
+        case defaultHost
+        case verifiedHost
+        case verifiedSecondaryHost
+    }
+    func radarRequest(host: RadarHost, method: String, url: String, query: [URLQueryItem] = [], headers: [String: String] = [:], body: [String: Any?] = [:]) async throws -> (Data, HTTPURLResponse) {
+        let hostUrl: String
```
nickdonnelly: are we on new enough swift version to do this?
```
let hostURL = switch host {
case .defaultHost:
    RadarSettings.DefaultHost
case .verifiedHost:
    RadarSettings.DefaultVerifiedHost
case .verifiedSecondaryHost:
    RadarSettings.DefaultVerifiedHostSecondary
}
```

RadarSDK/RadarSdkConfiguration.swift
```diff
@@ -83,8 +86,12 @@ class RadarSdkConfiguration: NSObject {
             "offlineEventGenerationEnabled": offlineEventGenerationEnabled,
             "useSwiftLocationManager": useSwiftLocationManager,
             "startUpdatesWhileInUse": startUpdatesWhileInUse,
-            "remoteTrackingOptions": RadarRemoteTrackingOptions.toDictionaries(remoteTrackingOptions) as Any,
```
ShiCheng-Lu, line 86: remote tracking option could crash when saving to NSUserDefault because RadarRemoteTrackingOptions.toDictionaries can return NSNull, which is not saveable. 
We're all good since this usually always comes from the server, in which case we have originalDict that gets returned, but in testing it will construct the dictionary from properties, which causes this crash.

RadarSDK/RadarVerificationManager.m
```diff
@@ -26,24 +26,46 @@
...
 @property (copy, nonatomic) NSString *expectedCountryCode;
 @property (copy, nonatomic) NSString *expectedStateCode;
+@property (weak, nonatomic) RadarVerificationManagerSwift *swiftInstance;
+
+@end
+
+
+@interface RadarVerificationManager () <RadarVerificationManagerSwiftHost>
+
+@property (nonatomic, retain) nw_path_monitor_t monitor;
```
ShiCheng-Lu, line 49: there is something special about how `monitor` is used, for now, I'm going to pretend it doesn't exist.

RadarSDK/RadarVerifiedLocationToken+Internal.h
```diff
@@ -10,15 +10,15 @@
...
 
-- (instancetype _Nullable)initWithUser:(RadarUser *_Nonnull)user
-                                events:(NSArray<RadarEvent *> *_Nonnull)events
-                                 token:(NSString *_Nonnull)token
-                             expiresAt:(NSDate *_Nonnull)expiresAt
-                             expiresIn:(NSTimeInterval)expiresIn
-                                passed:(BOOL)passed
-                        failureReasons:(NSArray<NSString *> *_Nonnull)failureReasons
-                                   _id:(NSString *_Nonnull)_id
-                              fullDict:(NSDictionary *_Nonnull)fullDict;
```
ShiCheng-Lu, line 21: this function is only used within `initWithObject`, the swift implementation will only allow `initWithObject` until we find a need to init with variables.

HenryZhang0:
bugbot run

RadarSDK/RadarLocationManager+Swift.swift
```diff
@@ -203,6 +205,24 @@ final class RadarLocationManagerSwift: NSObject {  // swiftlint:disable:this typ
...
+                host: host,
+                authorizationStatus: authorizationStatus,
+                locationManager: host.locationManager,
+                completionHandler: { status, location, stopped in
+                    continuation.resume(returning: (status, location, stopped))
+                }
+            )
+        }
+        return result
+    }
```
ShiCheng-Lu, line 224: added a getLocation (async) this will be used by trackVerified to get the location in parallel with the getConfig call. should improve our latency.

currently unused.

RadarSDK/RadarAPIClient.swift
```diff
@@ -147,27 +163,38 @@ public final class RadarAPIClient: Sendable {
...
+            value != nil ? URLQueryItem(name: key, value: value) : nil
+        }
+
+        let (data, response) = try await apiHelper.radarRequest(host: host, method: "GET", url: "config")
+
+        try assertResponseCode(response.statusCode)
+
+        let json = try JSONSerialization.jsonObject(with: data, options: []) as? [String: Any]
+        return RadarConfig.from(dictionary: json)
+    }
```
ShiCheng-Lu, line 197: currently unused

RadarSDK/RadarSDKFraud.swift
```diff
@@ -57,4 +61,14 @@ final class RadarSDKFraud: @unchecked Sendable {
         }
         return (.success, payload)
     }
+
+    static let isSharingSelector = NSSelectorFromString("isSharing")
+    public func isSharing() -> Bool {
+        instance.perform(RadarSDKFraud.isSharingSelector).takeUnretainedValue() as? Bool ?? false
```
alanjcharles14: `isSharing` returns a primitive `BOOL`, but `perform` treats it as an object. I think this will crash for both `true` (invalid obj pointer) and `false` (nil unwrap) before reaching the cast.
https://developer.apple.com/documentation/objectivec/nsobjectprotocol/perform%28_%3A%29

RadarSDK/RadarAPIHelper.swift
```diff
@@ -113,18 +111,35 @@ final class RadarAPIHelper: Sendable {
...
+    enum RadarHost {
+        case defaultHost
+        case verifiedHost
+        case verifiedSecondaryHost
+    }
+    func radarRequest(host: RadarHost, method: String, url: String, query: [URLQueryItem] = [], headers: [String: String] = [:], body: [String: Any?] = [:]) async throws -> (Data, HTTPURLResponse) {
+        let hostUrl =
+            switch host {
+            case .defaultHost:
+                RadarSettings.DefaultHost
```
alanjcharles14: won't these bypass configured host overrides for requests outside of this migration? We don't want to use `RadarSettings.host` and `RadarSettings.verifiedHost`?

RadarSDK/RadarVerificationManager.swift
```diff
@@ -0,0 +1,176 @@
...
+        let tokenValid = lastDistanceToStateBorder > 1609 && timeElapsed < lastToken.expiresIn && tokenPassed
+
+        RadarLogger.debug(
+            "Last token \(tokenValid ? "valid" : "invalid") | lastToken.expiresIn = \(lastToken.expiresIn); lastTokenElapsed = \(timeElapsed); lastToken.passed = \(lastToken.passed); lastDistanceToStateBorder = \(lastDistanceToStateBorder)"
+        )
+
+        return tokenValid
+    }
+
+    func setExpectedJurisdiction(countryCode: String, stateCode: String) {
```
alanjcharles14: we currently accept nullable country/state codes and these convert `nil` to empty strings which would change the stored state and request fields. I think they should be `String?` ?

RadarSDK/RadarLocationManager+Swift.swift
alanjcharles14: nice!

RadarSDK/RadarSDKFraud.swift
ShiCheng-Lu: good catch!

alanjcharles14 (approved):

PR #719, opened 2026-09-14
FRA-2220: Add encrypted payloads

alanjcharles14:
## Summary

Encrypts fraud payloads for `trackVerified` and `revealRisk`. Verification managers collect fraud signals once, and each network attempt creates a fresh HPKE seal, encryption attempt ID, and timestamp using the outgoing request’s headers and install ID. Transport retries reuse the collected signals without recollecting them.

Preserves verified-host routing, the existing lost-connection retry policy, and replay/offline behavior. Collection or sealing failures stop the request without plaintext fallback or network-failure replay buffering. Buffered replays omit the encrypted payload.

The fraud SDK remains optional, but consumers using fraud features must upgrade core and the fraud SDK together. Core now requires the prepared-payload API; the matching fraud SDK change removes the legacy plaintext and one-shot encrypted-payload methods.

## Related
https://github.com/radarlabs/radar-sdk-ios-fraud/pull/26

## Manual test steps

Tested on a physical iPhone running iOS 26.6.1 against a local server through ngrok, using matching test keys and enabled decryption.

1. Run `trackVerified` and `revealRisk` from the Example app. Confirm no invalid-signature reason.
2. Change the fraud SDK's logical audience, rebuild, and repeat both requests. Confirm `fraud_compromised_invalid_request_signature`.
3. Restore the correct audience and rebuild.

Both valid/invalid audience comparisons passed locally. Production rejection behavior remains unresolved.

## Checklist


RadarSDK/RadarTrackVerifiedRequestPreparer.swift
```diff
@@ -0,0 +1,61 @@
...
+    @objc(prepareRequest:completionHandler:)
+    func prepareRequest(
+        _ request: URLRequest,
+        completionHandler:
+            @escaping @Sendable (
+                RadarStatus,
+                URLRequest?,
+                NSError?
+            ) -> Void
+    ) {
```
ShiCheng-Lu, line 39:
```
fun prepareRequest(_:) async -> (RadarStatus, URLRequest?, NSError?) {

}
```

maps to objc completionHandler function. Since this is a new function, use async.

RadarSDK/Include/RadarSDKFraudProtocol.h
```diff
@@ -21,6 +21,9 @@ typedef void (^RadarFraudPayloadCallback)(NSDictionary<NSString *, id> *_Nullabl
 
 - (void)getFraudPayloadWithOptions:(NSDictionary<NSString *, id> *)options completionHandler:(RadarFraudPayloadCallback)completionHandler;
 
+- (void)getEncryptedFraudPayloadWithOptions:(NSDictionary<NSString *, id> *)options
+                          completionHandler:(RadarFraudPayloadCallback)completionHandler;
+
```
ShiCheng-Lu, line 26: I'm pretty sure this is unnecessary unless we're calling the fraud function in ObjC, which we shouldn't be.

RadarSDK/RadarSDKFraud.swift
```diff
@@ -39,14 +39,27 @@ final class RadarSDKFraud: @unchecked Sendable {
     }
 
     static let getFraudPayloadSelector = NSSelectorFromString("getFraudPayloadWithOptions:completionHandler:")
-    public func getFraudPayload(sdkConfiguration: RadarSdkConfiguration?) async -> (RadarStatus, String?) {
-        let options = sdkConfiguration?.dictionaryValue() ?? [:]
+    static let getEncryptedFraudPayloadSelector = NSSelectorFromString(
```
ShiCheng-Lu, line 42: add selector check to this object's initializer.

prob change it to a guard function too so it's a list instead of a or statement
```
init?(instance: NSObject) {
        guard instance.responds(to: RadarSDKFraud.initializeSelector),
 instance.responds(to: RadarSDKFraud.getFraudPayloadSelector),
  instance.responds(to: encrypted...) {
            return nil
        }
        self.instance = instance
    }
```

RadarSDK/RadarSDKFraud.swift
```diff
@@ -55,6 +68,83 @@ final class RadarSDKFraud: @unchecked Sendable {
...
         }
+
         return (.success, payload)
     }
+
+    func prepareEncryptedRequest(
+        _ request: URLRequest,
+        canonicalRoute: String,
+        options fraudOptions: [String: Any]
+    ) async throws -> URLRequest {
```
ShiCheng-Lu, line 79: this is not a RadarSDKFraud function, should not live in RadarSDKFraud.

RadarSDK/RadarSDKFraud.swift
```diff
@@ -39,14 +39,27 @@ final class RadarSDKFraud: @unchecked Sendable {
     }
 
     static let getFraudPayloadSelector = NSSelectorFromString("getFraudPayloadWithOptions:completionHandler:")
```
ShiCheng-Lu, line 41: remove this if its unused.

RadarSDK/RadarSDKFraud.swift
ShiCheng-Lu, line 17: maybe we can add a log warn here if someone tries to initialize RadarSDKFraud with an object that doesn't respond to the correct selectors.

ShiCheng-Lu: something I should've added.

RadarSDK/RadarAPIHelper.m
```diff
@@ -68,13 +68,55 @@ - (instancetype)init {
...
+                       url:url
+                   headers:headers
+                    params:params
+                     sleep:sleep
+                logPayload:logPayload
+           extendedTimeout:extendedTimeout
+            prepareRequest:prepareRequest
+ preparationFailureHandler:nil
+         completionHandler:completionHandler];
+}
```
ShiCheng-Lu, line 109: lets just use the full function and pass in nil instead of making anotherone with just one less param.

RadarSDK/RadarAPIClient.swift
```diff
@@ -147,7 +146,13 @@ public final class RadarAPIClient: Sendable {
             "xPlatformSDKVersion": RadarSettings.xPlatform ? RadarSettings.xPlatformSDKVersion : nil,
         ]
 
-        let (data, response) = try await apiHelper.radarVerifiedRequest(method: "POST", url: "reveal/risk", body: params)
+        let (data, response) = try await apiHelper.radarVerifiedRequest(
+            method: "POST",
+            url: "reveal/risk",
+            body: params,
+            useSecondaryVerifiedHost: useSecondaryVerifiedHost,
+            prepareRequest: prepareRequest
```
ShiCheng-Lu, line 154: TBH, I really don't like this prepareRequest thing that modified the request after building it. If it changes the body, just modify the body.

ShiCheng-Lu: and it should happen before being sent to RadarAPIHelper.

nickdonnelly: I'm inclined to agree, it should either be done prior to getting to the `RadarAPIHelper` or entirely a black box responsibility of the API helper to invoke encryption

RadarSDK/RadarRevealRiskManager.swift
```diff
@@ -46,21 +46,25 @@ final class RadarRevealRiskManager: NSObject, @unchecked Sendable {
...
+    ) async throws -> RadarRevealRiskToken {
         guard let fraudSDK else {
             throw RadarError(status: .errorPlugin)
         }
 
-        let (status, payload) = await fraudSDK.getFraudPayload(sdkConfiguration: RadarSettings.sdkConfiguration)
-        guard let payload, status == .success else {
-            throw RadarError(status: status)
-        }
+        let fraudOptions = RadarSettings.sdkConfiguration?.dictionaryValue() ?? [:]
```
nickdonnelly, line 56: is this a behavioral change? is defaulting to `[:]` on null the behavior we actually want?

RadarSDK/RadarSDKFraud.swift
```diff
@@ -55,6 +68,83 @@ final class RadarSDKFraud: @unchecked Sendable {
...
+        canonicalRoute: String,
+        installId: String,
+        options fraudOptions: [String: Any]
+    ) throws -> [String: Any] {
+        var options = fraudOptions
+        options["method"] = request.httpMethod
+        options["canonicalRoute"] = canonicalRoute
+        options["encryptionAttemptId"] =
+            try RadarUtils.makeFraudEncryptionAttemptId()
+        options["issuedAt"] = Int(Date().timeIntervalSince1970)
```
nickdonnelly, line 136: what are the consequences of this timestamp being off if the user's clock is set to the wrong time?

cameron-radar: it would defeat the purpose:

Cameron Morrow  [5:54 PM]
a few points about the request encryption doc:

the current pr doesn't include timestamp/requestId. because these would come from the client and not the server they could be compromised and wouldn't guarantee freshness or replay prevention, so I don't think they're worth adding.

RadarSDK/RadarAPIClient.swift
alanjcharles14: Agreed that passing `prepareRequest` through these layers makes this harder to follow. I’ll move encrypted-body preparation into the request-building path and keep RadarSDKFraud focused on calling the fraud library. The reason for the current placement was to encrypt again on each network attempt using the actual outgoing headers and route. I’ll preserve that retry behavior but can remove the general purpose request mutation callback.

ShiCheng-Lu: I don't think we need to worry about re-encrypting the request. just encrypting once is ok, and retry with the same encrypted body.

RadarSDK/RadarTrackVerifiedRequestPreparer.swift
alanjcharles14: Changed this to an async method with the completion handler bridge for objc. It’s now called `prepareBody` since encryption happens before building the request.

RadarSDK/Include/RadarSDKFraudProtocol.h
alanjcharles14: ah yeah good catch. removed.

RadarSDK/RadarSDKFraud.swift
alanjcharles14: Moved the encrypted-selector check into the initializer and converted the checks to a guard

RadarSDK/RadarSDKFraud.swift
alanjcharles14: Moved context and body preparation into RadarFraudPayloadPreparer, shared by trackVerified and revealRisk. Removed the old request-building method from RadarSDKFraud, which now only handles the fraud-library bridge.

RadarSDK/RadarSDKFraud.swift
alanjcharles14: removed

RadarSDK/RadarSDKFraud.swift
alanjcharles14: added a warning

RadarSDK/RadarAPIHelper.m
alanjcharles14: Removed both preparation-related overloads entirely. Encryption now happens before calling RadarAPIHelper, so only the original request method remains

RadarSDK/RadarAPIClient.swift
alanjcharles14: okay fair enough, apologies for over-complicating. 

Reworked both paths to prepare the encrypted body before calling RadarAPIHelper and removed the request mutation callbacks. Immediate retries now reuse the same encrypted body and headers without re-encrypting. Updated the tests to cover this and preserve existing replay behavior.

RadarSDK/RadarRevealRiskManager.swift
alanjcharles14: This preserves the existing behavior. `RadarSDKFraud.getFraudPayload` already used `sdkConfiguration?.dictionaryValue() ?? [:]`. The fallback moved here with the refactor. We still attempt fraud collection when configuration is unavailable.

RadarSDK/RadarSDKFraud.swift
alanjcharles14: A device clock outside the server’s configured window (default ±60 seconds) causes payload validation to fail, so clock skew can affect legitimate users. @cameron-radar 's  point is valid. A client-provided timestamp doesn’t guarantee freshness from a compromised client or prevent replay within that window. I included it to match the existing server contract, which requires and validates `issuedAt`. Removing it would need a coordinated server/SDK change.

RadarSDK/RadarSDKFraud.swift
```diff
@@ -42,15 +44,28 @@ final class RadarSDKFraud: @unchecked Sendable {
...
-        let options = sdkConfiguration?.dictionaryValue() ?? [:]
+    static let getEncryptedFraudPayloadSelector = NSSelectorFromString(
+        "getEncryptedFraudPayloadWithOptions:completionHandler:"
+    )
+
+    public func getEncryptedFraudPayload(
+        options: [String: Any]
+    ) async -> (RadarStatus, String?) {
+        guard instance.responds(to: Self.getEncryptedFraudPayloadSelector) else {
+            RadarLogger.shared.warning("RadarSDKFraud does not support encrypted payloads; update the fraud SDK.")
```
ShiCheng-Lu: still think we can force fraud + main to both be new version or both be old version.

especially because `trackVerified` not working but `isSharing` working is not really a valid configuration. this version requires the new fraud version because of encryption.

RadarSDK/RadarAPIClient.swift
```diff
@@ -136,9 +136,43 @@ public final class RadarAPIClient: Sendable {
...
         useSecondaryVerifiedHost: Bool,
+        fraudPayloadPreparer: RadarFraudPayloadPreparer
     ) async throws -> RadarRevealRiskToken {
+        let body = await makeRevealRiskBody()
+
+        let method = "POST"
+        let canonicalRoute = "/v1/reveal/risk"
+        let host =
+            useSecondaryVerifiedHost
+            ? RadarSettings.DefaultVerifiedHostSecondary
```
ShiCheng-Lu: lets use the defaultVerifiedHostSecondary instead. If we ever make it a stored value.

RadarSDK/RadarAPIClient.swift
```diff
@@ -136,9 +136,43 @@ public final class RadarAPIClient: Sendable {
...
+            canonicalRoute: canonicalRoute,
+            headers: headers
+        )
+
+        let (data, response) = try await apiHelper.request(
+            method: method,
+            url: "\(host)\(canonicalRoute)",
+            headers: headers,
+            body: encryptedBody
+        )
```
ShiCheng-Lu: I would like this to keep using apiHelper.radarRequest if possible.

ShiCheng-Lu: and pass in the AAD values for origin/product/sdk ver/authorization instead of the whole header

RadarSDK/RadarFraudPayloadPreparer.swift
```diff
@@ -0,0 +1,74 @@
...
+//
+
+import Foundation
+
+struct RadarFraudPayloadPreparer {
+    let fraudSDK: RadarSDKFraud
+    let options: [String: Any]
+
+    func prepareBody(
+        _ body: [String: Any],
```
ShiCheng-Lu: we're not doing anything with the body? is there a plan to encrypt the entire body? this is basically just getFraudPayload.

ShiCheng-Lu (commented):
I guess it works, but much of the changes feel a little unnecessary when all we're doing about the body is just fraudPayload

couldn've just replaced `fraudSDK.getFraudPayload` in RadarRevealRiskManager + getFraudPayload in VerificationManager with the new fraudSDK.getEncryptedFraudPayload function.

RadarSDK/RadarSDKFraud.swift
alanjcharles14: you were right i was wrong 😭

RadarSDK/RadarAPIClient.swift
alanjcharles14: good catch. fixed

RadarSDK/RadarFraudPayloadPreparer.swift
alanjcharles14: eh yeah not sure what the long-term plan is tbh. I know we want to move more stuff into the encrypted payload but that hasn't been scoped yet so i just refactored again (see other comment)

alanjcharles14:
@ShiCheng-Lu no you're not wrong at all. My intent originally was to ensure encryption used the exact outgoing request context but it made things more complicated than necessary. 

I moved encrypted payload collection back into the managers. the client now gets the payload and the request context instead of a collector. Removed the api client collection stuff and updated the tests. Thanks for the detailed reviews 🙏

RadarSDK/RadarRevealRiskManager.swift
```diff
@@ -46,21 +46,36 @@ final class RadarRevealRiskManager: NSObject, @unchecked Sendable {
...
+            options: RadarSettings.sdkConfiguration?.dictionaryValue() ?? [:]
+        )
+
+        let payload = try await preparer.getEncryptedPayload(
+            installId: installId,
+            canonicalRoute: "/v1/reveal/risk",
+            origin: headers["Origin"],
+            product: headers["X-Radar-Product"],
+            sdkVersion: headers["X-Radar-SDK-Version"],
+            authorization: headers["Authorization"]
```
ShiCheng-Lu: i'd like these to be from 

```
origin: Bundle.main.bundleIdentifier
product: RadarSettings.product
sdkVersion: RadarUtils.sdkVersion
authorization: RadarSettings.publishableKey
```

all of these are synchronous. (also "Origin" doesn't exist, equivalent would probably be "X-Radar-Mobile-Origin")

and leave the APIClient + APIHelper completely alone, don't pass in a pre prepped headers

RadarSDK/RadarUtils.swift
```diff
@@ -285,6 +286,29 @@ class RadarUtils: NSObject {
...
+                message: "Failed to generate encryption attempt ID"
+            )
+        }
+
+        return Data(bytes)
+            .base64EncodedString()
+            .replacingOccurrences(of: "+", with: "-")
+            .replacingOccurrences(of: "/", with: "_")
+            .replacingOccurrences(of: "=", with: "")
+    }
```
ShiCheng-Lu: feels like this should be a `RadarFraudPayloadPreparer` static function instead of RadarUtil, it's pretty specific to fraud encryption.

If you want to make this function more util like maybe
`func base64RandomBytes(size: 16)` would be better.

ShiCheng-Lu (commented):
everything else looks good, 

basically stop trying to mess with APIClient >:(

RadarSDK/RadarRevealRiskManager.swift
alanjcharles14: fair enough 😅  i changed it. 

re `Origin`: The server already uses `X-Radar-Mobile-Origin` for mobile-ID restrictions, but fraud decryption only reads standard `Origin`. Was that intentional do you know, or should mobile encryption bind to the mobile-origin header? I'm using nil to match the current contract for now but can follow up with a server fix before merging this.

ShiCheng-Lu (dismissed):
I would just double check `origin = nil` actually with the decryption, if the request it sends contains an origin header that apple automatically attach for some reason, then we'll be in a pickle.

RadarSDK/RadarAPIClient.swift
alanjcharles14: @ShiCheng-Lu just re-reading the [tech-spec](https://app.notion.com/p/radarlabs/Radar-request-encryption-REVISED-3c7e42b18f9680c59ed8ecd84d4debf0?source=copy_link) now that I'm getting into android and came across this: 

> A retry creates a new HPKE seal, encryption attempt ID, and timestamp. It never retries as plaintext.

just want to callout as it's something we should be clear about. We don't retry as plaintext but we do reuse the encrypted body now. Should we actually be re-encrypting on every request? Wdyt?

ShiCheng-Lu (approved):

PR #723, opened 2026-09-18
FENCE-3042: fix `RadarInitializeOptions` linkage

alanjcharles14:
## Summary

Fix `RadarInitializeOptions` linkage for binary sdk consumers by using `@objc @implementation` to implement its existing public Objective-C interface. This exports the class without generating a duplicate declaration and preserves the existing public API.

## Manual test steps
Validated release device and simulator archives, exports across all XCFramework architectures, and standalone swift and objective-c consumers. All 14 focused tests and swift lint passed.

## Checklist


ShiCheng-Lu (commented):
there are a lot of other classes that needs to be fixed, ~~I also think we should just make the swift class public and delete the .h file if we can.~~ actually I think there might have been some issue with that related to linking with the generated swift header. keeping the `.h` and `@implementation extension` is good.

ShiCheng-Lu (approved):

stevepopovich (approved):
Interesting 🤔 I am surprised by this. Can we ask AI to write a lint rule for this maybe? If this is the correct way to do it

PR #724, opened 2026-09-21
FENCE-3044: Fix public Objective-C class linkage

stevepopovich:
## Summary

Fix public Objective-C class linkage for the 24 remaining Swift-converted classes on top of the `RadarInitializeOptions` fix from PR 723. This restores the Objective-C class symbols required by Release consumers.

Resolves FENCE-3044.

## Changes

- Implement public-header-backed Swift classes with Swift 6.1 `@objc @implementation extension` declarations.
- Restore `RadarCoordinateSwift` as the internal Codable value model and use `RadarCoordinate` as the public Objective-C wrapper.
- Rename the internal route value model to `RadarRouteValue` while preserving route decoding and dictionary output.
- Preserve Codable behavior through private/value payload models where imported Objective-C classes cannot safely conform directly.
- Add source and Release-binary linkage checks, including an Objective-C fixture that links and constructs `RadarTripOptions`.
- Run the linkage check in CI and document the migration requirement for future conversions.

#720 remains separate and should be rebased after this change merges. TBH I will probably just close it and like the nightly action re-run to verify it does the future conversions correctly

## Test Plan

- Build the Example app using the Release configuration.
- Launch the Example app and initialize Radar with a valid publishable key.
- Create trip options with an external ID and destination settings.
- Confirm the app links successfully and launches without an Objective-C class linkage failure.


RadarSDK/RadarRevealRiskToken.swift
stevepopovich, line 1: This one was probably the scariest one to me. Extra eyes here would be good

stevepopovich:
Superseded by #725, the minimal replacement focused on the 25 affected Objective-C-backed classes and required linkage call sites.

PR #725, opened 2026-09-22
FENCE-3044: Fix Objective-C linkage for Swift-backed SDK models

stevepopovich:
## Summary

This replacement keeps the Objective-C linkage fix focused on the 25 Swift-backed model classes affected by #723. Each implementation now implements its existing Objective-C interface directly, while preserving the public headers, selectors, defaults, parsing, dictionary output, and wire formats.

The change also keeps the Swift-only value models and payload helpers private to their existing files, applies only the required imported-method call-site updates, and marks the internal Objective-C headers private in the podspec.

Adding better CI and readme improvements got out of hand, so trying to keep this scoped for now. I know there are some weird things here but I think this is good now. Please let me know if there are questions

Related to #723.
Supersedes #724.

## Test Plan

1. Build the Example app in Release configuration.
2. Launch the Example app on an iOS device or simulator.
3. Initialize Radar from the app.
4. Construct trip options with an external ID and destination values.
5. Confirm the app links and runs without an Objective-C class linkage failure.
6. Generally smoke test the various functions in the example app


RadarSDK/RadarFraud+Internal.h
```diff
@@ -5,7 +5,7 @@
 //  Copyright © 2021 Radar Labs, Inc. All rights reserved.
 //
 
-#import "RadarFraud+Internal.h"
+#import "RadarFraud.h"
```
stevepopovich, line 8: This makes the RadarFraud class available to CocoaPods, but private to public consumers

RadarSDK/RadarRevealRiskToken.swift
stevepopovich, line 1: Riskiest/scariest change, if anyone can double check here

RadarSDK/RadarTrip.swift
```diff
@@ -132,9 +102,8 @@ class RadarTrip: NSObject {
             dictionary["externalId"] = externalId
         }
 
-        if let id = _id {
```
alanjcharles14, line 135: why did we remove this check?

RadarSDK/RadarCircleGeometry.swift
```diff
@@ -7,17 +7,21 @@
...
+    var center: RadarCoordinate = RadarCoordinate()
+    var radius: Double = 0
 
-    /// Keeps the hand-written Objective-C initializer working after the implementation moved to Swift.
+    override init() {
+        super.init()
+    }
+}
+
+extension RadarCircleGeometry {
```
ShiCheng-Lu, line 19: it's fine to split it but I don't think we really need to.

I think the header itself not declaring the function should be fine even if someone could technically call the function via selector.

All our old objc was technically callable, as long as it's not publicly advertised in the `.h` file (won't apear in intellisense) we should be fine.

But also ok if you want to keep the internal functions totally internal.

RadarSDK/RadarChain.swift
```diff
@@ -1,101 +1,40 @@
...
-            }
-            self.metadata = NSDictionary(dictionary: foundationMetadata)
-        } else {
-            self.metadata = nil
-        }
+@objc @implementation extension RadarChain {
+    var slug = ""
+    var name = ""
+    var externalId: String?
+    var metadata: [AnyHashable: Any]?
```
ShiCheng-Lu, line 25: why are these `vars` instead of `let`

RadarSDK/RadarChain.swift
```diff
@@ -1,101 +1,40 @@
...
-//
-//  Copyright © 2019 Radar Labs, Inc. All rights reserved.
-//
-
 import Foundation
 
-@objc(RadarChain)
-@objcMembers
-final class RadarChain: NSObject, Codable {
+struct RadarChainPayload: Codable {
```
ShiCheng-Lu, line 3: what's the point of this? why is it not used except by tests?

alanjcharles14:
i think overall this is good. I built the example on my device with a `Release` build and tested: 
- trackOnce
- trackVerified
- startTrackingVerified
- addTripLeg
- reorderTripLeg
- startTrip
- completeTrip
- cancelTrip
- revealRisk

image

once the comments are addressed, happy to approve and get this patch release out. nice work!

RadarSDK/RadarCoordinate.swift
```diff
@@ -1,10 +1,11 @@
 import CoreLocation
 import Foundation
 
-@objc(RadarCoordinate)
-final class RadarCoordinateSwift: NSObject, Codable, Sendable {
```
ShiCheng-Lu, line 5: The whole point of this was to use the same class for the objective C implementation and swift. Which are used in regular API responses, sync region save states, and notification coordinate comparisons. And now we're going to split it back up again?

RadarSDK/RadarRoute.swift
```diff
@@ -29,140 +21,152 @@ public struct RadarRoute: Codable, Sendable {
...
+    geometry: nil)
 
-// MARK: - ObjC classes, backed by RadarRoute struct
+@objc @implementation extension RadarRouteDistance {
+    private final var route = emptyRoute
 
-@objc(RadarRouteDistance)
-class RadarRouteDistance: NSObject {
-    @objc public var value: Double { route.distance.value }
-    @objc public var text: String { route.distance.text }
```
ShiCheng-Lu, line 40: I want to keep @objc and public, even though it doesn't do anything, to mark function that has an objc header declaration, and public if it's public.

can you tell AI to keep them?

RadarSDK/RadarChain.swift
ShiCheng-Lu: oh extension moment...

RadarSDK/RadarCircleGeometry.swift
ShiCheng-Lu: Oh, @objc @implementation must contain only what's in the header. So it must be split

